PR #5, opened 2017-07-13
OpenDBC updates

palmerit:
Found the Brake Position.


palmerit:
This may not actually be the brake position.  Another user noticed a couple spots where it didn't match up.  Closing the PR until I can confirm.

PR #8, opened 2017-08-04
Subie

palmerit:
Add door status and cruise control.

I suspect BrakePosition or BrakeStatus isn't correct, but leaving them for myself or another to investigate.  This file is still a work in progress.

### DIFF
--- a/subaru_outback_2016_eyesight.dbc
+++ b/subaru_outback_2016_eyesight.dbc
@@ -61,4 +61,4 @@ VAL_ 324 CruiseSetButton 1 "On" 0 "Off" ;
 VAL_ 324 CruiseResumeButton 1 "On" 0 "Off" ;
 VAL_ 324 CruiseStatus 1 "On" 0 "Off" ;
 VAL_ 324 BrakeStatus 1 "On" 0 "Off" ;
-VAL_ 324 DoorStatus 3 "Both" 2 "Passenger" 1 "Driver" 0 "Closed"
+VAL_ 884 DoorStatus 3 "Both" 2 "Passenger" 1 "Driver" 0 "Closed"


subaru_outback_2016_eyesight.dbc
```diff
@@ -34,19 +34,31 @@ NS_ :
...
+ SG_ CruiseStatus : 48|1@0+ (1,0) [0|0] ""  XXX
+ SG_ BrakeStatus : 51|1@0+ (1,0) [0|0] ""  XXX
 
 BO_ 320 Throttle: 8 XXX
- SG_ ThrottlePosition : 7|8@0+ (1,0) [0|255] "" XXX
+ SG_ ThrottlePosition : 7|8@0+ (1,0) [0|255] ""  XXX
 
 BO_ 209 NEW_MSG_1: 8 XXX
- SG_ BrakePosition : 23|8@0+ (1,0) [0|255] "" XXX
+ SG_ BrakePosition : 23|8@0+ (1,0) [0|255] ""  XXX
```
rbiasini: @palmerit : is this really brake position or rather a measure of the brake fluid pressure?

subaru_outback_2016_eyesight.dbc
```diff
@@ -34,19 +34,31 @@ NS_ :
...
 
 BO_ 2 NEW_MSG_2: 8 XXX
- SG_ Steering_Angle : 7|16@0- (1,0) [0|65535] "" XXX
-
-
-
-
+ SG_ SteeringAngle : 7|16@0- (1,0) [0|65535] ""  XXX
```
rbiasini: it would be nice to include a conversion factor so that it can be read in meaningful units (deg?). Also, it should have different limits, since it's signed.

subaru_outback_2016_eyesight.dbc
rbiasini: @palmerit , is this really a measurement of the brake pedal position or rather a measure of the brake fluid pressure?

subaru_outback_2016_eyesight.dbc
TheMutley: it's probably 540-540 deg, but I don't have a Subaru to test how far it goes both ways, so I left it at 1.0 for now

subaru_outback_2016_eyesight.dbc
```diff
@@ -34,19 +34,31 @@ NS_ :
...
 BU_: XXX
+VAL_TABLE_ OnOffList 1 "On" 0 "Off" ;
 
+BO_ 884 NEW_MSG_3: 8 XXX
+ SG_ DoorStatus : 31|8@0+ (1,0) [0|255] "" XXX
+
+BO_ 324 NEW_MSG_3: 7 XXX
+ SG_ BrakeApplied : 8|1@0+ (1,0) [0|0] ""  XXX
+ SG_ CruiseSetButton : 3|1@0+ (1,0) [0|0] ""  XXX
+ SG_ CruiseResumeButton : 4|1@0+ (1,0) [0|0] ""  XXX
```
rbiasini: can cruise Set and Resume buttons be controlled independently? if yes, then it's ok to keep them in separate signals. If not they should be merged in a 2 bits signal (in the Acura/Honda all the Cruise buttons are mutually exclusive).

palmerit:
Submitting a new PR.

PR #10, opened 2017-08-10
Update Subaru Outback

palmerit:
After some comments back, and some additonal messages being found, I'm submitting this PR.

SteeringAngle now includes limits, and units of measure (in degrees)
Added DoorOpen status for all 5 doors (didn't check if the hood needed another switch, yet)
Update CruiseButtons to be mutually exclusive

### DIFF
--- a/subaru_outback_2016_eyesight.dbc
+++ b/subaru_outback_2016_eyesight.dbc
@@ -36,19 +36,34 @@ BS_:
 BU_: XXX
 
 
-BO_ 884 Door_Status: 8 XXX
+BO_ 884 Doors_Status: 8 XXX
  SG_ DoorOpen_FD : 24|1@0+ (1,0) [0|1] "Front Driver Door" XXX
  SG_ DoorOpen_FP : 25|1@0+ (1,0) [0|1] "Front Passenger Door" XXX
  SG_ DoorOpen_RP : 26|1@0+ (1,0) [0|1] "Rear Passenger Door" XXX
  SG_ DoorOpen_RD : 27|1@0+ (1,0) [0|1] "Rear Driver Door" XXX
  SG_ DoorOpen_Hatch : 28|1@0+ (1,0) [0|1] "Rear Hatch" XXX
 
-BO_ 324 ACC: 7 XXX
+BO_ 324 NEW_MSG_3: 7 XXX
  SG_ BrakeApplied : 8|1@0+ (1,0) [0|0] "" XXX
- SG_ CruiseSetButton : 3|1@0+ (1,0) [0|0] "" XXX
- SG_ CruiseResumeButton : 4|1@0+ (1,0) [0|0] "" XXX
  SG_ CruiseStatus : 48|1@0+ (1,0) [0|0] "" XXX
  SG_ BrakeStatus : 51|1@0+ (1,0) [0|0] "" XXX
+ SG_ C : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cr : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cru : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Crui : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cruis : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cruise : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cruise_ : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cruise_S : 4|2@0+ (1,0) [0|3] "" XXX
+ SG_ Cruise_Se : 4|2@0+ (1,0) [0|3] "" XXX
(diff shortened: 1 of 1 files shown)


subaru_outback_2016_eyesight.dbc
```diff
@@ -36,17 +36,32 @@ BS_:
 BU_: XXX
 
 
+BO_ 884 DoorStatus: 8 XXX
+ SG_ DoorOpenFD : 24|1@0+ (1,0) [0|1] "Front Driver Door" XXX
+ SG_ DoorOpenFP : 25|1@0+ (1,0) [0|1] "Front Passenger Door" XXX
+ SG_ DoorOpenRP : 26|1@0+ (1,0) [0|1] "Rear Passenger Door" XXX
+ SG_ DoorOpenRD : 27|1@0+ (1,0) [0|1] "Rear Driver Door" XXX
+ SG_ DoorOpenHatch : 28|1@0+ (1,0) [0|1] "Rear Hatch" XXX
```
rbiasini: I believe it's good practice to just put the unit within quotes. In this case the signal would be unitless ("").
You can add comment separately at the bottom of the file (see other dbcs as examples)... but in this case the signal name itself is very self-explanatory.

subaru_outback_2016_eyesight.dbc
```diff
@@ -36,17 +36,32 @@ BS_:
...
+
+BO_ 324 CruiseControl: 7 XXX
+ SG_ BrakeApplied : 8|1@0+ (1,0) [0|0] "" XXX
+ SG_ CruiseEnabled : 48|1@0+ (1,0) [0|0] "" XXX
+ SG_ BrakeStatus : 51|1@0+ (1,0) [0|0] "" XXX
+ SG_ CruiseButtons : 4|2@0+ (1,0) [0|3] "" XXX
```
rbiasini: if you have confirmed that the cruise buttons are mutually exclusive, then you should add the list of values (VAL_) at the end of the dbc, like you did for brake applied etc...

subaru_outback_2016_eyesight.dbc
```diff
@@ -36,17 +36,32 @@ BS_:
...
  SG_ ThrottlePosition : 7|8@0+ (1,0) [0|255] "" XXX
 
 BO_ 209 NEW_MSG_1: 8 XXX
  SG_ BrakePosition : 23|8@0+ (1,0) [0|255] "" XXX
 
-BO_ 2 NEW_MSG_2: 8 XXX
- SG_ Steering_Angle : 7|16@0- (1,0) [0|65535] "" XXX
-
+BO_ 2 Steering: 8 XXX
+ SG_ SteeringAngle : 7|16@0- (1,0) [-500|500] "degree" XXX
```
rbiasini: here the factor should be 0.1, not 1, right? so basically:
SG_ SteeringAngle : 7|16@0- (0.1,0) [-500|500] "degree" XXX

subaru_outback_2016_eyesight.dbc
palmerit: Thanks for the feedback. These were originally included when the 2 buttons were separate.  I'll have to go back out and get the values again.  That will probably be tonight or tomorrow.

I would just get the numbers from a previous commit, but it has occured to me that the buttons increment by either 1 or 5 mph depending on how "deeply" you press the button.  So I'll want to make sure I have all of the possible values.

rbiasini: yeah, let's get this sorted out and then we'll merge.

palmerit: @rbiasini This should be sorted out now. I also added turn signal indicators.  Forgot them in the original PR.

rbiasini: @palmerit , cool. So just to confirm, what happens if you press set and resume buttons at the same time?

subaru_outback_2016_eyesight.dbc
```diff
@@ -36,17 +36,38 @@ BS_:
...
 
 
 
 
+VAL_ 324 BrakeApplied 1 "On" 0 "Off" ;
+VAL_ 324 CruiseEnabled 1 "On" 0 "Off" ;
+VAL_ 324 BrakeStatus 1 "On" 0 "Off" ;
+VAL_ 324 CruiseButtons 2 "Set" 1 "Resume" ;
 
+VAL_ 624 TurnSignal 2 "Left" 1 "Right"
```
rbiasini: missing ; at the end

subaru_outback_2016_eyesight.dbc
palmerit: @rbiasini it's a rocker style switch,  I don't think you physically can.

See this image for reference:  (Not my image)
http://www.cars101.com/subaru/outback/outback17-steering-wheel9.jpg

subaru_outback_2016_eyesight.dbc
palmerit: Whoopsies.  Muh baaad.  Fixed

subaru_outback_2016_eyesight.dbc
rbiasini: great, I'll merge

PR #17, opened 2017-08-30
Several ADAS updates to 2017 CR-V dbc

energee:


honda_crv_2017_can.dbc
```diff
@@ -2,255 +2,306 @@ VERSION ""
...
 
 BO_ 344 POWERTRAIN_DATA: 8 PCM
- SG_ TRANSMISSION_SPEED : 7|16@0+ (0.002759506,0) [0|70] "m/s"  NEO
- SG_ ENGINE_RPM : 23|16@0+ (1,0) [0|15000] "rpm"  NEO
- SG_ TRANSMISSION_SPEED2 : 39|16@0+ (0.002759506,0) [0|70] "m/s"  NEO
- SG_ COUNTER : 61|2@0+ (1,0) [0|3] ""  NEO
- SG_ CHECKSUM : 59|4@0+ (1,0) [0|3] ""  NEO
+ SG_ XMISSION_SPEED : 7|16@0+ (0.002759506,0) [0|70] "m/s" NEO
+ SG_ ENGINE_RPM : 23|16@0+ (1,0) [0|15000] "rpm" NEO
+ SG_ XMISSION_SPEED2 : 39|16@0+ (0.1723,0) [1|0] "" NEO
```
rbiasini: What do you think the unit is here? not m/s anymore? What about the limits?

energee: I'll revert that to what we have in the other Hondas, I was doing something else with it

rbiasini:
@energee , sounds great. does this imply any required change to openpilot repo?

energee:
No, I don't have any code in the comma repo yet for this car, but will be coming soon!

energee:
2017 != 2016

rbiasini:
@energee , right, this is for 2017. ACC should be just like the 2016+ Civic. 

I would just rename this dbc to honda_crv_touring_2017_can.dbc to be consistent with other dbc file names, other than that it looks good to me.

energee:
I don't have the touring trim, i have the EX, should I still rename it?

rbiasini:
yes. It's better to have the trim in the dbc file name. Let's rename it as honda_crv_ex_2017_can.dbc
For the near future, I'm thinking about a cleaner dbc file name convention... I'll accept suggestions on Slack

PR #18, opened 2017-09-09
Add chffr metrics 

andyh2:
Chffr metrics map signals to canonical metrics we display in Chffr 2.0. They are stored in the top-level DBC comment field `CM_ {comment}`. 

Format: `CHFFR_METRIC {MSG_ADDRESS} {SIGNAL_NAME} {METRIC_NAME} {FACTOR} {OFFSET}` 

Current metric names:
STEER_ANGLE: number in [0, 360) where 180 is centered wheel position. 
ENGINE_RPM: as reported by car.
ENGINE_TEMPERATURE: engine temperature, celsius. 

PS seems my editor stripped some trailing spaces.

PR #19, opened 2017-09-09
Latest dbc updates, including chffr metric for engine temp

energee:
Adds chffr metric for engine temp and structural updates, minor tweaks to signals


geohot:
See #18 for the new chffr metrics format

energee:
I think this is ready to get merged

PR #22, opened 2017-09-22
Create vw_golf_mk4.dbc

jessrussell:


rbiasini:
Awesome! is this dbc file covering the whole '97-'04 models or is this for a particular MY? Also, does the vehicle trim matter? @jessrussell , I'll wait on your answer about the need of specifying a year/trim in the dbc file name before merging.
(Optional for later) would be great to have it translated in English.

PR #24, opened 2017-10-01
Create vw_mqb_2010.dbc

jessrussell:


AntonBaburov:
Hello, jessrussel. Your vw_mqb_2010.dbc file looks very usefull for Skoda superb 2017. But some CAN IDs don't present in this file while i can see them in Skoda bus. Maybe you have vw MQB dbc-file for other CAN-bus? This file is in German, so i can help with translating if you are interested.

PR #31, opened 2017-10-04
Create gm_global_a_lowspeed_1818125.dbc

jessrussell:


TheMutley:
Hey @jessrussell , thanks for this DBC, greatly appreciated!! you think you could the same for all the other CAN network from GM ?

TheMutley:
I'm currently modding this DBC in order to incorporate the UseGMParameterIDs and GMLAN extended ID-Format so that the ID is encoded correctly.

rfrowe:
@TheMutley @jessrussell Do you happen to have a reference for these signals? I'd love to get the full names for them, since they're a bit hard to decipher in their abbreviated form.

PR #33, opened 2017-10-05
More updates to support work on 2017 CR-V port

energee:
Added and fixed checksum and counters, parser validation was failing in some places
Adjusted wheel speeds for cr-v
Fixed some issues with packet sizes
Added better comments on some of the newer messages


honda_crv_ex_2017_can.dbc
```diff
@@ -109,73 +116,85 @@ BO_ 401 GEARBOX: 8 PCM
...
 BO_ 464 WHEEL_SPEEDS: 8 VSA
- SG_ WHEEL_SPEED_FL : 7|15@0+ (0.002759506,0) [0|70] "m/s" NEO
- SG_ WHEEL_SPEED_FR : 8|15@0+ (0.002759506,0) [0|70] "m/s" NEO
- SG_ WHEEL_SPEED_RL : 25|15@0+ (0.002759506,0) [0|70] "m/s" NEO
- SG_ WHEEL_SPEED_RR : 42|15@0+ (0.002759506,0) [0|70] "m/s" NEO
- SG_ CHECKSUM : 59|4@0+ (1,0) [0|3] "" NEO
+ SG_ WHEEL_SPEED_FR : 8|15@0+ (0.00286,0) [0|70] "m/s" NEO
+ SG_ WHEEL_SPEED_RL : 25|15@0+ (0.00286,0) [0|70] "m/s" NEO
+ SG_ WHEEL_SPEED_RR : 42|15@0+ (0.00286,0) [0|70] "m/s" NEO
+ SG_ WHEEL_SPEED_FL : 7|15@0+ (0.00286,0) [0|70] "m/s" NEO
```
rbiasini: how are you finding the factor? trying to match some gps speed or comparing with other speed signals?

energee: Comparing with dash speed, it was off by about 1-3 miles an hour... Any recommendations on getting this exact?

energee: Also, wouldn't it be better to have this value the same across hondas and use a multiplier based on wheel size in openpilot?

rbiasini: yeah, dash speed is always fictitiously increased (by law, so people drive at speed limits conservatively)
I'm sure the conversion factor (wheel radius) is probably already done somewhere in the honda code, otherwise the discrepancy would be bigger: aprrox tire diam for honda civic is 25.5", while for CRV is 29.1". If you want this speed to have physical meaning, I would probably recommend to keep the factor as it was. Not a big deal either way... just FYI

energee: ok thanks, in that case, i'll revert this in the next update I send. I will look for somewhere else to adjust this factor to match the dash since its a little confusing when setting at a value and not seeing it reflected on neo speed, especially when relying on stock ACC.

PR #34, opened 2017-10-07
Fixed missing items and converted to GM extended CAN

TheMutley:


rbiasini:
@vntarasov , what do you think?

TheMutley:
@rbiasini I compiled the previous DBC that are located here with findings from @vntarasov and mine. You can merge :)

PR #37, opened 2017-10-08
Reverse Lights on 0x326

vanillagorillaa:
Added signal for when reverse lights are on/off

### DIFF
--- a/honda_civic_touring_2016_can.dbc
+++ b/honda_civic_touring_2016_can.dbc
@@ -399,12 +399,12 @@ BO_TX_BU_ 829 : NEO,ADAS;
 BO_TX_BU_ 862 : NEO,ADAS;
 BO_TX_BU_ 927 : NEO,ADAS;
 
-
 CM_ SG_ 401 GEAR "10 = reverse, 11 = transition";
 CM_ SG_ 490 LONG_ACCEL "wheel speed derivative, noisy and zero snapping";
 CM_ SG_ 780 CRUISE_SPEED "255 = no speed";
 CM_ SG_ 804 CRUISE_SPEED_PCM "255 = no speed";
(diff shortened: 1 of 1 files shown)


honda_civic_touring_2016_can.dbc
```diff
@@ -274,6 +274,7 @@ BO_ 806 SCM_FEEDBACK: 8 SCM
  SG_ MAIN_ON : 28|1@0+ (1,0) [0|1] "" NEO
  SG_ RIGHT_BLINKER : 27|1@0+ (1,0) [0|1] "" NEO
  SG_ LEFT_BLINKER : 26|1@0+ (1,0) [0|1] "" NEO
+ SG_ REVERSE_LIGHT : 18|1@0+ (1,0) [0|1] "" NEO
  SG_ COUNTER : 61|2@0+ (1,0) [0|3] "" NEO
  SG_ CHECKSUM : 59|4@0+ (1,0) [0|3] "" NEO
```
rbiasini: How do you know this is reverse lights and not reverse gear engaged?

vanillagorillaa: Just a theory since this in on a message with other indicator lights and not a message for gearbox. I don't currently have a set up to set messages to the bus to confirm though

rbiasini: The msg comes from steering column, which is where buttons and turns levers are. Seems strange it sends info about reverse gear or lights. Anyway, if you saw the correlation between reverse gear it's sufficient and better than nothing. Can you just add a clarification comment at the end of the dbc to highlight the uncertainty?

vanillagorillaa: Added

PR #39, opened 2017-10-08
ECON Mode On 0x221

vanillagorillaa:
Defined ECO Mode on

### DIFF
--- a/honda_civic_touring_2016_can.dbc
+++ b/honda_civic_touring_2016_can.dbc
@@ -198,6 +198,8 @@ BO_ 513 GAS_SENSOR: 5 INTERCEPTOR
  SG_ CHECKSUM : 35|4@0+ (1,0) [0|3] "" NEO
 
 BO_ 545 XXX_10: 6 XXX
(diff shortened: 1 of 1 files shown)


honda_civic_touring_2016_can.dbc
```diff
@@ -198,6 +198,8 @@ BO_ 513 GAS_SENSOR: 5 INTERCEPTOR
  SG_ CHECKSUM : 35|4@0+ (1,0) [0|3] "" NEO
 
 BO_ 545 XXX_10: 6 XXX
+ SG_ ECON_ON_2 : 37|2@0+ (1,0) [0|3] "" NEO
```
rbiasini: 2 bits? What value corresponds to eco mode on? I would add a value table at the end of dbc to say what 0,1,2,3 mean.

vanillagorillaa:
I haven't seen any other value besides 0=Off and 3=On. Would you like value table added or just remove for now since it already has another message?

rbiasini:
Value table is good, just 3 and 0

vanillagorillaa:
Added value table

PR #42, opened 2017-10-13
Removed wrong gear signal

vanillagorillaa:
Removed wrong gear signal and renamed correct signal for Gearbox

### DIFF
--- a/honda_civic_touring_2016_can.dbc
+++ b/honda_civic_touring_2016_can.dbc
@@ -100,8 +100,7 @@ BO_ 399 STEER_STATUS: 7 EPS
 
 BO_ 401 GEARBOX: 8 PCM
  SG_ GEAR_SHIFTER : 5|6@0+ (1,0) [0|63] "" NEO
- SG_ GEAR : 35|4@0+ (1,0) [0|15] "" NEO
- SG_ GEAR_2 : 43|4@0+ (1,0) [0|15] "" NEO
+ SG_ GEAR : 43|4@0+ (1,0) [0|15] "" NEO
  SG_ COUNTER : 61|2@0+ (1,0) [0|3] "" NEO
  SG_ CHECKSUM : 59|4@0+ (1,0) [0|3] "" NEO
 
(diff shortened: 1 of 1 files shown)


vanillagorillaa:
I have not, looking at cabana it looks like the signal that was originally defined as gear is actually a larger bit message not related to the gear. I can try this weekend.

vanillagorillaa:
I can send you a link to a recent drive via slack if you want to take a look

energee:
Its possible that it was incorrect not failing due to the value it was before so I would just take a look closer

rbiasini:
actually this change seems fine: in carstate the gear value isn't set for the Civic (the wrong dbc signal seems a leftover from the Acura)

```self.gear = 0  # TODO: civic has CVT... needs rev engineering```

@vanillagorillaa, if you go through the openpilot code and check that this change will not break it, I'm fine with the change.

vanillagorillaa:
I will check tonight/weekend to confirm that no issues arise

geohot:
Will merge after test. Is your previous PR still valid?

vanillagorillaa:
@geohot @rbiasini I forked Openpilot to my own and changed ("GEAR", 0x191, 0) to ("GEAR_2", 0x191, 0) on line 63 and self.gear = 0  # TODO: civic has CVT... needs rev engineering to self.gear_2 = 0  # TODO: civic has CVT... needs rev engineering and used my PR #40 and pulled the opendbc once loaded to neo. Did not have a problem with the short drive. I assume this accomplishes the same thing.

energee:
My CR-V is probably like this too, I'll check it out

vanillagorillaa:
Quick update. I've been testing a little bit more on my own fork of openpilot/opendbc. I actually went ahead and defined the updated gear signal in carstate.py. I've been driving with it and have had no errors. Looks to me that adding this PR would be ok. I might even be making a PR in Openpilot to update carstate.py.

rbiasini:
@vanillagorillaa , great. 
About the OP change: the gear there is intended to be the actual gear number rather than the gear shifter position.
Originally it was used to not wind up the integrator during gear changes in longitudinal control (It's unused now). 
I would update OP if you can find the signal that represents the actual gear number.

vanillagorillaa:
@rbiasini ah got it, that would explain the need for the reverse eng. I'll leave it as if in OP, doubt I have the systems understanding to RE the gear.

PR #48, opened 2017-11-01
Created toyota iQ base plus reversed some signals

alessbelli:
Still WIP, could be used for panda+chffr support & display data in app.


rbiasini:
@alessbelli  so on the same can bus there are obstacle info (from radar) and powertrain info (such as steering angle etc...)? I would expect them to be on separate buses.

alessbelli:
@rbiasini there is no radar to my current knowledge, but yes, I have found some places where messages contained both dynamics info and other unrelated sensors. My guess would be that some signals get sent with a lower resolution and rate to the main board display (where I can see speed, engine rpm and whether my seatbelt is on for instance) while other more frequent and precise signals about the onboard train are used by the control unit to evaluate fuel injection etc.

alessbelli:
I haven't seen any other CAN bus elsewhere in the car, but I'll be happy to reverse signals from anywhere else if you tell me where to find this other eventual port

toyota_iQ_2009_can.dbc
```diff
@@ -0,0 +1,432 @@
...
+    BU_EV_REL_
+    BU_BO_REL_
+    SG_MUL_VAL_
+
+BS_:
+
+BU_: XXX
+
+
+BO_ 528 TRACK_A_0: 8 XXX
```
rbiasini: @alessbelli , can you remove all those radar signals? basically from here to line 256. Since this is a new dbc file, you should only report the messages and signals that you were able to reverse engineer.

alessbelli: You're entirely right, I think I started with a toyota base, but I forgot to remove non-matching signals since they did not appear in cabana. Cleaning it right away

alessbelli:
@rbiasini just cleaned up a bit, sorry for the mess earlier!

rbiasini:
@alessbelli , no worries. Looks good now.

PR #50, opened 2017-11-05
add   toyota_prius_2010_cam.dbc

robinsonmas:
Please add toyota_prius_2010_cam.dbc


rbiasini:
@robinsonmas , why the 2 dbc files look the same? I would expect different info on different can buses.

robinsonmas:
yes , please  delete file" toyota_prius_2010_pt.dbc" ..  Thanks

.gitignore
```diff
@@ -1,2 +0,0 @@
-*.pyc
-.*.swp
```
rbiasini: you might want to revert this change :)

toyota_prius_2010_cam.dbc
```diff
@@ -0,0 +1,203 @@
+VERSION ""
```
rbiasini: can you please change the name into toyota_prius_2010_pt.dbc? so it's consistent with other dbc file names. Other than that (and the comment above), it looks good.

robinsonmas:
the file toyota_prius_2010_pt.dbc ,,I deleted it, the correct file is toyota_prius_2010_cam.dbc. please can you check it , Thanks :-)

PR #59, opened 2017-12-18
I'd like to contribute my dbc files

Brainstem2000:
I'd like to contribute the Ford_2017_F150_Lariat dbc file.


Brainstem2000:
I've added the dbc for a 2017 Lincoln Continental

vanillagorillaa:
You will probably want to have signals/ID's defined first

Brainstem2000:
@vanillagorillaa - thanks. I was not sure what the "proper" procedure was. If merely contributing a fresh dbc was a positive contribution, I was wanting to do that. I'll debug and attempt to determine signals and IDs and then send a note as such back here.

PR #61, opened 2017-12-21
Added EPB and Brake_Hold

joeljacobs:
Searched through Cabana with @energee to find proper signals for EPB and Brake_Hold. We found them.


honda_odyssey_exl_2018.dbc
```diff
@@ -111,6 +114,11 @@ BO_ 432 STANDSTILL: 7 VSA
  SG_ CHECKSUM : 51|4@0+ (1,0) [0|3] "" EON
 
 BO_ 450 XXX_6: 8 XXX
+ SG_ EPB_Brake_and_Pull_Indicator : 6|1@0+ (1,0) [0|1] "" XXX
```
rbiasini: @joeljacobs can you use CAPITAL notation here too? other than that it looks good

PR #64, opened 2017-12-31
Add Hyundai i20 Magna 2013 DBC 

SafwanAhmad:
Add Hyundai i20 Magna 2013 DBC.


espes:
The file is identical to hyundai_i30_2014.dbc?

PR #72, opened 2018-02-15
Added based on Nissan Murano 2009 SL tests

magillus:
Please direct me to best names to use for some messages/values that could be supported in chffr.
What would be the car's fingerprint (list of fields) from basic white Panda?
Not all messages are decoded.

A lot of the messages are coming from https://github.com/Knio/carhack/blob/master/Cars/Nissan.markdown


pd0wm:
For chffr signals look at these special comments: https://github.com/commaai/opendbc/blob/master/toyota_prius_2017_pt_generated.dbc#L223

I'm closing this PR now because it looks stale. Feel free to reopen when more signals are reverse engineered!

PR #80, opened 2018-03-08
New 2012 Camry Hybrid dbc

Chrispreilly:
Started with the rav4hybrid dbc and pulled out all the signals that aren't used on the camry


pd0wm:
Can you run the generator so the generated file is also in the PR?

PR #86, opened 2018-03-17
Fix Checksum errors for CH-R

vanillagorillaa:
message is actually 8 bytes


vanillagorillaa:
Fix the generator file so the generated file will be correct

PR #87, opened 2018-03-23
Hyundai Ioniq 2017 HEV LKAS System.

quarksteilchen:


quarksteilchen:
hi! @rbiasini 
it's the 2017 model of the car.

by the way: if more information is needed, i can look for more specific information in the data stream of the can-bus.

more information about the changes:
https://docs.google.com/spreadsheets/d/1FP6YSVUwHDrYLlhCJwmzsHnozszrKLjCjVWLE5_bsi8/edit?usp=sharing

pd0wm:
What message are different from this DBC https://github.com/commaai/opendbc/blob/master/hyundai_2015_ccan.dbc?

We should probably refactor the common messages in a separate DBC and use the generator.

rbiasini:
no response

PR #91, opened 2018-05-02
Add Highlander and Avalon DBC's

vanillagorillaa:


rbiasini:
@vanillagorillaa need to run the dbc generator script. Other than that looks good.

vanillagorillaa:
@rbiasini opps! forgot about those. Should be good now. Just initial DBC commit for now. May need some tweaks as ports move forward

PR #94, opened 2018-05-12
Add's back import file for honda pedal's

vanillagorillaa:
add's back import file for honda's that use comma dbc for comma pedal. Does not add back to Bosch cars' dbcs


energee:
Why do we have interceptor on Honda Civic?

PR #96, opened 2018-05-22
Changes for Pedal Threshold

joeljacobs:
Altered Pedal's offset to -120 after testing with Odyssey, RDX and 4 different Pedals.
This corresponds with this [OpenPilot PR ](https://github.com/commaai/openpilot/pull/251)
and this [Panda PR](https://github.com/commaai/panda/pull/117)


rbiasini:
Obsolete, also I don't think this is needed.

PR #97, opened 2018-06-05
Update generator.py

dekerr:
-improved readability
-files now build in the order they appear instead of reverse
-added "with" statements for error handling


pd0wm:
Thanks for the pull request!

Currently the `CM_ "IMPORT _xxx.dbc"` headers remain in the generated files. Could you update your pull request to remove those when inserting the original DBC file?

pd0wm:
Looks good!

There is a small bug in the code (see comment), but if you fix that I will merge the PR.

generator/generator.py
```diff
@@ -2,36 +2,51 @@
...
+  
+def create_dbc(filename):
+    dbc_file_in = read_dbc(filename) 
+
+    includes = include_pattern.findall(dbc_file_in)
+ 
+    output_filename = filename.replace('.dbc', '_generated.dbc')
+    output_file_location = os.path.join(generator_path, output_filename)
+    
+    if os.isfile(output_file_location):
```
pd0wm: this should be os.path.isfile

dekerr:
Everything should be working now. Thank you so much for your help and for your time.

PR #105, opened 2018-08-07
Refactored clarity DBC to match style of other Hondas

energee:


rbiasini:
@energee since the Clarity is still WIP, I would prefer to keep the Clarity dbc file separated until the end. Sounds good? I'm worried that there are more messages (other than the brake cmd) that are not common with the rest of the Hondas.

energee:
Sounds good

PR #108, opened 2018-08-23
Toyota Pedal Support

wocsor:
add pedal signal with toyota scaling to all dbc


rbiasini:
@wocsor , what are the `.DS_store` files? not needed, right?

wocsor:
Not at all. Those are because I used a Mac and forgot to set up gitignore.. my bad

rbiasini:
Cool. Wil merge when they are removed. Thanks!

wocsor:
That should be it. Sorry about those

PR #113, opened 2018-09-17
Vector syntax fix

ngbrown:
Fix syntax problems with DBC files that prevented loading into Vector CANdb++.

A better `generator.py` which correctly orders the output .dbc file was added.


ngbrown:
@pd0wm and @dekerr, you both contributed to the `generator.py`.  I think these changes to make generated `.dbc` file is valid for other tools are valuable improvements .  Can you review?

dekerr:
cool. I actually wanted to work on doing this, but I decided to make a VAL_ parser for openpilot instead. 
Two thoughts:
-A second layer of sorting by message address
-Maybe keep a list of import sources at the top?

All in all I think it's a good start to cleaning things up. Nice work.

pd0wm:
Sorry, missed the notification on this PR.

What exactly is the format that vector expects? From the code I see that it assumes some order of the sections? Is there some documentation about the expected order?

PR #122, opened 2018-11-09
Adds dbc for 2019 Honda Insight

energee:


energee:
@rbiasini This is ready to be merged!

PR #123, opened 2018-11-26
Add GM FCW dashboard alert

Jamezz:
These are the 2 bits that control the driver forward collision HUD and audible collision alert.

I can provide a cabana of this if proof is needed, PM me on slack (someone else's capture, don't want to disclose without permission)

+ @vntarasov


gm_global_a_powertrain.dbc
```diff
@@ -180,6 +180,7 @@ BO_ 880 ASCMActiveCruiseControlStatus: 6 K124_ASCM
  SG_ ACCGapLevel : 21|2@0+ (1,0) [0|0] ""  NEO
  SG_ ACCResumeButton : 1|1@0+ (1,0) [0|0] ""  NEO
  SG_ ACCCmdActive : 23|1@0+ (1,0) [0|0] ""  NEO
+ SG_ FCWAlert : 40|2@1+ (1,0) [0|3] "" XXX
```
rbiasini: merged, but just realized this should have same endianess as other signals. @Jamezz

PR #128, opened 2018-12-11
GM Object Front Cam Signals

CryptoKylan:
This brings over all the front camera messages  The vast majority of these messages populate data, at least on my vehicle

the pedestrian tracking works too.  lots of cool info in here

hope this looks good.  compiles and works fine for me on Openpilot and Cabana


rbiasini:
looks good!

PR #131, opened 2018-12-26
Split signal for mini car and distance button

arne182:


rbiasini:
@arne182 this can be closed, right?

arne182:
Yes

PR #132, opened 2018-12-27
Chrysler message to play an audible beep & ACC cancel

adhintz:
Sending the AUDIBLE_BEEP works reliably with no issues.

Sending the ACC_CANCEL with the previously seen WHEEL_BUTTONS counter + 1  sometimes works, but not reliably.


adhintz:
looks like the old commits are showing up as new. I'll fix it and make a new PR

PR #133, opened 2018-12-27
Chrysler message to play an audible beep & ACC cancel

adhintz:
Still a big commit log, but it merges cleanly.

Sending the AUDIBLE_BEEP works reliably with no issues.

Sending the ACC_CANCEL with the previously seen WHEEL_BUTTONS counter + 1  sometimes works, but not reliably.


adhintz:
Let me know if you want to rebase or squash commits so there's not a bit list of commits.

PR #147, opened 2019-03-13
Add files via upload

AdasCoder:
Delphi ESR as used in passenger vehicles from Volvo, such as the V60, S60 and V40.

### DIFF
--- a/ESR.dbc
+++ b/ESR.dbc
@@ -1,0 +1,1080 @@
+VERSION ""
+
+
+NS_ : 
+	NS_DESC_
+	CM_
+	BA_DEF_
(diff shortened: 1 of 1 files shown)


AdasCoder:
This radar is normally used in a sensor fusion system together with a camera. However, the radar spits out up to 64 targets which are later to be processed together with the targets from the camera.
Vehicles using this radar are built on Volvo EUCD platform.

rbiasini:
thanks!

EwoutH:
(for historic reference: This PR replaces https://github.com/commaai/openpilot/pull/565)

PR #149, opened 2019-03-25
New camry steer message

bugsy924:
More accuracy and less latency


legonigel:
This value is listed as signed, but has a range of [0|65535]. Please correct this. Either the range or the signed is wrong.

rbiasini:
@bugsy924  nice. This must be the good signal to use for closed loop control. Btw, in Subaru I found that it sometimes shows a weird offset on startup (compared to the old STEER_ANGLE_SENSOR). Does it happen in the Camry too?

bugsy924:
I haven't had that issue on the Outback
On the Camry, it would add -20 degrees for 4/5 startups, but now it only does it every 1/30. I'm not sure if it's a new car thing

rbiasini:
On the Subaru I see 2-3 deg random offset. Not a big deal since we are soon going to release the auto-calibration feature.

PR #153, opened 2019-04-03
Fixes to make DBC files compatible with CANdb++ Editor

msalau:
Hi,

Here are few fixes to make DBC files compatible with Vector CANdb++ Editor.
The generator script was tuned to merge several DBC files and keep ordering correct.
Approach may be not the best, but it works with this set of DBC files. A better solution may be to use [canmatrix](https://github.com/ebroecker/canmatrix) to do this job.

Best regards,
Maksim


msalau:
Seems like there is a better fix for `generator.py` in PR #113

PR #155, opened 2019-04-05
Added SPORT_ON message for Corolla

ShaneSmiskol:


rbiasini:
you also need to run the `generator.py` script. Other than that it looks good.

ShaneSmiskol:
@rbiasini Am I good to add parentheses in generator around the print statements? Doesn't run in python 3 and it will still work for python 2.x

ShaneSmiskol:
All done!

PR #162, opened 2019-05-08
Lexus NXH DBC.

kengggg:


rbiasini:
@kengggg , in the autogenerated file you should only put the messages that are your car specifc, then run `generate.py` to combine common messages (in `_toyota.dbc`) with `lexus_nx_hybrid_2018_pt.dbc`

rbiasini:
@kengggg , let me know if you plan in updating this. I'll close it otherwise. Thanks.

PR #170, opened 2019-07-02
Lexus RX 350 DBC

wocsor:


generator/toyota/lexus_rx_350_2016_pt.dbc
```diff
@@ -0,0 +1,74 @@
...
+
+CM_ SG_ 550 BRAKE_PRESSURE "seems prop to pedal force";
+CM_ SG_ 550 BRAKE_POSITION "seems proportional to pedal displacement, unclear the max value of 0x1c8";
+CM_ SG_ 610 TYPE "seems 1 on Corolla, 0 on all others";
+VAL_ 610 IPAS_STATE 5 "override" 3 "enabled" 1 "disabled" ;
+VAL_ 610 LKA_STATE 25 "temporary_fault" 9 "temporary_fault2" 5 "active" 1 "standby" ;
+VAL_ 956 SPORT_ON 0 "off" 1 "on" ;
+VAL_ 956 GEAR 0 "D" 1 "S" 8 "N" 16 "R" 32 "P" ;
+VAL_ 956 ECON_ON 0 "off" 1 "on" ;
+CM_ "IMPORT _toyota_2017.dbc"
```
rbiasini: import should go at the top

generator/toyota/lexus_rx_350_2016_pt.dbc
```diff
@@ -0,0 +1,74 @@
+VERSION ""
+
+
+NS_ :
```
rbiasini: this is already in _toyota_2017.dbc

rbiasini:
@wocsor what are the differences between this dbc and the dbc from some other Toyota? wondering if we need a new dbc file

wocsor:
it's a lot like the rav4 DBC but it has the brake messages of a RX450. some of the signals in the RX450 do not line up with this DBC, though. if i remember correctly, it was the gas pedal and brake messages

PR #172, opened 2019-07-12
Correct Message ID on LKAS11

TK211X:
TechInfo Diagram denote MFC unit as having BCA-R function. Profile and message characteristics match.

https://i.imgur.com/vANN6Ho.png


TK211X:
Will seek and replace other prevalent code from old message.

This message cannot be static. It's a safety message and must be forwarded.

rbiasini:
ok, thanks. I'll wait for the change in openpilot as well

TK211X:
How's that?

PR #173, opened 2019-07-18
Nissan: Added nissan_2017.dbc

bugsy924:
Tested on a Nissan X-Trail 2017

### DIFF
--- a/nissan_2017.dbc
+++ b/nissan_2017.dbc
@@ -1,0 +1,184 @@
+VERSION ""
+
+
+NS_ :
+    NS_DESC_
+    CM_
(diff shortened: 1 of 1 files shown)


rbiasini:
merged but it would be better to be consistent on endianess.
even within the same message, some signals are big endian and some low endian.

bugsy924:
Will fix, should be all big endian

On Jul 19, 2019 5:51 AM, rbiasini <notifications@github.com> wrote:

merged but it would be better to be consistent on endianess.
even within the same message, some signals are big endian and some low endian.

PR #175, opened 2019-08-02
Add DAW

TK211X:
Found Driver Attention Warning

### DIFF
--- a/hyundai_kia_generic.dbc
+++ b/hyundai_kia_generic.dbc
@@ -991,7 +991,7 @@ BO_ 1342 LKAS12: 6 LDWS_LKAS
  SG_ CF_LkasTsrSpeed_Display_Navi : 24|8@1+ (1.0,0.0) [0.0|255.0] ""  BCM,CLU
  SG_ CF_Lkas_TsrAddinfo_Display : 32|2@1+ (1.0,0.0) [0.0|3.0] ""  CLU
  SG_ CF_LkasDawStatus : 40|3@1+ (1,0) [0|7] "" CLU
- SG_ CF_Lkas_Daw_USM : 39|3@0+ (1,0) [0|7] "" CLU
+ SG_ CF_Lkas_Daw_USM : 39|3@1+ (1,0) [0|7] "" CLU
 
 BO_ 1338 TMU_GW_E_01: 8 CLU
  SG_ CF_Gway_TeleReqDrLock : 0|2@1+ (1.0,0.0) [0.0|3.0] ""  BCM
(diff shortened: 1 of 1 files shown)


rbiasini:
Would be great if you could add VAL or CM lines to document what they are.

TK211X:
USM = User select-able Mode.

Status = Status.

I had a user toggle modes in their DAW equipped vehicle so I can't test in depth. 
https://i.imgur.com/XAXCPav.png

So we know the USM modes 1:OFF 3:Normal 4:High

For Status we only know 0:OFF 5:Ready/Idle Since I don't have a cabana of it in action on the road to inspect yet.

https://i.imgur.com/YVILPwo.png

I would assume based on the image above that Status mode might start with mode 5 as attentive and work it's way down to 1 for inattentive and the CLU handles the time since last break along when deciding to alert the driver it's time to take a break.

This is all I can infer so far:

https://i.imgur.com/6EV6OCH.png

Can't think of another unit that would listen to it other than CLU so it's all I added.

TK211X:
In stock operation the MFC system sees if the driver is exhibiting inattentive driving practices through steering input and (location relative to lane?). Possible use of this message is to have Commas facial recognition meter the attentiveness of the driver and translate it to the sysmode to encourage the driver to stay attentive or let them know it's time to take a break if their distraction is habitual. Since Openpilot is now in control the Comma monitoring becomes the only way to gauge driver attention since stock can no longer reliably do so.

 It's more of a gimmick than anything but it's generated by the MFC unit so it's part of our efforts to document every message created by it.

firefox_1hOMdpVbMv

TK211X:
I also wanna say due to how USM works it's possible to force a certain mode to stay on so the user cannot change it. Comma can have it be part of driver monitoring.

hyundai_kia_generic.dbc
```diff
@@ -990,6 +990,8 @@ BO_ 1342 LKAS12: 6 LDWS_LKAS
  SG_ CF_Lkas_TsrSpeed_Display_Clu : 16|8@1+ (1.0,0.0) [0.0|255.0] ""  CLU
  SG_ CF_LkasTsrSpeed_Display_Navi : 24|8@1+ (1.0,0.0) [0.0|255.0] ""  BCM,CLU
  SG_ CF_Lkas_TsrAddinfo_Display : 32|2@1+ (1.0,0.0) [0.0|3.0] ""  CLU
+ SG_ CF_LkasDawStatus : 40|3@1+ (1,0) [0|7] "" CLU
+ SG_ CF_Lkas_Daw_USM : 39|3@0+ (1,0) [0|7] "" CLU
```
rbiasini: big endian?

rbiasini: mmm... CF_LkasDawStatus is still marked as little endian, while CF_Lkas_Daw_USM is big endian. You sure?

rbiasini: why resolved? CF_Lkas_Daw_USM should be little endian.

bbsusheelkumar: All the CAN data in my hyundai car was found to be little endian. I verified even this signal too in my car. I think it should be changed to little endian.

geohot:
Change to little endian?

TK211X:
It not being little endian was because that's what Cabana defaulted to when I selected it. Not by my intention. Feel free to change it, I don't care. This data is predicated already. It makes no difference to me if it gets up-streamed or not.

I don't stand to benefit from driver attention stuff. Comma does. This is why I did the research on the signals and pointed it out when I was bored. The valuable information here is not the signal but the stuff I explained in the post on Aug 2, 2019 in this thread.

TK211X:
Imagine letting something so minor drag on for half a year.

PR #179, opened 2019-08-26
Adjust 2017 Lexus IS Hybrid steer_torque_eps value

eFiniLan:
Increase steer_torque_eps value to 1.00 as it can do 1.30


rbiasini:
@eFiniLan , can you please attach a cabana plot that shows how the factor of `1` is a better over `0.66`. I don;t quite understand this comment:
> Increase steer_torque_eps value to 1.00 as it can do 1.30

eFiniLan:
> @eFiniLan , can you please attach a cabana plot that shows how the factor of `1` is a better over `0.66`. I don;t quite understand this comment:
> 
> > Increase steer_torque_eps value to 1.00 as it can do 1.30

Hi @rbiasini, isn't the factor number should match safetyParam in interface.py? (correct me if I'm wrong :disappointed: ) I found it turns better if I increase the safetyParam to 130 (was at 66). Here is the cabana for running on 1.30 (eps factor) and 130 (safetyParam)

https://my.comma.ai/cabana/?route=1dd19ceed0ee2b48%7C2019-09-03--17-23-39&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F1dd19ceed0ee2b48%2F3e10b9b698e57f584239caa9fe7c79ab_2019-09-03--17-23-39&seekTime=1

rbiasini:
@eFiniLan yeah, looked at your route and it really seems like you should use a higher factor for STEER_TORQUE_EPS. @quillford , is it the same for your car?

eFiniLan:
move to https://github.com/commaai/opendbc/pull/192

PR #180, opened 2019-09-02
Fix Spelling

arne182:


rbiasini:
thanks. You need to run generate.py though.

PR #182, opened 2019-09-04
Prius EPS torque factor

zorrobyte:
image
image

By the way, maybe update the Toyota nodsu dbc as the C-HR has the same actuator assembly as Prime. Does it impact LQR?


zorrobyte:
Even though this seems to be a better scale.. setting a matching safetyparam results in shitty behavior cc @bugsy924 

🤷‍♂

rbiasini:
@zorrobyte what do you mean with bad (euphemism) behavior? this parameter should have no effect at all, unless the user is overriding the steering himself.

zorrobyte:
@rbiasini excuse my language, I'm somewhat frustrated that even though this scale factor seems more correct (but you may want to double check as Cabana gets kinda weird when adjusting scaling factor, and maybe could be tuned in even further), making the required change to safetyparam leads to seemingly less torque and safety clipping output more, resulting in (euphemism) behaviour.

My 2019 Prius Prime has steering angle from EPS, like cars that use the nodsu/hybrid DBC. Actuator part numbers did not change between Prime years, nor does it change for C-HR, even though EPS angle didn't get added to C-HR until 2018. Prius seems to have a different torque factor than any other supported model, which I find strange as Toyota uses the same actuator assemblies between many models (but rav and Camry).

Everyone views Prius as a special case, and I wonder if Prius vs Prime further complicates things; as non-Prime prius seems to have different sR than Prime from techinfo.

Besides the point, I build tire angle sensor today :D

rbiasini:
so, my recommendation is: keep using 0.66, which is correct for probably every pre-2019 Prius.
A bit of overestimation in EPS torque is ok. Will just keep a bit more torque during a user override event, shouldn't even be noticeable. 

On the other other hand, if we change it to 0.5, then everybody who is on pre-2019 Prius will have a risk of unexpected torque limitations (system limits command if CMD is much higher than ESP).

Eventually (VIN fingerprint), we'll distinguish pre-2019 Prius form 2019+ Prius and each of them will have its own dbc file. Makes sense?

rbiasini:
@zorrobyte , closing this PR per comments above. Will be reconsidered after VIN fingerprint and when the Prime will eventually have its own dbc.

PR #190, opened 2019-09-24
Add blindspot code

arne182:
Specifiying this blind spot things defeats the purpose of the debug codes. 
This should be more generic but I have no idea what data is being returned and how it is formatted.


pd0wm:
The 1880 signal is a generic address used for all debug information. It uses the ISOTP standard for communication. So it's not really possible to capture the responses in a DBC file.

When we get around to implementing this I'll probably implement a clean ISOTP handler in openpilot that can be used for multiple things, and is not tied to the toyota carcontroller.

arne182:
Hope to see this in the next release

PR #191, opened 2019-10-03
Updates for Volkswagen support

jyoung8607:
This is the second in a series of PRs to upstream the Volkswagen, Audi, SEAT, and Škoda community port, covering all MQB platform vehicles with factory ACC.

- Update CRC and COUNTER signal names for select messages, for full incoming and outgoing message validation like Toyota and Honda. I will open a standalone PR for the [CRC code](https://github.com/jyoung8607/openpilot/blob/vw-063-devel/selfdrive/can/parser.cc#L135).
- Add a critical signal to GRA_ACC_01, necessary to support third-stalk type ACC controls for Škoda.
- Restore VIN_01 mux lost during Cabana editing.
- Add fifteen new canonical message names discovered during research and empirical testing.
- Various minor cleanup and reordering for clarity.

Free bonus: the "money message" for older Volkswagen, Audi, SEAT and Škoda vehicles based on the PQ35/PQ46/NMS platform. We're engaged in discussions with some B6/B7 Passat and Octavia owners. We've identified the message for Heading Control Assist torque, the important signals therein, and worked out the checksum (credit to Bugsy for this one). We're still a ways out for community support, much less upstream, but sharing what we have so far.

bugsy-checksum-work


jyoung8607:
Also added the VIN_1 mux message, similar to that on MQB cars, that will let us auto-identify characteristics for some cars that are hard to tell apart by fingerprint.

vw_mqb_2010.dbc
```diff
@@ -451,8 +451,8 @@ BO_ 296 Getriebe_06: 3 Getriebe_DQ_Hybrid_MQB
  SG_ GE_Testparameter_2 : 16|8@1+ (1,0) [0|255] "" Waehlhebel_MQB
 
 BO_ 173 Getriebe_11: 8 Getriebe_DQ_Hybrid_MQB
- SG_ Getriebe_11_CRC : 0|8@1+ (1,0) [0|255] "" Gateway_MQB,Motor_Diesel_MQB,Motor_Hybrid_MQB,Motor_Otto_MQB
- SG_ Getriebe_11_BZ : 8|4@1+ (1,0) [0|15] "" Gateway_MQB,Motor_Diesel_MQB,Motor_Hybrid_MQB,Motor_Otto_MQB
+ SG_ CRC : 0|8@1+ (1,0) [0|255] "" Gateway_MQB,Motor_Diesel_MQB,Motor_Hybrid_MQB,Motor_Otto_MQB
+ SG_ COUNTERXX : 8|4@1+ (1,0) [0|15] "" Gateway_MQB,Motor_Diesel_MQB,Motor_Hybrid_MQB,Motor_Otto_MQB
```
rbiasini: Typo? Also, why can't we keep the CRC/Counter signal names as they are? The suffix _CRC and _BZ seems ok. Generally, I would like to keep opendbc as close as possible to the stock dbc file.

rbiasini:
Just one comment. Other than that looks good!

vw_mqb_2010.dbc
jyoung8607: > Typo?

More of an ugly hack. I thought I had put a comment on that signal, but it appears to have been lost at some point. I'll add it back.

The Getriebe_11 message shows up on Extended CAN at 20Hz with the counter incrementing by five instead of one. I'm pretty sure what's happening is Getriebe_11 is sent on powertrain CAN at 100Hz and the J533 gateway is rate-limiting to conserve bandwidth. OP complains loudly about all the missing messages; that's the only reason I know this is happening! Renaming it prevents OP from doing COUNTER management on this particular message.

> Also, why can't we keep the CRC/Counter signal names as they are? The suffix _CRC and _BZ seems ok. Generally, I would like to keep opendbc as close as possible to the stock dbc file.

I'm in total agreement with you; I didn't want to change the names. Unfortunately the OP checksum code didn't really lend itself to the checksum/CRC signal having a dynamic name. It's been a couple months since I've had my head in that code so I can't call out specifics ATM.

Later today I'm going to file a PR that covers just the OP Volkswagen CRC support, because I think you'll want to look at it separately from the vehicle port. Perhaps we should hold this PR in suspense until then? I'd actually be very happy if we can arrive at a solution that keeps the DBC aligned with canonical message and signal names.

jyoung8607:
> Also, why can't we keep the CRC/Counter signal names as they are? The suffix _CRC and _BZ seems ok. Generally, I would like to keep opendbc as close as possible to the stock dbc file.

Per my earlier comment, see commaai/openpilot#836 for the CRC validation code. I'm more than happy to revert the CRC and COUNTER renaming if we can find a better solution on that side.

vw_mqb_2010.dbc
rbiasini: `COUNTERXX`: yeah, reasonable hack. With the comment added is good.

`_CRC`: can't you just change https://github.com/commaai/openpilot/pull/836/files#diff-56069f2917a3366b3df2428fd6ac3202R34 so that it checks if `_CRC` is in the signal name?

I'm happy to merge this PR right away, particularly if you are ok reverting the signal checksum/counter name changes (except for the one message at 5hz with skips).

jyoung8607: > `_CRC`: can't you just change https://github.com/commaai/openpilot/pull/836/files#diff-56069f2917a3366b3df2428fd6ac3202R34 so that it checks if `_CRC` is in the signal name?

It might be possible in the higher level Python code, but I'm screwed once I get down to the CANPacker code where I can't regexp search on key-value pair maps:

https://github.com/commaai/openpilot/blob/799302b842198b2cccb83c601bbe79ad1eaa5e32/selfdrive/can/packer.cc#L80

https://github.com/commaai/openpilot/blob/799302b842198b2cccb83c601bbe79ad1eaa5e32/selfdrive/can/packer.cc#L109

If I knew the message name, I could concatenate that and "_CRC" or "_BZ", but I only have the address and not the name at that point, and it's not clear to me if I can look it up. That's about the point where my eyes started bleeding and I gave up. Hints welcome.

rbiasini: I can help making the change to parser.cc and packer.cc to handle custom VW nomenclature. The parser side is super simple. The packer side a bit trickier, but should be easily done.
In the meanwhile, do you mind to revert the CRC/COUNTER name changes? Would like to merge these changes.

rbiasini:
ok, good to merge. Per our conversation, decided to just use canonical `COUNTER` and `CHECKSUM` signal names forthe time being to minimize required changes to openpilot. We'll revisit this after an MVP port is merged.

PR #193, opened 2019-10-17
Fixes and new message for VW MQB, fix for Accord Touring

jyoung8607:
Add a new MQB message to detect gearshift position for the Volkswagen e-Golf, which doesn't have a traditional gearbox and therefore doesn't have the signal we usually listen for.

As part of the VW CRC work, our updated/refactored DBC processor in commaai/openpilot#836 correctly spots a problem in the 2016 Accord Touring DBC. Two checksum signals were incorrectly marked as little-endian when they should be big. In fact, the start bits were already set to the big-endian side, so I'm not sure how this even worked before.

While I'm in the neighborhood, quit throwing stones from my glass house and fix byte order issues with certain other homebrewed VW MQB signals. No functional issues here, just cleanliness and clarity.


honda_accord_touring_2016_can.dbc
```diff
@@ -177,14 +177,14 @@ BO_ 597 ROUGH_WHEEL_SPEED: 8 VSA
...
- SG_ CHECKSUM : 59|4@1+ (1,0) [0|15] "" XXX
+ SG_ CHECKSUM : 59|4@0+ (1,0) [0|15] "" XXX
 
 BO_ 660 SCM_COMMANDS: 8 SCM
  SG_ RIGHT_BLINKER : 6|1@0+ (1,0) [0|1] "" NEO
  SG_ LEFT_BLINKER : 5|1@0+ (1,0) [0|1] "" NEO
  SG_ WIPERS_SPEED : 4|2@0+ (1,0) [0|3] "" NEO
  SG_ COUNTER : 61|2@0+ (1,0) [0|3] "" XXX
- SG_ CHECKSUM : 59|4@1+ (1,0) [0|15] "" XXX
+ SG_ CHECKSUM : 59|4@0+ (1,0) [0|15] "" XXX
```
rbiasini: Good catch!

PR #194, opened 2019-10-22
Update toyota_nodsu_pt_generated.dbc

Hubblesphere:
Added "ABS_ENGAGE" message signal to dbc file.


rbiasini:
@Hubblesphere , you need to modify the file in `opendbc/generator/toyota` and then run generator.py.
Thanks.

PR #199, opened 2019-11-23
Can migration

rbiasini:
Move Can libraries, including dbc parser and packer, from openpilot to opendbc for better abstraction.

TODO: Makefile still depends on phonelibs.


can/dbc_template.cc
```diff
@@ -0,0 +1,81 @@
...
+namespace {
+
+{% for address, msg_name, msg_size, sigs in msgs %}
+const Signal sigs_{{address}}[] = {
+  {% for sig in sigs %}
+    {
+      {% if sig.is_little_endian %}
+        {% set b1 = sig.start_bit %}
+      {% else %}
+        {% set b1 = (sig.start_bit//8)*8  + (-sig.start_bit-1) % 8 %}
```
ryleymcc, line 12: Misplaced comment?

PR #211, opened 2020-01-19
Hyundai-Kia-Genesis (HKG)

xx979xx:
add electrical gear
fix driver torque signal: scale up to match LKAS StrToqReq signal (modified by this [commit](https://github.com/commaai/opendbc/commit/01cffdca8c7ff65cbb7d51c3f7a8aa4a4cc87bac))


hyundai_kia_generic.dbc
```diff
@@ -792,7 +792,7 @@ BO_ 1362 SNV11: 4 SNV
  SG_ CF_Snv_IRLampControl : 10|2@1+ (1.0,0.0) [0.0|3.0] ""  BCM,CLU,HUD
 
 BO_ 593 MDPS12: 8 MDPS
- SG_ CR_Mdps_StrColTq : 0|11@1+ (0.0078125,-8.0) [-8.0|7.9921875] "Nm"  LDWS_LKAS
+ SG_ CR_Mdps_StrColTq : 0|11@1+ (1.0,-1024.0) [-1024.0|1024.0] ""  LDWS_LKAS
```
geohot: Why change this? It's better to keep the DBC in real units if possible, unless the real units are wrong.

devtekve: @geohot I know it was a long long time ago, but do you remember why we went with this change? I am having a similar "conflict" for CANFD and wanted to update them to use real units and found that this was intentional here.

PR #216, opened 2020-02-06
Add RPM signal

arne182:
Thanks @Hubblesphere


nelsonjchen:
I'm trying to diagnose why the engine braking isn't working with the ACC_CONTROL command and having this signal in the DBCs would be very helpful.

pd0wm:
Where is the 0.78125 coming from?

arne182:
https://github.com/Glidepathuav/opendbc/blob/master/2017_toyota_master.dbc
The leaked Toyota opendbc

PR #219, opened 2020-02-15
Toyota Blind Spot Monitor (TSS2-only?)

nelsonjchen:
I'm not sure what was happening in #190 with @arne182's PR. Was that pulled out of the BSM debug/calibration mode in Techstream or something? 🤷‍♂ 

Anyway, I compared the traffic of my 2020 Corolla Hatchback SE w/ BSM with @Hubblesphere's 2019 Corolla Hatchback SE w/o BSM and I think I found the signal for the BSM indication in my car. I don't know if this is generic to other Toyotas.

This is also my first time editing DBC files. Am I doing this right?

Manual Page describing the ADJACENT/UPCOMING difference:

Screen Shot 2020-02-15 at 12 15 37 AM

example drive: https://my.comma.ai/cabana/?route=1ce1b041fb068985%7C2020-02-09--19-03-13&exp=1613289249&sig=36Plokk1LIIRXmRmokD%2FhESI%2BgmBceumViFs9q708Q4%3D&max=122&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F1ce1b041fb068985%2Fc9957116b7073aafaadfe5e87ac129fb_2020-02-09--19-03-13

Freeway stuff is in the middle.


arne182:
This looks good. But that signal might be a tssp2 thing. On our old system we did not have that cabana data. Might be interesting if the BSM method that we had for tssp1 works for 2 as well.

pd0wm:
Cool stuff!

If it's TSS2 only it can just be added to the `no_dsu` dbc files.

nelsonjchen:
I wrote a message in #toyota-lexus asking for more samples and drives to see if this is TSS2 only. If I don't get enough samples or get a conclusive analysis in a week or two, I'll remake the definitions to the `no_dsu` files and ask for another review.

nelsonjchen:
One example received: "2017 Prius Two with BSM and cross traffic alert."

The bits that would be `ADJACENT_ENABLED` and `APPROACHING_ENABLED` are high but the rest of the signals aren't there.

nelsonjchen:
Received another example from @simstig: "2019 Rav4 hybrid with BSM"

There seem to be more bits high at `3f6` than my corolla probably for the RCTA stuff? Unfortunately, this example doesn't have any passing action. Still, the bits laid low so that's actually something. Anyway I'll be hanging tight until I get another sample with passing action.

nelsonjchen:
I got some passing action. Same bits go high. 

For now, I've moved the BSM definitions to something that the `nodsu` files import.

pd0wm:
@nelsonjchen Good use of the pre-processor! Is there anything that needs to be done or can I merge this PR?

nelsonjchen:
I think it's good enough for now.

nelsonjchen:
That said, I do leave some unresolved things that don't block this merge:

* Does `3f6` bring other bits high if you signal while a car is in a conflict zone? This results in rapid flashing on the mirror. 
* If above, does that ^ occur on non-TSS2 Toyotas? How about at `3f6` even though the presence signals for TSS2 don't appear? 

I could try to answer the first question as I did have a drive where that did happen I was on the outside lane of a two-lane left turn but cabana was just barfing loading the drive. Oh well, I'll keep in mind if I go on another drive that hopefully actually loads in cabana with such an event.

arne182:
Blinking in the mirror does occur if car in conflict zone on non tss2 cars. Before we used to use this signal of the light but after some research from @heatnation and me we figured out how to measure the distance to obstacle from the two radar sensors. Which gives you a much better resolution as it picks up barriers and various other obstacles.

PR #223, opened 2020-03-03
Add Honda-Bosch lane line detection signals.

DisgracedPilot:
This pull request is joint with jpancotti.

This seeks to add information about the Honda-Bosch camera system lane line detection signals. It includes values to support a degree three polynomial lane line model (LINE_OFFSET, LINE_ANGLE, LINE_CURVATURE). Line curvature rate is currently listed as LINE_PARAMETER because it is unknown if this is a low-quality curvature rate value or if it is something else. Have seen it spike up in magnitude for curbs, so it may be some kind of line upward / downward slope or line height value. Other signals include an indicator for solid / dashed line texture, line detection probability, line distance visible, a value related to line thickness. Many signal values are still undecoded. Please see the comments in the DBC. There may also be some additional unidentified signals such as line colour.

The Bosch camera outputs appear to stay in the same message structure across different vehicles. Observed in the Accord, Civic, CR-V.

For decoding the polynomial coefficients, attached is a comparison of OpenPilot and (approximately) decoded Bosch lane line parameters. The scenario is of the vehicle taking a moderate left curve then driving straight. Note that the suspected "line jerk" (curvature rate) value does not match the OpenPilot curvature rate term at all. It is also interesting to note that Bosch detects the curve at least a second or two later than OpenPilot.

image

Additional justification comes from displaying the decoded Bosch lane lines on the EON UI.

https://www.youtube.com/watch?v=7fSF1RXnJTY

### DIFF
--- a/generator/honda/_bosch_2018.dbc
+++ b/generator/honda/_bosch_2018.dbc
@@ -160,6 +160,82 @@ BO_ 545 XXX_16: 6 SCM
  SG_ COUNTER : 45|2@0+ (1,0) [0|3] "" BDY
  SG_ CHECKSUM : 43|4@0+ (1,0) [0|15] "" BDY
 
+BO_ 576 LEFT_LANE_LINE_1: 8 CAM
+ SG_ LINE_DISTANCE_VISIBLE : 39|9@0+ (1,0) [0|1] "" XXX
+ SG_ LINE_PROBABILITY : 46|6@0+ (0.015625,0) [0|1] "" XXX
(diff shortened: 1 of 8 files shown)


rbiasini:
nice. we'll merge after the tests pass.

DisgracedPilot:
Hm...it's complaining that the generator test has failed. Do I need to run generator.py with my new _bosch_2018.dbc and provide those output files?

rbiasini:
yeah, run the generator and commit the changes

DisgracedPilot:
Looks like the tests pass now. Thank you very much!

PR #225, opened 2020-03-03
fix: Replicate changes done on _honda_2017.dbc

AravindaDP:
After _dual_can_nidec_2018.dbc was created following commits were done on
_honda_2017.dbc
- 0132110 : Reverse engineer AEB in Honda
- 7bb1e33 : honda nidec AEB values
- cb27d6e : Honda Nidec: add new ACC_HUD signals to all other cars…
- fa5dc68 : Fix honda dbc files after steer torque addition

Replicating them in _dual_can_nidec_2018.dbc which is basically almost similar to _honda_2017.dbc
But some messages have different structure. (Specially BO_ 506 BRAKE_COMMAND)

### DIFF
--- a/generator/honda/_dual_can_nidec_2018.dbc
+++ b/generator/honda/_dual_can_nidec_2018.dbc
@@ -56,12 +56,20 @@ BO_ 380 POWERTRAIN_DATA: 8 PCM
 
 BO_ 420 VSA_STATUS: 8 VSA
  SG_ USER_BRAKE : 7|16@0+ (0.015625,-1.609375) [0|1000] "" EON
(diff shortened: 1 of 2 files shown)


rbiasini:
@AravindaDP , is the Honda Fit Hybrid port broken without these changes then? I see it's a non-officially supported car, which is why tests did not catch it.
I'll merge these changes, but would be nice to perhaps restructure the inherited dbc files to not have so much repetitions.

AravindaDP:
@rbiasini Thanks for merging.

Yes, Honda Fit Hybrid broke without these changes. And yes it's an unsupported car at moment. 

And I guess we would better to keep it separate from other Nidec Honda following same reason at #105 until it's vehicle port is upstreamed.

PR #233, opened 2020-03-12
Fixed up sign on Driver Torque, added unit

avolmensky:
Fixing DriverTorque sign as lane change is not working properly

### DIFF
--- a/nissan_2017.dbc
+++ b/nissan_2017.dbc
@@ -86,7 +86,7 @@ BO_ 389 Steering: 8 XXX
  SG_ LKAS_Torque : 47|8@0+ (1,0) [0|255] "" XXX
  SG_ CRC : 63|8@0+ (1,0) [0|127] "" XXX
  SG_ Angle : 23|18@0+ (-0.01,1310) [0|262143] "" XXX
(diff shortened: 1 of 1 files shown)


nissan_2017.dbc
```diff
@@ -86,7 +86,7 @@ BO_ 389 Steering: 8 XXX
  SG_ LKAS_Torque : 47|8@0+ (1,0) [0|255] "" XXX
  SG_ CRC : 63|8@0+ (1,0) [0|127] "" XXX
  SG_ Angle : 23|18@0+ (-0.01,1310) [0|262143] "" XXX
- SG_ DriverTorque : 7|12@0+ (1,-2047) [0|4095] "" XXX
+ SG_ DriverTorque : 7|12@0+ (-1,2047.5) [0|4095] "Ncm" XXX
```
rbiasini: what's `Ncm`? :D

rbiasini: also, the offset should be an integer. 2047 or 2048?

avolmensky: Ncm = Nm * 100. Should I scale this to Nm?

rbiasini: yeah, set the factor to 0.01

PR #235, opened 2020-03-17
wheel sensor error bit toyota

wocsor:


pd0wm:
Can you use the same naming as the speeds, so `SENSOR_ERROR_FL`?

wocsor:
Done

pd0wm:
Is it also present on TSS2? Is there a good way to trigger an error?

wocsor:
it should be present on all Toyotas with ABS, but i don't have a TSS2 car to test this on. you can trigger an error by unplugging a sensor, but i'm not sure if it's able to detect anything outside of an open circuit

PR #245, opened 2020-04-18
Create honda_hrv_touring_2019_can_generated.dbc

cowanhm:
To match PR#1393
honda_hrv_touring_2019_can_generated.dbc will need added to match the HRV PR#1393 with commaai/openpilot.

After in-depth cabana review, consulting with @csouers, and lastly Honda, we have established that HR-V Door signals are not on F-CAN.  Since all other Honda's use BO_ 1029 "generated" from the "_honda_2017.dbc" seed file, it creates a can_valid() error in parser.cc if BO_1029 is included in the _generated.dbc version.

I've created the "honda_HRV_touring_2019_can_generated.dbc" file to NOT include 1029, but instead include BO_ 1137 for the "doors".  When parser.cc is called, it will see 1137 in the CAN-stream, but it's never referenced in code.

If this _generated file is to be created with "generator.py" each honda~_can.dbc will need updated with BO_ 1029 in the specific file, rather than in the seed file.


cowanhm:
Found a simple work-around.  Will resubmit

cowanhm:
Found a work-around.  Will resubmit

PR #247, opened 2020-04-18
Honda HRV DBC Updates

cowanhm:
These are the DBC files to support Honda HRV additions. 
running generator.py will cause
./opendbc/generator/honda/honda_hrv_touring_2019_can.dbc
to create 
./opendbc/honda_hrv_touring_2019_can_generated.dbc
and is called by 
./selfdrive/car/honda/values.py under class CAR.HRV


cowanhm:
Resolving conflict

PR #248, opened 2020-04-18
Create DBC for HRV

cowanhm:
Values from honda_fit almost work, with exception of DOOR_STATUS BO_1029 is not valid in the HRV can. If DOOR_STATUS is called, it will result in can_valid() == False!  After extensive searching, DOOR_STATUS signal does not exist on CAN-F. 

This DBC has been created for if DOOR signal is discovered.
Redundant safety will be inherited from seatbelt_unlanched signal disengaging OP


cowanhm:
Added correct door signal for hrv
(BO_ 422 SCM_BUTTONS: 8 SCM)
 SG_ DRIVERS_DOOR_OPEN : 63|1@0+ (1,0) [0|1] "" EON

PR #252, opened 2020-04-29
WIP: Add RDM/LDW signaling to Honda LKAS_HUD and RDM bit to 0xe4 for Honda Bosch

csouers:
- Clarifies preexisting signals and defines new ones. 
- Rearranges signals to better align with their "group". 
- 0xe4 on Bosch has a bit active during RDM, but it doesn't have any obvious effect on steering.

RDM_HUD_2 is secondary bit to RDM_HUD. It seems to only be active primarily on Bosch cars if the lane departure event is bad enough. I haven't seen it active on Nidec.
RDM_ON_1 is active with an enabled system, but only after a button press (Nidec). RDM_ON_0 represents the true state of the system, again, at the button.


ffhspa:
Would be great if this could be merged before 0.8. Thanks a lot!

jyoung8607:
Will pull these changes in, but have to do some cleanup/consolidation in the Honda DBC generator first.

mvl-boston:
Let me know if you need assistance as I'm happy to support.   There is directional LDW that is missing and can be added here too (eg: different LDW left vs LDW right signals).

jyoung8607:
@mvl-boston @csouers I tried to pull this one forward in time, and somewhat succeeded, but not quite. Also, there will be some actual code changes needed for the openpilot longitudinal path.

If you'll open a clean working version of these changes against modern opendbc, I can take another look.

PR #257, opened 2020-05-03
Mazda: Better GEAR signal tracking the gear stick rather than the gear box

Jafaral:
A simpler signal for better gear parsing
Signed-off-by: Jafar Al-Gharaibeh <to.jafar@gmail.com>

### DIFF
--- a/mazda_cx5_gt_2017.dbc
+++ b/mazda_cx5_gt_2017.dbc
@@ -485,15 +485,13 @@ BO_ 606 SPEED_TBD: 8 XXX
  SG_ NEW_SIGNAL_1 : 23|12@0- (1,0) [0|65535] "" XXX
 
 BO_ 552 GEAR: 8 XXX
- SG_ NEW_SIGNAL_2 : 8|1@0+ (1,0) [0|255] "" XXX
  SG_ NEW_SIGNAL_3 : 11|1@0+ (1,0) [0|1] "" XXX
- SG_ NEW_SIGNAL_4 : 18|3@0+ (1,0) [0|255] "" XXX
  SG_ NEW_SIGNAL_5 : 26|3@0+ (1,0) [0|255] "" XXX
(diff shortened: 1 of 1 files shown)


Jafaral:
@quillford This is the change I brought up earlier today, checked with multiple drives. I have 4 gear positions that corresponds to exactly 4 values. I retained the same signal name and renamed the old one, so this change should be transparent to OP. 

Once you get a chance to review and sync OP module again I should be able to test and finish up any remaining items in OP Mazda.

mazda_cx5_gt_2017.dbc
```diff
@@ -485,15 +485,15 @@ BO_ 606 SPEED_TBD: 8 XXX
...
  SG_ NEW_SIGNAL_3 : 11|1@0+ (1,0) [0|1] "" XXX
- SG_ NEW_SIGNAL_4 : 18|3@0+ (1,0) [0|255] "" XXX
  SG_ NEW_SIGNAL_5 : 26|3@0+ (1,0) [0|255] "" XXX
  SG_ NEW_SIGNAL_6 : 31|5@0+ (1,0) [0|31] "" XXX
  SG_ NEW_SIGNAL_7 : 39|1@0+ (1,0) [0|255] "" XXX
- SG_ MORE_GEAR : 7|4@0+ (1,0) [0|255] "" XXX
- SG_ GEAR : 35|3@0+ (1,0) [0|15] "" XXX
- SG_ REVERSE : 36|1@0+ (1,0) [0|1] "" XXX
+ SG_ NEW_SIGNAL_4 : 18|3@0+ (1,0) [0|255] "" XXX
+ SG_ NEW_SIGNAL_2 : 8|1@0+ (1,0) [0|255] "" XXX
```
quillford, line 493: if we don't know what these signals are, we can just leave it out

Jafaral: sure, dropped!

PR #258, opened 2020-05-04
Toyota ACC_CONTROL PERMIT_BRAKING and ACCEL_CMT_ALT speculated definition

nelsonjchen:
Not sure if these names are accurate. Also not quite sure of their effect other than existing openpilot's usage (or non-usage). These are the best names I can guess to give them for now. They're better than `SET_ME_1` and/or nothing at least?

So, might as well give these signals a name, even if it might not be right, and make some comments on them. Hopefully it makes looking at these in Cabana a little less mysterious.

Unfortunately, merging this in will necessitate some changes to openpilot for tests and the toyota CAN for the existing `SET_ME_1` signal in the accel command: https://github.com/commaai/openpilot/blob/ad2c54e9916301d63907d96c3e9dec046bd016b1/selfdrive/car/toyota/toyotacan.py#L37 . `ACCEL_CMD_ALT` can be left unimplemented as a value of `0` for it is accepted fine like current openpilot.

Drive from an openpilot branch that copies `ACCEL_CMD` to `ACCEL_CMD_ALT` and sets `PERMIT_BRAKING` to high when there's a lead car like stock ACC on a 2020 Corolla: https://my.comma.ai/cabana/?route=1ce1b041fb068985%7C2020-05-03--12-28-37&exp=1620090014&sig=subo0Dfr9iy9%2Fyd36VVHGNPXA9v8u%2BlVxPETS2Xyep4%3D&max=16&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F1ce1b041fb068985%2F2353d33549b4b1b7c6898bf665188c06_2020-05-03--12-28-37 

The above branch I used is just experimental and that's the only drive recording I have on hand right now for viewing.

### DIFF
--- a/generator/toyota/_toyota_2017.dbc
+++ b/generator/toyota/_toyota_2017.dbc
@@ -62,10 +62,10 @@ BO_ 180 SPEED: 8 XXX
 
 BO_ 353 DSU_SPEED: 8 XXX
  SG_ FORWARD_SPEED : 15|16@0- (0.00390625,-30) [0|255] "kph" XXX
- 
+
 BO_ 452 ENGINE_RPM: 8 CGW
  SG_ RPM : 7|16@0- (0.78125,0) [0|0] "rpm" SCS
- 
+
 BO_ 466 PCM_CRUISE: 8 XXX
  SG_ GAS_RELEASED : 4|1@0+ (1,0) [0|1] "" XXX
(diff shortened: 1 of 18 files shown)


pd0wm:
`PERMIT_BRAKING` is what the stock system uses to switch between engine braking and friction brakes? If I recall correctly the stock system only uses friction brakes when there is a lead?

What does `ACCEL_CMD_ALT` do in the stock system?

PR #259, opened 2020-05-09
DBC progress checkpoint on Volkswagen PQ35/PQ46/NMS

jyoung8607:
This is not necessarily finalized quality, but there's five people driving around on it, there's an upstream safety, and lack of current DBC data is pausing @adeebshihadeh's work on message length safety checks. There will be at least one more round of revisions to come.


adeebshihadeh:
Thanks @jyoung8607, we can leave this in a PR for now

jyoung8607:
Yes, I would like to, especially since this DBC doesn't really related to MkIV Golf anymore. However, it will need follow-on changes to keep everything in sync, packer detection of which checksum algo to use in addition to the obvious changes in OP. I am likely to spend time on that toward the end of May or beginning of June.

vw_golf_mk4.dbc
```diff
@@ -1040,11 +1062,338 @@ BO_ 1324 ADR_1: 8 XXX
...
+ SG_ ACA_gemZeitl : 40|4@1+ (1,0) [0|15] "" XXX
+ SG_ ACA_ACC_Verz : 44|1@1+ (1,0) [0|1] "" XXX
+ SG_ ACA_StaGRA : 48|3@1+ (1,0) [0|7] "" XXX
+ SG_ ACA_ID_StaGRA : 51|5@1+ (1,0) [0|31] "" XXX
+ SG_ ACA_Codierung : 56|1@1+ (1,0) [0|1] "" XXX
+ SG_ ACA_Tachokranz : 57|1@1+ (1,0) [0|1] "" XXX
+ SG_ ACA_Aend_Zeitluecke : 58|1@1+ (1,0) [0|1] "" XXX
+ SG_ ACA_Zaehler : 60|4@1+ (1,0) [0|15] "" XXX
+
+BO_ 208 Lenkhilfe_3: 8 XXX
```
adeebshihadeh, line 1105: This is 6 bytes according to the can log from [this route](https://github.com/commaai/panda/blob/master/tests/safety_replay/test_safety_replay.py#L21)

jyoung8607: Fixed!

jyoung8607:
Still a WIP. Holding in draft until I'm able to test more thoroughly, and also set up the companion PRs for openpilot and Panda to track the updated DBC filename for VW MQB.

PR #270, opened 2020-05-31
Subaru preglobal DBC update

martinl:
- Moved preglobal dbc files to use dbc generator, since there is currently 2 different messages between Outback and Forester
- Changed SG_ Steer_Torque_Sensor scaling factor from 8 to 1 which is part of the fix for panda safety tests (https://github.com/commaai/panda/pull/532)
`SG_ Steer_Torque_Sensor : 29|11@1- (1,0) [-1000|1000] "" XXX`
- Renamed Dash_State Units2 > Units and removed original Units signal which proved unreliable after few weeks of testing
- Fixed Forester Steer_Torque_Sensor signal sign (correct scaling factor is -1)
- Renamed ES_DashStatus 3SecondDisengage to Seatbelt_Disengage to match DBC spec
- Removed redundant comments
- Fixed G_Sensor signals scaling factors
- Fixed min, max value definitions according to field length
- Fixed some missing units


generator/subaru/_subaru_preglobal_2015.dbc
```diff
@@ -0,0 +1,214 @@
...
+
+BS_:
+
+BU_: XXX X
+
+
+BO_ 2 Steering: 8 XXX
+ SG_ Steering_Angle : 7|16@0- (0.1,0) [-500|500] "degree" XXX
+ SG_ Counter : 27|3@0+ (1,0) [0|15] "" XXX
+ SG_ Checksum : 39|8@0+ (1,0) [0|255] "" XXX
```
adeebshihadeh, line 42: Why is the endianness different for this msg?

generator/subaru/subaru_forester_2017.dbc
```diff
@@ -0,0 +1,39 @@
...
+ SG_ Cruise_Activated : 54|1@1+ (1,0) [0|1] "" XXX
+
+BO_ 881 Steering_Torque: 8 XXX
+ SG_ Steering_Motor_Flat : 0|10@1+ (32,0) [0|16500] "" XXX
+ SG_ Steer_Torque_Output : 16|11@1- (-32,0) [-16500|16500] "" XXX
+ SG_ LKA_Lockout : 27|1@1+ (1,0) [0|1] "" XXX
+ SG_ Steer_Torque_Sensor : 29|11@1- (-1,0) [-1500|1500] "" XXX
+ SG_ Steering_Angle : 40|16@1- (-0.033,0) [-600|600] "" XXX
+
+CM_ SG_ 320 Off_Throttle_2 "Less sensitive";
```
adeebshihadeh, line 17: Comment lines should go in the same DBC the signal is defined in

generator/subaru/_subaru_preglobal_2015.dbc
martinl: I guess Subaru had a good reason for doing this, since Global platform has similar exception with Steering_Angle signal:
```
BO_ 2 Steering: 8 XXX
 SG_ Steering_Angle : 7|16@0- (0.1,0) [0|65535] "" XXX
 SG_ Counter : 25|3@1+ (1,0) [0|7] "" XXX
 SG_ CHECKSUM : 32|8@1+ (1,0) [0|255] "" XXX
```

generator/subaru/subaru_forester_2017.dbc
martinl: Done

martinl:
I took another pass through the signals and fixed min/max values to match field lengths and types

generator/subaru/subaru_forester_2017.dbc
```diff
@@ -0,0 +1,15 @@
...
+ SG_ Cruise_On : 16|1@1+ (1,0) [0|1] "" XXX
+ SG_ Cruise_Set_Speed : 24|8@1+ (1,0) [0|255] "" XXX
+ SG_ Counter : 40|3@1+ (1,0) [0|7] "" XXX
+ SG_ Cruise_Activated : 54|1@1+ (1,0) [0|1] "" XXX
+
+BO_ 881 Steering_Torque: 8 XXX
+ SG_ Steering_Motor_Flat : 0|10@1+ (32,0) [0|1000] "" XXX
+ SG_ Steer_Torque_Output : 16|11@1- (-32,0) [-1000|1000] "" XXX
+ SG_ LKA_Lockout : 27|1@1+ (1,0) [0|1] "" XXX
+ SG_ Steer_Torque_Sensor : 29|11@1- (-1,0) [-1000|1000] "" XXX
```
adeebshihadeh, line 14: Hmm, that's odd that they'd just use the same message except flip the sign on this signal. Are you sure about this one?

martinl: Before making the change, I confirmed from cabana drive that the user torque signal is reversed for Foresters. I have two Forester users who confirmed assisted lane change works correctly now. They had to nudge the wheel to the opposite side to initiate lane change before

PR #276, opened 2020-06-18
UpdateValid: break loop if can_valid is false

deanlee:


adeebshihadeh:
If there are multiple messages that have timed out, breaking out of the loop early prevents the debug prints for the other messages.

PR #290, opened 2020-07-27
Add 2020 Acura RDX

csouers:
This vehicle's powertrain's CAN topology is extremely similar to 2.0L Accord. Door ajar sensors not available on F-CAN except for the Driver's Door (on SCM_FEEDBACK). STEERING_SENSORS is different from every other Bosch car, so I've split it up.


adeebshihadeh:
Hmm, instead of all the changes to the existing bosch DBCs, can we rename _bosch_2018 to _bosch and make _bosch_2018 and _bosch_2020?

csouers:
Seems reasonable. Take a look. Also the port should work on 2019 RDX as well as it's part of the same generation. We don't have a tester for it though. I suspect it's mostly the same car.

adeebshihadeh:
Hmm, I meant make _bosch_2018 and _bosch_2020 import common signals from _bosch. Other than that, looks good.

csouers:
Ah. Ok. Will fix on the next push.

adeebshihadeh:
_bosch_{2018,2020} should import _bosch, not the car model

csouers:
Having `_bosch_2018` and `_bosch_2020` import common signals from `_bosch` will require changes to generator.py.

csouers:
Nevermind. Tried this earlier and it didn't work.

csouers:
@adeebshihadeh Any advice on the naming convention for this? `__pre` is what I came up with. I'm open to suggestions for that.

csouers:
@adeebshihadeh I think I've done as you were explaining. This allows the generator to pull in `includes` two levels deep. If you're happy with the file structure now, I'll clean up the generator.

adeebshihadeh:
Let's do this: 
* to get this merged, copy the base bosch dbc and call it _bosch_2020 with the necessary changes. we'll have a DBC with duplicate signals, but the diff will be minimal and we can clean it up later
* make a separate PR for the generator changes to clean up the duplicated signals. ideally this should also include unit tests for the generator

csouers:
Rebased and changes made. I think this should be good to go.

PR #292, opened 2020-08-07
Create hyundai_kia_generic_hybrid.dbc

xps-genesis:
Separate Hyundai Hybrid DBC due to common message with different signals packed between EV and HEV


adeebshihadeh:
I don't think this is necessary at this point, since we have a common gas pressed signal. If we do need to start breaking different models out, we'll use the generator.

PR #295, opened 2020-09-06
Add SECONDARY_STEER_ANGLE for ZSS

ShaneSmiskol:
Needed for https://github.com/commaai/openpilot/pull/2136

Adds the ZSS SECONDARY_STEER_ANGLE message to Prius DBC so openpilot can read sensor and use it over stock sensor


adeebshihadeh:
This should be another dbc that’s included in the prius dbc. Maybe make one like the _comma dbc and call it _community_features?

ShaneSmiskol:
How's that?

PR #302, opened 2020-10-08
Switch update string to using char *.

alexlitz:
The attempt at making the patch requested [here](https://github.com/commaai/cereal/pull/93#issuecomment-703524013).


can/common.h
```diff
@@ -56,7 +56,7 @@ class CANParser {
             const std::vector<SignalParseOptions> &sigoptions);
   void UpdateCans(uint64_t sec, const capnp::List<cereal::CanData>::Reader& cans);
   void UpdateValid(uint64_t sec);
-  void update_string(std::string data, bool sendcan);
```
adeebshihadeh, line 59: why is this uint32_t? should be size_t

alexlitz: That seems accurate.

PR #307, opened 2020-10-18
Added Pedal messages for GM

JasonJShuler:
Added support for the Pedal Interceptor for GM Powertrain.
Will be required to add support for GM vehicles that lack stock ACC

Also added AEB message sent by front facing cameras in vehicles without radar. This will be useful for hijacking AEB to provide braking again in vehicles without stock ACC. This will be especially important in non-EV cars that lack single pedal mode

512, 513 & 800


JasonJShuler:
Not sure what the delay is, but I moved it into my repo if maybe that had something to do with it.

PR #314, opened 2020-10-28
Added messages in preparation for Bolt EV port (Pedal and AEB)

JasonJShuler:
Added support for the Pedal Interceptor for GM Powertrain (512 and 513)
Will be required to add support for GM vehicles that lack stock ACC. Note that the translation and scale values have been optimized for the Pedal on the Chevy Bolt EV, but they have been confirmed to work in some other GM cars.

Also added AEB message sent by front facing cameras in vehicles without radar (800: this may be used by the ASCM as well but that hasn't been looked into). This will be useful for hijacking AEB to provide braking again in vehicles without stock ACC. This will be especially important in non-EV cars that lack single pedal mode


JasonJShuler:
I think these are fairly innocuous additions - do let me know if you feel otherwise

TheMutley:
Have you tested the braking with the vision based AEB ? I did on a Bolt and it’s only usable for AEB, not a viable solution for ACC braking if it’s what you’re trying to achieve.

JasonJShuler:
> Have you tested the braking with the vision based AEB ? I did it on a Bolt and it’s only usable for AEB, not a viable solution for ACC braking if it’s what you’re trying to achieve.

No, I'm just adding the CAN message for now.

JasonJShuler:
Going to make sure naming conventions and everything is correct. closing for now.

PR #315, opened 2020-11-10
ZSS support for Prius

zorrobyte:
This adds native community support for ZSS

Related OP changes here:
https://github.com/commaai/openpilot/pull/1082/files


adeebshihadeh:
See https://github.com/commaai/openpilot/pull/2136

PR #321, opened 2020-11-30
Fixed CM_ (comments) syntax

briskspirit:
Comments (CM_) in .dbc files should end with the semicolon. genetator.py injected comments without semicolon.


generator/toyota/_comma.dbc
```diff
@@ -5,10 +5,10 @@ BO_ 359 STEERING_IPAS_COMMA: 8 IPAS
  SG_ SET_ME_X00 : 31|8@0+ (1,0) [0|255] "" XXX
  SG_ DIRECTION_CMD : 38|2@0+ (1,0) [0|3] "" XXX
  SG_ SET_ME_X40 : 47|8@0+ (1,0) [0|255] "" XXX
- SG_ SET_ME_X00 : 55|8@0+ (1,0) [0|255] "" XXX
+ SG_ SET_ME_X00_1 : 55|8@0+ (1,0) [0|255] "" XXX
```
adeebshihadeh, line 8: why did this change? if you want to change this, do it in another PR

briskspirit: This also belongs to syntax fixes. Signal names can't be the same inside one message.

pd0wm: It's a duplicate of the signal above. This message is not used anywhere in code. Still better to do it in a separate PR if we ever need to look for the change.

briskspirit: Reverted back, will do in a separate PR

briskspirit:
Will redo everything and submit a new PR.

PR #322, opened 2020-12-03
Add ZSS signal

ShaneSmiskol:
Can make _community_features dbc instead if we don't want to import this to a bunch of cars other than the Prius


generator/toyota/_comma.dbc
```diff
@@ -25,3 +25,6 @@ CM BO_ STEERING_IPAS_COMMA "Copy of msg 614 so we can do angle control while the
   SG_ CHECKSUM_PEDAL : 47|8@0+ (1,0) [0|255] "" EON
 
  VAL_ 513 STATE 5 "FAULT_TIMEOUT" 4 "FAULT_STARTUP" 3 "FAULT_SCE" 2 "FAULT_SEND" 1 "FAULT_BAD_CHECKSUM" 0 "NO_FAULT" ;
+
+BO_ 35 SECONDARY_STEER_ANGLE: 8 XXX "ZSS is a high-precision steering angle sensor created by zorrobyte to replace the crappy sensors in most TSS1 Toyotas. Learn more: https://github.com/commaai/openpilot/wiki/Toyota-Lexus#zorro-steering-sensor-zss"
```
adeebshihadeh, line 29: This isn't the right way to do write a comment. See https://github.com/commaai/opendbc/blob/master/generator/toyota/_comma.dbc#L11

ShaneSmiskol: so just a semicolon? how's that now?

geohot:
LGTM

generator/toyota/_comma.dbc
```diff
@@ -25,3 +25,8 @@ CM BO_ STEERING_IPAS_COMMA "Copy of msg 614 so we can do angle control while the
   SG_ CHECKSUM_PEDAL : 47|8@0+ (1,0) [0|255] "" EON
 
  VAL_ 513 STATE 5 "FAULT_TIMEOUT" 4 "FAULT_STARTUP" 3 "FAULT_SCE" 2 "FAULT_SEND" 1 "FAULT_BAD_CHECKSUM" 0 "NO_FAULT" ;
+
+BO_ 35 SECONDARY_STEER_ANGLE: 8 XXX
+ SG_ ZORRO_STEER : 7|24@0- (0.004901594652,0) [-500|500] "" XXX
+
+CM BO_ SECONDARY_STEER_ANGLE "ZSS is a high-precision steering angle sensor created by zorrobyte to replace the crappy sensors in most TSS1 Toyotas. Learn more: https://github.com/commaai/openpilot/wiki/Toyota-Lexus#zorro-steering-sensor-zss";
```
adeebshihadeh, line 32: let's not editorialize the comment. we can just say it's an improved steer angle sensor.

PR #326, opened 2020-12-07
Lexus ES300h 2018

TheInventorMan:
Adding DBC file for 2018 Lexus ES 300h. Will submit openpilot PR once this one gets merged.


adeebshihadeh:
This exactly matches the Lexus CT200H 2018 DBC. Let's use that. If they diverge at some point, we can make a new DBC.

PR #328, opened 2020-12-08
Added CRUISE_STATE value 11 description (timer_3sec)

briskspirit:
Added description for CRUISE_STATE value 11 : 3 seconds waiting timer. Will switch cruise state to standstill (7) after 3 seconds.
Added description for CRUISE_STATE value 9 : set cruise speed higher.
Added description for CRUISE_STATE value 10 : set cruise speed lower.


adeebshihadeh:
Let's also add a comment that goes into more detail about the states.

briskspirit:
Where do you want to add it? As a CM_ SG_ value?

adeebshihadeh:
Yes

briskspirit:
Ok no problem, will add tomorrow.

briskspirit:
Comment was added. Sorry for the delay.

PR #331, opened 2020-12-23
Add Toyota Odometer Reading to Toyota DBCs

nelsonjchen:
The unit is dependent on unit setting.

PS: Cabana currently doesn't parse the last comment's terminator correctly.

### DIFF
--- a/generator/toyota/_toyota_2017.dbc
+++ b/generator/toyota/_toyota_2017.dbc
@@ -215,6 +215,7 @@ BO_ 1410 VIN_PART_3: 8 CGW
 
 BO_ 1553 UI_SETTING: 8 XXX
  SG_ UNITS : 26|2@0+ (1,0) [0|3] "" XXX
+ SG_ ODOMETER : 43|20@0+ (1,0) [0|1048575] "" XXX
 
 BO_ 1556 STEERING_LEVERS: 8 XXX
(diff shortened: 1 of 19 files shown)


nelsonjchen:
Bizzare, not sure how the generator can differ. It's clean on mine. I'll look into this later.

pd0wm:
You need to rebase with master and rerun the generator. A new DBC file was added (`lexus_nx300_2018_pt_generated.dbc`) causing the test to fail.

nelsonjchen:
@pd0wm Thanks! That was indeed it.

jono-radar:
Hello, I want to test lexus_rx_hybrid_2017_pt_generated.dbc on my vehicle, but I only see two messages over OBD2 on a PeakCan USB adapter.  How to I get access to the full set of messages?

nelsonjchen:
@jono-radar You should have made an explicit issue in this repo rather than *just commenting on a random unrelated issue which is not great GitHub etiquette*. OpenDBC is probably most prominently used by Openpilot but that doesn't mean it is openpilot only. 

That said, you should know there's likely a CAN Gateway that blocks these kinds of messages to the OBD2 port. In openpilot, users are observing the traffic behind the forward-facing camera which is less filtered than the traffic to the OBD2 port. That's likely why you aren't seeing it.

PR #332, opened 2020-12-24
Pacifica: Add cruise state indicator

vanillagorillaa:
This message appears to be an indication of current cruise state, both ACC and Non-ACC modes as well as engaged/disengaged and off states. 

Found from this route - 3c946decd6bc40b9|2020-12-23--09-52-15 

Drive starts with all cruise off. Around 64 seconds in Cruise Control is turned on (non acc). Around 98 seconds Cruise Control is engaged, disengaged at ~128 seconds, and Cruise control is turned off ~146 seconds. 

ACC is turned on ~188 seconds, ~237 seconds ACC is engaged, ~307 ACC is disengaged and ~361 ACC is turned off. 

Plots attached
Screenshot_14
Screenshot_15


adhintz:
LGTM
Confirmed these message are also correct on my pachy 2017 from route=d3950a2b84b695f3%7C2020-12-23--16-28-23

PR #335, opened 2020-12-27
added honda_accord_2016_2017_nidec.dbc for honda serial steering 

reddn:


vanillagorillaa:
There is already a DBC for the 2016 Accord, why not just use that and merge in the differences? https://github.com/commaai/opendbc/blob/master/honda_accord_touring_2016_can.dbc

reddn:
Naming convention and it doesn't show it's a nidec.  If you want me to
merge it then rename the other one i will.  The current touring dbc had
less frame descriptors then this one.  Most of the ones we needed were in
the acura ilx.

vanillagorillaa:
None of the other DBC's list Nidec/Bosch in file name, IE honda_civic_touring_2016_can_generated.dbc

If you were to do it, I'd say match the rest of the files and use the generator/generated files. See https://github.com/commaai/opendbc/tree/master/generator/honda for more information.

reddn:
merged into honda_accord_2106_touring_can.dbc

PR #344, opened 2021-01-27
CANParser::update_string :use const std::string& as paramater

deanlee:


adeebshihadeh:
Can you rebase to get the tests running?

deanlee:
rebased

PR #350, opened 2021-02-16
Update for signal length - ACCEL_Z and ACCEL_X

romybompart:
To avoid - bitstruct.Error: "s15" requires -16384 <= integer <= 16383, the signal length need to be 16 bits.


adeebshihadeh:
Where are you getting this error? Why do you think the signals are one bit larger?

PR #362, opened 2021-03-11
GM steering rate scale & limits fix

qadmus:
Very useful for INDI, and related to https://github.com/commaai/openpilot/issues/20232 

Tasks complete:

- ~~spin wheel to stops at 90deg/sec~~ Good: rate matches intended & measured angle over time.
- ~~try to decode last byte, probably checksum~~ Not able now.
- ~~merge / generate GM and Cadillac~~ Not now.

90deg-steps
rapid-steer


qadmus:
With the rate scale corrected to 1.0, it nicely matches constant rate periods (where I checked the mean rate) and the noisy first-difference.

fixed-rate-1

PR #365, opened 2021-03-14
Tesla AP status

robbederks:


jholster:
Hi, is this file for specific Tesla model or for both S/X & 3?

robbederks:
@jholster This PR specifically is reverse engineered on a 2017 Model S with AP2. I believe the X should be the same, but the 3 and Y might be completely different.

PR #372, opened 2021-04-01
Support checksum for extended IDs on 2021 Honda Bosch

csouers:
- Create bosch_2021.dbc
- Add 2021 Honda Accord 2.0 Touring & Hybrid Touring

These cars do have some newly active bits on 0xe4 and 0x33db but we haven't explored them very far as OP works fine without setting them. They mostly seemed related to lane state and engaged state history. Maybe part of a future UI update?

2021 Odyssey still uses the older STEERING_SENSORS frame ID, so we have to keep it split up like before.

Edit: Looks like it's time to revisit the `_bosch` clean up again. Will take a look tomorrow.


can/common.cc
```diff
@@ -4,11 +4,15 @@ unsigned int honda_checksum(unsigned int address, uint64_t d, int l) {
   d >>= ((8-l)*8); // remove padding
   d >>= 4; // remove checksum
 
+  bool extended = false;
+  if (address > 0x7FF) {extended = true;}
```
adeebshihadeh, line 8:
```suggestion
  bool extended = address > 0x7FF;
```

can/common.cc
```diff
@@ -4,11 +4,14 @@ unsigned int honda_checksum(unsigned int address, uint64_t d, int l) {
...
   d >>= 4; // remove checksum
 
+  bool extended = address > 0x7FF;
+
   int s = 0;
   while (address) { s += (address & 0xF); address >>= 4; }
   while (d) { s += (d & 0xF); d >>= 4; }
   s = 8-s;
   s &= 0xF;
+  if (extended){s += 3 & 0xF;}
```
adeebshihadeh, line 14: what's the intention here? 3 & 0xF = 3

csouers: This was a mistake that happened to work. See latest commit for what I actually meant to do.

can/common.cc
```diff
@@ -4,10 +4,12 @@ unsigned int honda_checksum(unsigned int address, uint64_t d, int l) {
   d >>= ((8-l)*8); // remove padding
   d >>= 4; // remove checksum
 
+  bool extended = address > 0x7FF;
   int s = 0;
   while (address) { s += (address & 0xF); address >>= 4; }
   while (d) { s += (d & 0xF); d >>= 4; }
   s = 8-s;
+  if (extended){ s += 3; }
```
adeebshihadeh, line 12:
```suggestion
  int s = 0;
  while (address) { s += (address & 0xF); address >>= 4; }
  while (d) { s += (d & 0xF); d >>= 4; }
  s = 8-s;
  if (address > 0x7FF) s += 3; // extended can
```

csouers: Nice! Thank you

csouers: The way `address` is used, this won't work as it's always 0 by the time the `3` for the extended id is added. The misra tests in panda actually caught this as I have it the same way there. I've put this back to how it was before.

adeebshihadeh:
I added the new LKAS_HUD msgs to the current Bosch base DBC instead. 2021 Accord should use the same DBC as the 2018 now.

PR #375, opened 2021-04-02
toyota hazard light

cydia2020:
tested on my prius


generator/toyota/_toyota_2017.dbc
```diff
@@ -219,6 +219,7 @@ BO_ 1553 UI_SETTING: 8 XXX
 
 BO_ 1556 STEERING_LEVERS: 8 XXX
  SG_ TURN_SIGNALS : 29|2@0+ (1,0) [0|3] "" XXX
+ SG_ HAZARD_LIGHT : 27|1@0+ (1,0) [0|1] "" XXX
```
KostyaSha, line 222: Is it packet to turn hazard light or just the state of hazzard light button? Btw hazzard button is not under steer...

cydia2020: It's the status, the signal changes to 1 when the hazards are on.
Signal addresses names don't mean anything in particular, it's just how comma had named them initially.

Issue #374, opened 2021-04-02
Parse DBCs dynamically


adeebshihadeh:
Currently we build the DBCs into a [shared library](https://github.com/commaai/opendbc/blob/master/can/SConscript), but we can get rid of this step at build-time if we can parse them at runtime.

adeebshihadeh:
@deanlee interested in this one?

deanlee:
ok

PR #376, opened 2021-04-03
Parse DBCs dynamically

deanlee:
resolve https://github.com/commaai/opendbc/issues/374


deanlee:
the execution time of `ensure_dbc` is 3-5 ms.

pd0wm:
This will break CAN/DBC parsing in plotjuggler, since we only link against the C++ version of canparser, which would no longer have any DBC files. Can we move over the parsing logic to C++ as well without introducing a lot of new code?

deanlee:
I'll try

pd0wm:
Thanks! The use case for dynamic parsing is twofold:
1) Reduce openpilot build time and release size by not compiling all dbc files
2) Dynamic parsing in plotjuggler so we don't have to ship a plugin update every time the DBCs change and support for your own DBC files.

deanlee:
Thanks for the explanation. I know what to do next.

deanlee:
test_dbc_exceptions.py can't catch c++ runtime_error exceptions. comment it out for now

adeebshihadeh:
I opened https://github.com/commaai/openpilot/pull/20741 to run the openpilot CI as a quick check and a bunch of the tests are failing.

deanlee:
fixing

adeebshihadeh:
Feel free to open a new PR against openpilot if you want to run the tests

adeebshihadeh:
I see this is passing the openpilot tests now. We'll review and get this in for 0.8.5!

adeebshihadeh (commented):
Can you also update the PlotJuggler plugins to use this? It should use BASEDIR to lookup the DBCs like it does for the cereal schema with `log.capnp`.

adeebshihadeh:
Any reason this is still a draft? Don't worry about the PlotJuggler side, I can take care of that.

deanlee:
> Any reason this is still a draft? Don't worry about the PlotJuggler side, I can take care of that.

That's great. then i think it's ready for review

adeebshihadeh:
Thanks @deanlee! Continued in #602

PR #378, opened 2021-04-07
Add STEER_RATE_FRACTION

ShaneSmiskol:
Increases accuracy of steer rate, will open a PR to master to utilize this like steer fraction when this is merged.

Here's before and after: https://imgur.com/a/cLWvvcj


ShaneSmiskol:
The scale seems to be -0.8 to 0.7, and I'm not sure if the scale is right. If it's supposed to be 1/15th then shouldn't it be 0.0666 instead of 0.1 for the scale?

pd0wm:
Interesting! Let's make sure we get the scale right, the steer rate is used by INDI so it can have quite an influence.

ShaneSmiskol:
That makes sense. I made sure the current rate's factor of 1 is good, it seems to be if I plot it against the steering rate calculated from the last 0.05 seconds (1 second was too smoothed). I can't quite tell from comparing different factors and offsets, but 0.0666 and 0.0333 for factor and offset seems like it might be appropriate. Here's some comparisons: https://imgur.com/a/XyadGNG

I'm offsetting by the way since the signal sometimes goes to -8 (`1000`, upper limit is still 7) whereas steer rate never does, its scale is -7 to 7.

Aside from the beginning where the wheel isn't moving, the signal usually only is -8 for a frame or two, where it goes to -7 more often. So it could be an outlier, and we don't need to offset the signal. What do you think?

Screenshot_212

rbiasini:
it does not seem that the STEER_RATE_FRACTION is behaving like the derivative of STEER_FRACTION. STEER_RATE_FRACTION shoudl be zero when "STEER_FRACTION" is constant, which does not seem to happen based on your plots.

ShaneSmiskol:
Hmm you're right. And I took a look at the other non-zero bits in the signal and none seemed to act as a derivative. Rate fraction always was a different static value when steer angle fraction stopped moving. I'll see about finding the period the rate is calculated in (it's about .05 to .1 second, and maybe we can just calculate the derivative and use that instead. Since integer steering rates when it's a low number isn't very accurate.

pd0wm:
How does it look at a very small scale? Could be an angular encoder inside of the EPS before it's geared down?

ShaneSmiskol:
rbiasini was right, the more I look at possible rate fraction signals the more it doesn't make sense. But besides that, the current rate signal doesn't make too much sense in some cases.

For example here you would expect the rate signal to start decreasing since the angle is decreasing and the rate increment is 1 deg/sec but it doesn't move until rate is -13, after 16 samples of being zero.

Screenshot_214

ShaneSmiskol:
Closing until I look into the signals more or just calculate the derivative by hand in carstate

PR #379, opened 2021-04-07
define toyota STEER_TORQUE_CMD as u16 instead of i16

andresv:
Currently it is defined as:
```
SG_ STEER_TORQUE_CMD : 15|16@0- (1,0) [0|65535] "" XXX
```

Notice that datatype is `i16`, but min and max values are `[0|65535]`.


andresv:
I wonder why CI failed
```
Run docker run opendbc bash -c "cd opendbc/generator && ./test_generator.py"
Traceback (most recent call last):
  File "./test_generator.py", line 16, in <module>
    test_generator()
  File "./test_generator.py", line 14, in test_generator
    assert len(comp.diff_files) == 0, f"Different files: {comp.diff_files}"
AssertionError: Different files: ['toyota_prius_2017_pt_generated.dbc']
Error: Process completed with exit code 1.
```

What is wrong with `toyota_prius_2017_pt_generated.dbc`?

adeebshihadeh:
The signal is signed - it's the range that's wrong. Feel free to submit a PR to fix the range.

PR #380, opened 2021-04-12
VW PQ: New and updated CAN messaging

jyoung8607:
Used by the Volkswagen community PQ35/PQ46/NMS port. Nothing comma-upstream uses this, so there will not yet be a corresponding openpilot PR.

LDW_1 still needs to be cleaned up and synced with the MQB version, but I'll have to do that in a future PR.


grekiki:
Now I can remove the opendbc fork for cabana exploration :)

PR #387, opened 2021-04-26
fix toyota STEER_TORQUE_CMD min|max

andresv:
`STEER_TORQUE_CMD` is `i16` signal so its full range can be from `-32768` to `32767`.
@adeebshihadeh confirmed in https://github.com/commaai/opendbc/pull/379 that signal itself is indeed `i16` and just min and max were wrong.


andresv:
What is wrong with `toyota_prius_2017_pt_generated.dbc`?

toyota_prius_2017_pt_generated.dbc
```diff
@@ -398,7 +398,7 @@ BO_ 608 STEER_TORQUE_SENSOR: 8 XXX
  SG_ STEER_TORQUE_DRIVER : 15|16@0- (1,0) [-32768|32767] "" XXX
  SG_ STEER_OVERRIDE : 0|1@0+ (1,0) [0|1] "" XXX
  SG_ CHECKSUM : 63|8@0+ (1,0) [0|255] "" XXX
- SG_ STEER_ANGLE : 31|16@0- (0.0573,0) [-500|500] "" XXX
```
adeebshihadeh, line 401: tests are failing on this. fix in source file and re-run generator

adeebshihadeh:
Closing due to inactivity

PR #388, opened 2021-04-27
Prius gear shifter panel signals

cydia2020:
drive mode is of specific interest as op can then adapt tuning based on drive mode. tested on prius


cydia2020:
idk how universal this is, but seems other vehicles have this as GEAR_PACKET

cydia2020:
closing because it's not universal

PR #391, opened 2021-05-07
Initial Avalon 2019+ DBC w/ longer EPS_STATUS (XX50)

nelsonjchen:
These Avalons appear to have a different length for the EPS_STATUS
message.

Without this change, these Avalons will spam "0x262 CHECKSUM FAIL"
in openpilot.

Checksum is still at the end of this message and is algorithmically
unchanged.

Also, FWIW, @rav4kumar called this a 2021 Avalon in his branches/commits
but I'm putting it at 2019+ from reading Wikipedia and observing Discord
Avalon users.

I was too lazy to rebuild the EPS_STATUS definition so I copied
@rav4kumar's.

Upstream, like in openpilot, I expect this to be fingerprinted as
 a different vehicle from the existing "XX40"'s "`AVALON`".

Co-Authored: Kumar <rav4kumar@users.noreply.github.com>


nelsonjchen:
Related: https://github.com/commaai/openpilot/issues/19739

adeebshihadeh:
Closing this - this should be fixed for all Toyota. See https://github.com/commaai/openpilot/issues/19739.

nelsonjchen:
Not all toyotas have the 8 byte EPS_STATUS with checksum at the end though. @bisra2 's IS300H for example is 5 bytes. 

image

Anyway, OK, let's leave this PR here for when someone does roll it up in a PR that encompasses more known Toyotas then.

PR #401, opened 2021-06-04
Different gas pedal signal for Hyundai PHEVs

ShaneSmiskol:
PR goes with: https://github.com/commaai/openpilot/pull/21147

Had someone with PHEV Ioniq test mashing gas to 100%, it reported 254. And From a few routes on the Ioniq EV it reached 254 as well.


ShaneSmiskol:
This might not be needed, will close as soon as I can check a PHEV route

hyundai_kia_generic.dbc
```diff
@@ -1523,7 +1523,8 @@ BO_ 881 E_EMS11: 8 XXX
  SG_ Gear_Change : 12|1@0+ (1,0) [0|31] "" XXX
  SG_ Cruise_Limit_Status : 13|1@1+ (1,0) [0|1] "" XXX
  SG_ Cruise_Limit_Target : 23|8@1+ (1,0) [0|15] "" XXX
- SG_ Accel_Pedal_Pos : 31|8@1+ (1,0) [0|7] "" XXX
+ SG_ Accel_Pedal_Pc : 31|8@1+ (0.390625,0) [0|100] "%" XXX
+ SG_ CR_Vcu_AccPedDep_Pc : 56|8@1+ (0.392156863,0) [0|100] "%" XXX
```
ShaneSmiskol, line 1527: Think these scales are right, `Accel_Pedal_Pc` has 9 bits, `CR_Vcu_AccPedDep_Pc` has 8 bits

hyundai_kia_generic.dbc
```diff
@@ -1646,4 +1646,4 @@ VAL_ 1157 LFA_SysWarning 0 "no_message" 1 "switching_to_hda" 2 "switching_to_scc
 VAL_ 1157 HDA_Icon_State 0 "no_hda" 1 "white_hda" 2 "green_hda";
 VAL_ 1157 HDA_SysWarning 0 "no_message" 1 "driving_convenience_systems_cancelled" 2 "highway_drive_assist_system_cancelled";
 
-CM_ "BO_ E_EMS11: With the exception of the Optima Hybrid, all hybrid and plug-in hybrids use this gas signal: CR_Vcu_AccPedDep_Pc. All EVs (and OptimaH) use the Accel_Pedal_EV_Pc signal. Both are 0 to 100%";
+CM_ "BO_ E_EMS11: With the exception of the Optima Hybrid, all (plug-in) hybrids use this gas signal: CR_Vcu_AccPedDep_Pos. All EVs (and OptimaH) use the Accel_Pedal_Pos signal.";
```
VirtuallyChris, line 1649: Might be better to keep the phrasing more generic incase there ends up being more outliers like the Optima Hybrid

ShaneSmiskol: How's that? Make it point to values.py so you can see which cars use the hev signal

PR #403, opened 2021-06-20
Subaru Global Dash state and LKAS messages

martinl:
This PR adds and fixes signals in Subaru Global ES_DashStatus and ES_LKAS_State messages, mainly so we can filter out some confusing stock ACC and LKAS dash notification messages while openpilot is active. I will open related openpilot and panda PR-s soon


adeebshihadeh:
Can you resolve the conflicts?

PR #412, opened 2021-08-07
Add LFA button and speed limit signal for newer Hyundai cars.

mob9221:
This PR adds the message containing speed limit data that is shown on some Hyundai navigation systems, gauge clusters and heads up displays.

### DIFF
--- a/hyundai_kia_generic.dbc
+++ b/hyundai_kia_generic.dbc
@@ -1489,6 +1489,7 @@ BO_ 1157 LFAHDA_MFC: 4 XXX
 BO_ 913 BCM_PO_11: 8 Vector__XXX
  SG_ BCM_Door_Dri_Status : 5|1@0+ (1,0) [0|1] "" PT_ESC_ABS
  SG_ BCM_Shift_R_MT_SW_Status : 39|2@0+ (1,0) [0|3] "" PT_ESC_ABS
+ SG_ LFA_Pressed : 4|1@0+ (1,0) [0|1] "" XXX
 
 BO_ 1426 LABEL11: 8 XXX
  SG_ CC_React : 34|1@1+ (1,0) [0|1] "" XXX
--- a/hyundai_kia_generic.dbc
+++ b/hyundai_kia_generic.dbc
@@ -1643,7 +1643,6 @@ BO_ 1348 Navi_HU: 8 XXX
 
 
 
-CM_ SG_ 1348 SpeedLim_Nav_Clu "Speed limit displayed on Nav, Cluster and HUD";
 VAL_ 871 CF_Lvr_Gear 5 "D" 8 "S" 6 "N" 7 "R" 0 "P" ;
 VAL_ 1322 CF_Clu_Gear 1 "P" 2 "R" 4 "N" 8 "D" ;
 VAL_ 274 CUR_GR 1 "D" 2 "D" 3 "D" 4 "D" 5 "D" 6 "D" 7 "D" 8 "D" 14 "R" 0 "P" ;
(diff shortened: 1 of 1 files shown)


oricalworks:
Does this work for 2018 Kia Stinger?
I kept getting the max returned value of 255 all the time.

gregjhogan:
I took a look at some routes for a 2020 sonata and 2020 palisade and I couldn't find a place where the value of the first byte of 0x544 changed (always 255 with openpilot running, stock drive for palisade always 0)

A while ago I did find address 0x255 is what the camera uses to determine how to adjust the set speed for HDA mode.  Note that this message sometimes uses the same bytes to mean different things and the speed limit changes are in the future (something in the message somehow indicates how far in the future)
https://github.com/gregjhogan/panda/blob/examples/examples/hda_test.py

mob9221:
> I took a look at some routes for a 2020 sonata and 2020 palisade and I couldn't find a place where the value of the first byte of 0x544 changed (always 255 with openpilot running, stock drive for palisade always 0)
> 
> A while ago I did find address 0x255 is what the camera uses to determine how to adjust the set speed for HDA mode. Note that this message sometimes uses the same bytes to mean different things and the speed limit changes are in the future (something in the message somehow indicates how far in the future)
> https://github.com/gregjhogan/panda/blob/examples/examples/hda_test.py

Hi Greg, I just thought it would be size 8 since it is most likely used for HDA_VSetReq in the dbc which is also 8 bits but I doubt we can find a speed limit sign to push it that far. I tried getting HDA to work again on these cars since its blocked by the panda I believe but just setting HDA_VSetReq didn't do the trick as I thought it would. Thanks for sharing the example I'll dig into it further.

jason-wen-waysun:
Can confirm the new signal for the LFA button is working on the 2021 Hyundai Sonata N Line.

sunnyhaibin:
Can confirm the signal for the LFA button is working on the 2021 Hyundai Sonata N Line.

gregjhogan:
`SpeedLim_Nav_Clu` is the speed limit displayed on the cluster for what vehicles and on what CAN bus?

at one point I had openpilot on my 2020 palisade always putting the car in HDA mode (had longer resume timeout when stopped behind a lead) and I believe with that code just changing `HDA_VSetReq` worked to change the set speed but can't remember for sure (I never had anything automatically changing the speed with speed limits)
https://github.com/gregjhogan/openpilot/commits/hyundai-hda

mob9221:
> `SpeedLim_Nav_Clu` is the speed limit displayed on the cluster for what vehicles and on what CAN bus?
> 
> at one point I had openpilot on my 2020 palisade always putting the car in HDA mode (had longer resume timeout when stopped behind a lead) and I believe with that code just changing `HDA_VSetReq` worked to change the set speed but can't remember for sure (I never had anything automatically changing the speed with speed limits)
> https://github.com/gregjhogan/openpilot/commits/hyundai-hda

I can't fully test which vehicles it is available on but it seems to be any vehicle with HDA and the new nav units from Hyundai (AVN5). I didn't see the HDA_C_State message before, Were these messages  changed / removed in future dbc commits?

gregjhogan:
2020 sonata and palisade limited have AVN5, correct?  I am not seeing anything in this message change (even looked at a stock palisade drive)

mob9221:
I drive the 2020 sonata limited. Here is the cabana link. Have a look at Navi_HU starting around 195s. 

https://my.comma.ai/cabana/?route=6bc7319847dbb610%7C2021-08-17--20-40-36&seekTime=1&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F6bc7319847dbb610%2F6ceacc1a122001e70bd8f8869920e392_2021-08-17--20-40-36

mob9221:
Hi @gregjhogan , could you explain what the data being sent in 
https://github.com/gregjhogan/panda/blob/examples/examples/hda_test.py
represents?

b"\x00\x00" + bytes([idx]) + b"\x00\x00\x30\x12\x6c"
This is the message being sent for HDA 55 mph in that example, but not sure which combination of bytes represent the actual speed.

gregjhogan:
I have not found a way to get `SpeedLim_Nav_Clu` to output the speed limit on my vehicle :(

In the example script I believe the last two bytes were the speed limit (in some strange format) and the other bytes include a message type (address 0x255 is used for several messages of different types) and how far ahead the speed limit changes.  I believe the examples all set the distance ahead where the speed limit changes to zero (never figured out the distance format)

The way I tested it was by putting a panda in between the camera and the rest of the car (no openpilot) which forwarded all messages except 0x255, then the test script was used to inject 0x255 messages ad-hoc (sent to the camera) causing the camera to change the speed limit using `HDA_VSetReq`.

james5294:
https://streamable.com/vh60oa

2022 Sonata seems to be getting Nav speed data, my 2021 Palisade still reads 0. Both n FW from june/july 2022.

PR #414, opened 2021-08-20
Subaru ACC speed units signals update

martinl:
This PR adds new Dashlights UNITS signal for Subaru Global models and changes Preglobal Dash_State2 UNITS signal. Both have been in use in subaru-community branch since beginning of July and they have been working reliably for all supported models.


adeebshihadeh:
No need to delete the old messages. Let's just add the new signals.

PR #417, opened 2021-08-22
clean up CANDefine constructor

belm0:
explicit `d[foo] = {}` not needed if `d` is `defaultdict(dict)`

### DIFF
--- a/can/parser_pyx.pyx
+++ b/can/parser_pyx.pyx
@@ -175,16 +175,15 @@ cdef class CANDefine():
       val = self.dbc[0].vals[i]
 
       sgname = val.name.decode('utf8')
-      address = val.address
       def_val = val.def_val.decode('utf8')
+      address = val.address
+      msgname = address_to_msg_name[address]
 
-      #separate definition/value pairs
+      # separate definition/value pairs
       def_val = def_val.split()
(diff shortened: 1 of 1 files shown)


can/parser_pyx.pyx
```diff
@@ -184,9 +184,7 @@ cdef class CANDefine():
       defs = def_val[1::2]
 
       if address not in dv:
-        dv[address] = {}
         msgname = address_to_msg_name[address]
-        dv[msgname] = {}
```
belm0, line 189: `dv` is `defaultdict(dict)` and these key are accessed below, so explicit init to `dict` is not needed

can/parser_pyx.pyx
```diff
@@ -184,9 +184,7 @@ cdef class CANDefine():
       defs = def_val[1::2]
 
       if address not in dv:
-        dv[address] = {}
         msgname = address_to_msg_name[address]
```
belm0, line 187: By the way, this looks like a bug.    The code below references `msgname` even when this `if` block is not taken, so it will get the previous iteration's value.  (If that's really the intention, I'll add a comment about it.)

adeebshihadeh: yeah, looks like a bug

PR #418, opened 2021-08-22
Do not reverse multiple includes order in dbc templates

martinl:
This PR fixes generated dbc files with multiple includes which had message and signal definitions before dbc header


generator/generator.py
```diff
@@ -23,7 +23,7 @@ def create_dbc(src_dir, filename, output_path):
   with open(output_file_location, 'w') as dbc_file_out:
     dbc_file_out.write('CM_ "AUTOGENERATED FILE, DO NOT EDIT";\n')
 
-    for include_filename in reversed(includes):
+    for include_filename in includes:
```
adeebshihadeh, line 26: @pd0wm any idea why this was reversed?

PR #419, opened 2021-08-24
expand Honda Fit DBC to support HR-V and Freed

belm0:
* add GAS_PEDAL message (HR-V)
  * add DRIVERS_DOOR_OPEN field to SCM_BUTTONS (HR-V and Freed)
  * honda_hrv_touring_2019_can is retired

Since these models are so close (HR-V and Freed are based on the Fit platform, per wikipedia), it seems better to avoid maintaining three DBC files.


belm0:
corresponding openpilot PR:  https://github.com/commaai/openpilot/pull/22003

adeebshihadeh:
Nice cleanup! I wonder if we can further cleanup the difference in `DRIVERS_DOOR_OPEN`.

belm0:
Since Freed port PR may be delayed, shall I make a separate openpilot PR just to sync opendbc and fix the HR-V dbc reference?

adeebshihadeh:
Already done #22086

PR #420, opened 2021-08-29
2020 Honda e

vanillagorillaa:
DBC for 2020 Honda e, Bosch vehicle. 

New DBC is needed due to Gear Shifter values are different than normal Bosch vehicles. Does match the shifter values of the Insight, but can't use the Insight DBC as other signals, such as Steer Status, do not match to this vehicle.


adeebshihadeh:
There's likely a common signal across the Bosch models we can use. I'd like to avoid adding a new DBC if possible.

belm0:
> I'd like to avoid adding a new DBC if possible. For example, I see a signal called `GEAR_SHIFTER` in the 2018 Accord DBC that seems to have the same values as the `GEAR_SHIFTER` here, although on different addresses.

I think avoiding DBC forks is the right idea (I tried to do that with https://github.com/commaai/opendbc/pull/419).

Perhaps add a `GEAR_SHIFTER_2` to the existing DBC, and put the switch logic in openpilot.  Such a pattern already exists, e.g. see  `GAS_PEDAL` / `GAS_PEDAL_2`.

vanillagorillaa:
> There's likely a common signal across the Bosch models we can use. I'd like to avoid adding a new DBC if possible. For example, I see a signal called `GEAR_SHIFTER` in the 2018 Accord DBC that seems to have the same values as the `GEAR_SHIFTER` here, although on different addresses.

`GEAR_SHIFTER` in the accord DBC is `VAL_ 401 GEAR_SHIFTER 32 "L" 16 "S" 8 "D" 4 "N" 2 "R" 1 "P" ;` where the Honda E values are `VAL_ 419 GEAR_SHIFTER 32 "L" 16 "S" 1 "D" 4 "N" 10 "R" 0 "P" ;` The P/R/D values are different between the two. 

There is a `GEAR` message in a lot of the Honda Vehicles. The ones that have it are all the same. The vehicle DBC's listed below don't have this GEAR value defined, not sure if it's been left out, or not there. Will need some drives from these vehicles to determine this. Not sure if this is a suitable replacement in openpilot as well. It looks like the Nidec vehicles use these as well. 

honda_clarity_hybrid_2018_can_generated.dbc
honda_crv_executive_2016_can_generated.dbc
honda_fit_hybrid_2018_can_generated.dbc
honda_insight_ex_2019_can_generated.dbc
honda_crv_hybrid_2019_can_generated.dbc
honda_odyssey_exl_2018_generated.dbc
honda_pilot_touring_2017_can_generated.dbc
honda_ridgeline_black_edition_2017_can_generated.dbc
honda_civic_sedan_16_diesel_2019_can_generated.dbc

belm0:
> The vehicle DBC's listed below don't have this GEAR value defined, not sure if it's been left out, or not there. Will need some drives from these vehicles to determine this.

Is it OK in practice to have a DBC define a message and/or field that isn't used for a particular model?  I haven't heard any guidelines on this.  Allowing / encouraging that may go a long way towards keeping the DBC file count reasonable (by way of more monolithic, per-make files).

vanillagorillaa:
All the vehicles above have a gear signal defined in the DBC, just not a value definition.

adeebshihadeh:
> Is it OK in practice to have a DBC define a message and/or field that isn't used for a particular model? I haven't heard any guidelines on this. Allowing / encouraging that may go a long way towards keeping the DBC file count reasonable (by way of more monolithic, per-make files).

Yes, this is a pretty common in the DBCs for other makes.

vanillagorillaa:
It may be possible to just use the 2020 RDX DBC for the Honda e without the need for any modifications. Testing now

vanillagorillaa:
closing, not needed. While the "gear shifter" value in this DBC does not match the parsing on other vehicles, we can use the 2020 RDX dbc as the "gear shifter" signal is in different bytes, and the values on both vehicles match.

PR #422, opened 2021-09-03
Add PQ Braking and ACC led control CAN messages

CineXWorm:
mMotor_Bremse used to control braking on cars that don't support ACC.
mAWV used to control ACC led in dashboard.

### DIFF
--- a/vw_golf_mk4.dbc
+++ b/vw_golf_mk4.dbc
@@ -1139,7 +1139,7 @@ BO_ 210 HCA_1: 5 XXX
  SG_ Vib_Freq : 32|4@1+ (4,0) [0|60] "Hz" XXX
  SG_ Vib_Amp : 36|4@1+ (0.5,0) [0|7.5] "Nm" XXX
  
-BO_ 644 mMotor_Bremse: 6 Motor_Diesel
+BO_ 644 Motor_Bremse: 6 XXX
  SG_ MOB_Standby : 12|1@1+ (1,0) [0|1] "" Bremse_MK25AESP
  SG_ MOB_Freigabe : 14|1@1+ (1,0) [0|1] "" BCM_Gateway,Bremse_MK25AESP,Gateway_separat
  SG_ MOB_Anhaltewunsch : 13|1@1+ (1,0) [0|1] "" Bremse_MK25AESP,Getriebe_DQ
--- a/vw_golf_mk4.dbc
+++ b/vw_golf_mk4.dbc
@@ -1138,6 +1138,39 @@ BO_ 210 HCA_1: 5 XXX
  SG_ LM_OffSign : 31|1@1+ (1,0) [0|1] "" XXX
  SG_ Vib_Freq : 32|4@1+ (4,0) [0|60] "Hz" XXX
  SG_ Vib_Amp : 36|4@1+ (0.5,0) [0|7.5] "Nm" XXX
+ 
+BO_ 644 mMotor_Bremse: 6 Motor_Diesel
+ SG_ MOB_Standby : 12|1@1+ (1,0) [0|1] "" Bremse_MK25AESP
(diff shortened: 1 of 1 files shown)


grekiki:
Would help to get this merged into openpilot to be able to tune pq long with plotjuggler, seems like it doesn't take local DBCs

grekiki:
Would be really nice to have this merged for using Plotjuggler for tuning long control on PQ. Seems like plotjuggler doesn't use the local DBC(or I am just doing something wrong) @adeebshihadeh

adeebshihadeh:
LGTM. any comments @jyoung8607?

@grekiki yeah, it should use the local DBC

vw_golf_mk4.dbc
```diff
@@ -1138,6 +1138,39 @@ BO_ 210 HCA_1: 5 XXX
  SG_ LM_OffSign : 31|1@1+ (1,0) [0|1] "" XXX
  SG_ Vib_Freq : 32|4@1+ (4,0) [0|60] "Hz" XXX
  SG_ Vib_Amp : 36|4@1+ (0.5,0) [0|7.5] "Nm" XXX
+ 
+BO_ 644 mMotor_Bremse: 6 Motor_Diesel
```
jyoung8607, line 1142:
```suggestion
BO_ 644 Motor_Bremse: 6 XXX
```

The prefixing `m` on message names is unique to the source of this data, we have been removing them.

vw_golf_mk4.dbc
```diff
@@ -1138,6 +1138,39 @@ BO_ 210 HCA_1: 5 XXX
...
+ SG_ MOB_Freigabe : 14|1@1+ (1,0) [0|1] "" BCM_Gateway,Bremse_MK25AESP,Gateway_separat
+ SG_ MOB_Anhaltewunsch : 13|1@1+ (1,0) [0|1] "" Bremse_MK25AESP,Getriebe_DQ
+ SG_ MOB_CHECKSUM : 0|8@1+ (1,0) [0|255] "" Bremse_MK25AESP
+ SG_ MOB_COUNTER : 8|4@1+ (1,0) [0|15] "" Bremse_MK25AESP
+ SG_ TSK_v_Begrenzung_aktiv : 15|1@0+ (1,0) [0|1] "" XXX
+ SG_ TSK_ax_Getriebe_01 : 40|8@1+ (0.048,0) [0|255] "m/s2" XXX
+ SG_ MOB_Bremsstgr : 16|11@1+ (0.76,0) [0|100] "Unit_PerCent" Vector__XXX
+ SG_ MOB_Bremsmom : 27|13@1+ (1,0) [0|8190] "Unit_NewtoMeter" Bremse_MK25AESP
+
+BO_ 870 mAWV: 5 ACC
```
jyoung8607, line 1153:
```suggestion
BO_ 870 AWV: 5 XXX
```

Same as above

vw_golf_mk4.dbc
```diff
@@ -1138,6 +1138,39 @@ BO_ 210 HCA_1: 5 XXX
  SG_ LM_OffSign : 31|1@1+ (1,0) [0|1] "" XXX
  SG_ Vib_Freq : 32|4@1+ (4,0) [0|60] "Hz" XXX
  SG_ Vib_Amp : 36|4@1+ (0.5,0) [0|7.5] "Nm" XXX
+ 
+BO_ 644 mMotor_Bremse: 6 Motor_Diesel
+ SG_ MOB_Standby : 12|1@1+ (1,0) [0|1] "" Bremse_MK25AESP
```
jyoung8607, line 1143: [ example only ]

```suggestion
 SG_ MOB_Standby : 12|1@1+ (1,0) [0|1] ""  XXX
```

Let's lose the network-node source/consumer labels unless and until we're willing to spend the time overhauling the file to make good use of them. No real sense in adding a few PQ25-specific dangling loose references without the other DBC infrastructure for DBC editors to even make use of them. I admit to having done this inconsistently in the past but we'll clean it all up eventually.

jyoung8607 (changes requested):

PR #424, opened 2021-09-13
Stellantis: RAM 1500 DBC, wave 1

jyoung8607:
For 2019-current 5th gen RAM 1500 light-duty trucks, and hopefully including the 2022+ JEEP Wagoneer/Grand Wagoneer.

Work-in-progress cleanup and revalidation of Tunder's original discovery work on this port. This is still a fast-moving target.


jyoung8607:
@adeebshihadeh can I have this, and bumps for openpilot and the Panda dockerfile? The DBC isn't anywhere near finished but the parts we need for GitHub CI to complete on Panda are in pretty good shape. I'll file another PR to keep going.

PR #428, opened 2021-09-19
Bugfix in dbc CAN encoder

MarcWim:
While encoding, we need to subtract the offset first before we apply the factor.

### DIFF
--- a/can/dbc.py
+++ b/can/dbc.py
@@ -148,7 +148,7 @@ class dbc():
       ival = dd.get(s.name)
       if ival is not None:
 
-        ival = (ival / s.factor) - s.offset
+        ival = (ival - s.offset) / s.factor
         ival = int(round(ival))
 
         if s.is_signed and ival < 0:


rbiasini:
Marc, you are right. A test to this file should probably be added to encode/decode a signal with both non-zero offset and non-unit factor.

adeebshihadeh:
Thanks for reviewing @rbiasini :)

We'll also need to verify that no existing signals used by openpilot rely on the current behavior.

can/dbc.py
```diff
@@ -264,9 +268,14 @@ def get_signals(self, msg):
...
    assert decoded == msg
+   msg = ('WHEEL_SPEEDS', {'WHEEL_SPEED_FR': 100.00, 'WHEEL_SPEED_FL': 100.00, 
+                           'WHEEL_SPEED_RR': 166.99, 'WHEEL_SPEED_RL': 166.99})
+   encoded = dbc_test.encode(*msg)
+   decoded = dbc_test.decode((0xAA, 0, encoded))
+   assert_message_equal(decoded, msg, decimal=10)
 
-   dbc_test = dbc(os.path.join(DBC_PATH, 'hyundai_santa_fe_2019_ccan.dbc'))
-   decoded = dbc_test.decode((0x2b0, 0, "\xfa\xfe\x00\x07\x12"))
+   dbc_test = dbc(os.path.join(DBC_PATH, 'hyundai_2015_ccan.dbc'))
```
MarcWim, line 277: The 2019 file does not seem to exist, this is why I changed it to the 2015 one

MarcWim:
> Marc, you are right. A test to this file should probably be added to encode/decode a signal with both non-zero offset and non-unit factor.

Done. Maybe there is a more elegant way than I did it?

PR #451, opened 2021-10-19
Follow Distance Button Signal Forwarded from sDSU

krkeegan:
The sDSU forwards the signal from the follow distance button:
https://github.com/wocsor/panda/commit/b5120f6427551345c543b490fe47da189c1e48e1

As a community supported feature, I am not sure whether sDSU messages are allowed in the repository or not.


adeebshihadeh:
DBCs for community supported hardware are fine, but they don't go in the main DBCs. This should be moved to `generator/toyota/_comma.dbc`.

krkeegan:
That makes perfect sense, I should have realized that.

I pushed an an additional commit which moves the definitions to the correct file.

PR #453, opened 2021-10-20
Add Toyota Prius Alpha dbc file

CT921:


adeebshihadeh:
We've done some cleanup and now this is the same as `toyota_rav4_2017_pt.dbc`.

PR #457, opened 2021-10-25
Toyota: expand toyota ACC_CONTROL signals

cydia2020:


cydia2020:
dirty_radar_message

Dirty radar message looks like this if you're interested.

cydia2020:
The 29th bit seems to be always set to 1 on the stock system. I still haven't figured out what it does exactly. In the reference DBC it's named as `LVSTP`.

STP might corresponds with Stop, but I'm not sure.

pd0wm:
Updated `openpilot` to not change the bytes sent for now. We can then change the behavior in https://github.com/commaai/openpilot/pull/22678 after @adeebshihadeh  has some time to test.

PR #462, opened 2021-11-02
GM: use steer command units for driver, eps torque

qadmus:
These aren't Newton-meters, and it makes life easier to stick to the same unit for now: steer command units, 0-300.

openpilot PR: https://github.com/commaai/openpilot/pull/22776


adeebshihadeh:
Why don't you think this is Newton-meters? This is likely the torque applied before the EPS assist.

qadmus:
@adeebshihadeh okay, wheel-torque Nm is the most sensible explaination, I just haven't seen it officially labeled.

I would like all of GM lateral to have same units. That desire could be useless, let me know. Would openpilot prefer Nm, steer units [-300,300], leave it alone, or another fix?

I think there is some logic in carcontroller that depends a steer command of integers, like rate limits. Panda also uses steer units natively.

adeebshihadeh:
Yes, it's nice to have real units in openpilot. The raw values are used in panda in most of the safety modes, but feel free to scale them if you wish for GM.

PR #463, opened 2021-11-03
VW PQ: Corrections to Lenkhilfe_3

jyoung8607:
Two fixes, prerequisites for commaai/panda#762.

1. Fix the overall message length, which was easy to find
2. Fix invalid endianness on the counter signal, which took considerably longer to find

The second problem caused some very strange CAN packer output. ISTM adding an assert somewhere might be a good idea.


adeebshihadeh:
I'd like to write a DBC linter at some point that should catch some of this stuff. We could also use the openpilot test routes to validate some things like message length.

PR #472, opened 2021-11-19
Add Lexus RC

ErichMoraga:
This related to https://github.com/commaai/openpilot/pull/22941
...in case @pd0wm wants me to separate/differentiate Lexus RC from IS (which I'm fully prepared to do).


adeebshihadeh:
What's unique about this DBC?

PR #473, opened 2021-11-19
ford: speed, brake and steering pinion

incognitojam:


incognitojam:
Would it be preferable to make separate PRs as I go along making use of these... or should I just create a PR with the leaked DBC file? :slightly_smiling_face:

adeebshihadeh:
Single PR is fine

PR #477, opened 2021-12-06
GM: Add Friction Brake, AEB messages

JasonJShuler:
Adds the messages used by Front-facing cameras for AEB, as well as the friction brake command used for camera-based ACC.


gm_global_a_powertrain.dbc
```diff
@@ -147,6 +147,20 @@ BO_ 489 EBCMVehicleDynamic: 8 K17_EBCM
  SG_ YawRate : 35|12@0- (0.625,0) [0|1] "" NEO
  SG_ YawRate2 : 51|12@0- (0.0625,0) [-2047|2047] "grad/s" NEO
 
+BO_ 512 GAS_COMMAND: 6 NEO
```
adeebshihadeh, line 150: these should go in their own DBC that gets included like the honda and toyota

JasonJShuler: That is a sizeable change. I will remove these from this PR

JasonJShuler: Removed.

PR #478, opened 2021-12-07
GM: use generator + add pedal messages

JasonJShuler:
Created base generator setup for GM - at present there are no documented incompatible divergences, so there is only one output.

File is renamed in OP and Panda repos in the following PRs

Panda PR: https://github.com/commaai/panda/pull/835
OP PR: https://github.com/commaai/openpilot/pull/23614


generator/gm/gm_global_a_powertrain.dbc
```diff
@@ -0,0 +1,2 @@
+CM_ "IMPORT _gm_global_a_powertrain.dbc";
+CM_ "IMPORT _comma.dbc";
```
adeebshihadeh, line 2: just import the comma DBC into the GM one. no need for an intermediate GM one

JasonJShuler: I didn't see any examples like that - Should I put the import at the end of the file, beginning or somewhere in the middle?

JasonJShuler:
@adeebshihadeh what would be the correct process to switch OP over to using the generated dbc then delete the old file from this repo?
Does it need to be done in steps, or should delete the file in this PR and create a corresponding PR in OP to switch the dbc names?

adeebshihadeh:
Usually, you'll make both an opendbc and openpilot PR with all the necessary changes. Then, I'll merge the opendbc and bump the opendbc submodule in your openpilot PR to opendbc master. So the only things left to do for this PR are:
* remove the old DBC
* open an openpilot PR to use the new DBC

JasonJShuler:
> Usually, you'll make both an opendbc and openpilot PR with all the necessary changes. Then, I'll merge the opendbc and bump the opendbc submodule in your openpilot PR to opendbc master. So the only things left to do for this PR are:
> 
>     * remove the old DBC
> 
>     * open an openpilot PR to use the new DBC

All steps should be complete. Panda tests reference the dbc name so there is a PR there as well.

Thanks!

generator/gm/gm_global_a_powertrain.dbc
```diff
@@ -276,3 +276,5 @@ VAL_ 715 GasRegenCmdActive 1 "Active" 0 "Inactive" ;
 VAL_ 320 Intellibeam 1 "Active" 0 "Inactive" ;
 VAL_ 320 HighBeamsActive 1 "Active" 0 "Inactive" ;
 VAL_ 320 HighBeamsTemporary 1 "Active" 0 "Inactive" ;
+
+CM_ "IMPORT _comma.dbc";
```
adeebshihadeh, line 280: imports should go on top

PR #481, opened 2021-12-19
2022 Honda Civic DBC

vanillagorillaa:
Three messages (so far) are different on the 2022 Civic than previous Bosch vehicles. Update generator files accordingly.


generator/honda/honda_civic_sedan_16_diesel_2019_can.dbc
```diff
@@ -18,11 +18,47 @@ BO_ 432 STANDSTILL: 7 VSA
  SG_ COUNTER : 53|2@0+ (1,0) [0|3] "" EON
  SG_ CHECKSUM : 51|4@0+ (1,0) [0|15] "" EON
 
+BO_ 479 ACC_CONTROL: 8 EON
```
adeebshihadeh, line 21: don’t copy these into individual car DBCs. shared definitions should only be defined once

vanillagorillaa: Don't think this is possible. On Bosch vehicles ACC_CONTROL is on ID 479 while the new 2022 Civic is 477. The LKAS_HUD message on Bosch vehicles is 5 bytes while the 2022 Civic is 8 bytes (so checksum and counter messages move).

adeebshihadeh: there’s two options here: give the new one a different name or split up the shared DBC into two different files. not sure which I prefer yet, depends on whether the new message is exactly the same but on a different address

ZgLijian: 中国的思域 同款车型 INTEGRA 是否支持

ZgLijian: Does China's civic intergra support the same model

ZgLijian:
Does China's civic intergra support the same model

PR #482, opened 2022-01-04
CANDefine: make various signal info from DBC accessible

sshane:
My first time with Cython and didn't ever work with this code before, but trying to resolve https://github.com/commaai/openpilot/issues/2514. Is this somewhere on the right track of what was desired?

### DIFF
--- a/can/parser_pyx.pyx
+++ b/can/parser_pyx.pyx
@@ -195,6 +195,7 @@ cdef class CANDefine():
     for i in range(self.dbc[0].num_msgs):
       msg = self.dbc[0].msgs[i]
       msg_name = msg.name.decode('utf8')
+      address = msg.address
 
       for x in range(msg.num_sigs):
         sig = msg.sigs[x]
--- a/can/parser_pyx.pyx
+++ b/can/parser_pyx.pyx
@@ -32,24 +32,22 @@ cdef class CANParser:
     dict ts
     bool can_valid
     int can_invalid_cnt
-    int toyota_safety_param
+    int toyota_eps_scale
 
   def __init__(self, dbc_name, signals, checks=None, bus=0, enforce_checks=True):
     if checks is None:
       checks = []
-    self.can_valid = True
     self.dbc_name = dbc_name
     self.dbc = dbc_lookup(dbc_name)
     if not self.dbc:
       raise RuntimeError(f"Can't find DBC: {dbc_name}")
+
     self.vl = {}
     self.ts = {}
-
+    self.can_valid = True
     self.can_invalid_cnt = CAN_INVALID_CNT
-    self.toyota_safety_param = -1
+    self.toyota_eps_scale = -1
 
-    cdef int i
-    cdef int x
     for i in range(self.dbc[0].num_msgs):
       msg = self.dbc[0].msgs[i]
       name = msg.name.decode('utf8')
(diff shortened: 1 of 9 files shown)


can/parser_pyx.pyx
```diff
@@ -45,10 +46,11 @@ cdef class CANParser:
     self.ts = {}
 
     self.can_invalid_cnt = CAN_INVALID_CNT
+    self.toyota_safety_param = -1
 
     cdef int i
-    cdef int num_msgs = self.dbc[0].num_msgs
-    for i in range(num_msgs):
+    cdef int x
```
sshane, line 52: Not sure if this is needed, it works without defining them beforehand

can/parser_pyx.pyx
```diff
@@ -59,6 +61,13 @@ cdef class CANParser:
       self.ts[msg.address] = {}
       self.ts[name] = {}
 
+      # store torque scaling if available
+      if name == "STEER_TORQUE_SENSOR":
```
sshane, line 65: Can also check the address

can/parser_pyx.pyx
adeebshihadeh: probably done to keep the code in C. you can check by generating a Cython report

sshane: I checked both with and without the cdef, the generated code for the for loop looks the same.

adeebshihadeh:
No, this isn't quite right. The opendbc PR shouldn't have anything to do with Toyota at all. All that's needed on the opendbc side is a generic way to get signal parameters, like offset, scale, etc.

sshane:
Yeah that would be the better idea. Let me know how this is, I chose to add an `info` dict like the existing vl and ts variables, and added what was available in the `Signal` struct. Also can add this functionality to CANDefine if that makes more sense

can/parser_pyx.pyx
```diff
@@ -189,4 +190,18 @@ cdef class CANDefine():
       dv[address][sgname] = dict(zip(values, defs))
       dv[msgname][sgname] = dv[address][sgname]
 
-      self.dv = dict(dv)
+    # Signal information
```
sshane, line 193: Not all signals are available in vals, plus you can't easily get msg name from val.

can/parser_pyx.pyx
```diff
@@ -169,8 +170,8 @@ cdef class CANDefine():
       address = msg.address
       address_to_msg_name[address] = name
 
+    # Value definitions
```
sshane, line 173: Not all signals are available in vals, plus you can't easily get signal from val.

jyoung8607: Can we just build v->d lookup tables to go along with the d->v tables? I looked at this awhile back because I have [some lists](https://github.com/commaai/openpilot/blob/fc8d402713f1b84b3e75067f09794236852f4f20/selfdrive/car/volkswagen/values.py#L48-L58) (with several more to come for long control) that I'd rather maintain in the DBC, but never got a round tuit.

sshane: Ah I see, yeah that shouldn't be too hard. Probably would want to make it another PR though. I can open one

jyoung8607: I started looking at this real briefly back in December, I thought it was cleaner to have explicit separate maps:

```
      vd[address][sgname] = dict(zip(defs, values))
      vd[msgname][sgname] = vd[address][sgname]

      self.vd = dict(vd)
```

But I never got to the point of testing it or using it in my car port work.

sshane: I agree it might look a little better to use, but it's so enticing being a one line change, plus I think it might make equal sense to have them in the same dict. And `dv` is already backwards...

sshane:
Not needed for anything yet

PR #485, opened 2022-01-08
Toyota: cleanup opendbc

cydia2020:
Clean up Toyota's generator and generated files.

1. Removed `DSU_CRUISE` from `_toyota_2017.dbc`, and moved it to `lexus_is_2018_pt.dbc` since it does not make sense to have this signal present on other vehicles, this signal is not `DSU_CRUISE` on vehicles except the Lexus IS and RC.
2. Removed all instances of `kph` and renamed them `km/h`.
3. Unified the expression of units, for example, `deg/sec` was renamed to `deg/s`, and all instances of `m/s2` were renamed to `m/s^2`.
4. Arranged most signals in numerical orders.
5. Formatted and cleaned up `_comma.dbc`.
6. Renamed `SEATS_DOORS` to `BODY_CONTROL_STATE`, since now we know that this signal controls a lot of features on the combination meter display.
7. Renamed `STEERING_LEVERS` to `BLINKERS_STATE`, this signal does not only indicate the blinker stalk's position, it also indicates whether `HAZARD_LIGHT`, so it makes sense to use this name.
8. Also some other stuff that's not worth mentioning.

commaai/openpilot#23452


adeebshihadeh:
Please break this up into the following PRs:
* sorting by address
* message/signal renamings
* non-message/signal renamings

Also please include a description of the unmentioned changes. It's completely unclear from this massive.

cydia2020:
> message/signal renamings

#486

PR #486, opened 2022-01-08
Toyota: signal names cleanup

cydia2020:
### Things that might affect openpilot
- STEERING_LEVERS -> BLINKERS_STATE (seems to be from the combination meter circuit plate's flasher circuit).
- SEATS_DOORS -> BODY_CONTROL_STATE (not only seat and doors, but a lot of bcm signals that the combination meter circuit plate would get)
- remove DSU_CRUISE from _toyota_2017.dbc and move it to the Lexus IS generator, as it is only used on the IS and RC, the signal doesn't make sense anywhere else.

### Things that wouldn't affect openpilot
- BRAKE_MODULE2 -> BRAKE_2 (keeping expressions unified, this signal isn't used in openpilot)
- CENTRAL_GATEWAY_UNIT -> CERTIFICATION_ECU (signal comes from gateway but it's actually sent by the certification ecu, this signal isn't used in openpilot)


cydia2020:
Doing this part first as this is the part that affects openpilot, so commaai/openpilot/pull/23452 can be merged.

adeebshihadeh:
I moved DSU_CRUISE back since I'd like to eventually use a single, or as few as possible, DBCs for Toyota.

PR #497, opened 2022-01-19
Toyota: LKAS status on gauge cluster

zorrobyte:
This signal is the LKAS status that is controlled by the steering wheel button. Such can be used for things like toggling OP lateral so a user can use stock or OP cruise without OP's lateral control by pushing the LKAS steering wheel button.

### DIFF
--- a/toyota_prius_2017_pt_generated.dbc
+++ b/toyota_prius_2017_pt_generated.dbc
@@ -231,6 +231,11 @@ BO_ 1409 VIN_PART_2: 8 CGW
 BO_ 1410 VIN_PART_3: 8 CGW
  SG_ VIN_17 : 7|8@0+ (1,0) [0|0] "" XXX
 
+BO_ 1552 BODY_CONTROL_STATE_2: 8 XXX
+ SG_ METER_SLIDER_BRIGHTNESS_PCT : 30|7@0+ (1,0) [12|100] "%" XXX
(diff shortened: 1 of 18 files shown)


zorrobyte:
Not sure why the build is failing, what did I miss?

toyota_prius_2017_pt_generated.dbc
```diff
@@ -451,7 +451,7 @@ BO_ 608 STEER_TORQUE_SENSOR: 8 XXX
  SG_ STEER_TORQUE_DRIVER : 15|16@0- (1,0) [-32768|32767] "" XXX
  SG_ STEER_OVERRIDE : 0|1@0+ (1,0) [0|1] "" XXX
  SG_ CHECKSUM : 63|8@0+ (1,0) [0|255] "" XXX
- SG_ STEER_ANGLE : 31|16@0- (0.0573,0) [-500|500] "" XXX 
+ SG_ STEER_ANGLE : 31|16@0- (0.0573,0) [-500|500] "" XXX
```
sshane, line 454: Is this change from an IDE?

zorrobyte: Yes, Atom

sshane: Make sure it's reverted, you don't want to edit any of the generated files manually

zorrobyte: Derp, I didn’t run the generator

PR #464, opened 2021-11-04
Toyota: use common gear packet for all cars

sshane:
- [x] Renamed 295 from `GEAR_PACKET` to `GEAR_PACKET_HYBRID`.
- [x] Moved `956` `GEAR_PACKET` from nodsu to _toyota_2017.
- [x] Add more signals to 956.
- [x] Merged nodsu DBCs.

Verified cars:
- [x] Lexus RX Hybrid TSS2 - `46a49a1b7b23fb4c|2022-01-22--18-05-38--0`
- [x] CHR Hybrid - `29c9bda9ab685a97|2022-01-20--17-34-58--32`
- [x] Camry Hybrid - `54e1134ab4d8afd0|2022-01-21--13-49-04--0`
- [x] [Camry Hybrid TSS2 - `f15e3c37c118e841|2022-01-19--00-20-15`](https://user-images.githubusercontent.com/25857203/150352218-4cb95ac1-ed81-4d58-8337-0d4c751c3686.png)
- [x] Highlander Hybrid TSS2 - `08a8bc466a90e993|2022-01-22--09-06-37--10`
- [x] Avalon Hybrid 2019 - `542604acd4799578|2022-01-20--08-42-45--11`
- [x] Corolla Hybrid TSS2 - `27fd0c08bd1df4ca|2022-01-25--20-53-16`
- [x] Lexus ES Hybrid TSS2 - `abc09032f402f271|2022-01-21--22-40-06--24`
- [x] [RAV4 Hybrid TSS2 - `65961b58358e467a|2022-01-19--10-50-19`](https://user-images.githubusercontent.com/25857203/150352101-d99a4c5d-7a40-4253-b488-68266fa21828.png)
- [x] [Prius TSS2 - `b7e555498879d5ad|2022-01-20--08-59-19`](https://user-images.githubusercontent.com/25857203/150435555-2abdf287-fc83-4feb-9612-24ef87a7db60.png)
- [x] [Prius - `399b1cca765e80b4|2022-01-19--04-53-33`](https://user-images.githubusercontent.com/25857203/150079780-06b35e45-d3e0-46ec-8f87-274b067470dd.png)
- [x] Mirai - `3e30822a7b06be2e|2022-01-21--10-34-01`

PR #498, opened 2022-01-19
Toyota: move gear packet to common dbc

cydia2020:
All supported cars have a common gear packet located on `956`, not sure why this wasn't used on all cars in the first place.

Tested on my Prius and @sshane 's no DSU Hybrid


adeebshihadeh:
Let's do this in @sshane's existing #464

PR #499, opened 2022-01-19
Toyota: LKAS status on gauge cluster

zorrobyte:
This signal is the LKAS status that is controlled by the steering wheel button.

Such could be used for things like toggling OP lateral so a user can use stock or OP cruise without OP's lateral control by pushing the LKAS steering wheel button.

### DIFF
--- a/generator/toyota/_toyota_2017.dbc
+++ b/generator/toyota/_toyota_2017.dbc
@@ -182,7 +182,7 @@ BO_ 1042 LKAS_HUD: 8 XXX
  SG_ RIGHT_LINE : 3|2@0+ (1,0) [0|3] "" XXX
  SG_ LEFT_LINE : 5|2@0+ (1,0) [0|3] "" XXX
  SG_ LKAS_STATUS : 7|2@0+ (1,0) [0|3] "" XXX
- SG_ SET_ME_X01_2 : 11|2@0+ (1,0) [0|3] "" XXX
+ SG_ SET_ME_X01 : 11|2@0+ (1,0) [0|3] "" XXX
  SG_ LDA_ALERT : 9|2@0+ (1,0) [0|3] "" XXX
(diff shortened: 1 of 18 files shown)


generator/toyota/_toyota_2017.dbc
```diff
@@ -181,7 +181,7 @@ BO_ 1042 LKAS_HUD: 8 XXX
  SG_ BARRIERS : 1|2@0+ (1,0) [0|3] "" XXX
  SG_ RIGHT_LINE : 3|2@0+ (1,0) [0|3] "" XXX
  SG_ LEFT_LINE : 5|2@0+ (1,0) [0|3] "" XXX
- SG_ SET_ME_X01 : 7|2@0+ (1,0) [0|3] "" XXX
+ SG_ LKAS_STATUS : 7|2@0+ (1,0) [0|3] "" XXX
  SG_ SET_ME_X01_2 : 11|2@0+ (1,0) [0|3] "" XXX
```
adeebshihadeh, line 185:
```suggestion
 SG_ LKAS_STATUS : 7|2@0+ (1,0) [0|3] "" XXX
 SG_ SET_ME_X01 : 11|2@0+ (1,0) [0|3] "" XXX
```

zorrobyte: Got it, changing now

zorrobyte:
@adeebshihadeh checks failed due to the generated files not being committed, I just force pushed them the moment before it was merged FYI

adeebshihadeh:
My bad, fixed on master

sshane:
@zorrobyte fyi you can force push on the current PR or add more commits, no need to open a new one since it'll get squashed.

zorrobyte:
@sshane yep, did that and appended my merge. It just was merged into master the second before I re-ran the generator and force pushed. Thank you, Shane 😎

PR #501, opened 2022-01-19
Toyota: LKAS_HUD expansion

cydia2020:
define each known signals, this makes Toyota's DBC files easier to work with for those that aren't working on openpilot.


adeebshihadeh:
LGTM. Will merge once the conflicts are fixed.

PR #507, opened 2022-01-20
Honda: remove redundant Ridgeline DBC

vanillagorillaa:
- Remove Ridgeline. Matches the Pilot (adds the GAS_PEDAL_2 message)


adeebshihadeh:
Honda's joined the cleanup party! Let's keep these PRs targeted and small - this is the perfect size as is.

PR #508, opened 2022-01-20
Fix imports leaving extra newlines

sshane:


generator/generator.py
```diff
@@ -4,7 +4,7 @@
 
 cur_path = os.path.dirname(os.path.realpath(__file__))
 opendbc_root = os.path.join(cur_path, '../')
-include_pattern = re.compile(r'CM_ "IMPORT (.*?)";')
+include_pattern = re.compile(r'CM_ "IMPORT (.*?)";\n')
```
sshane, line 7: Still catches all cases

PR #509, opened 2022-01-20
Make def_vals use set in dbc.py

sshane:
Just saves a line (minus formatting) and ensures the defs don't have any duplicates inside dbc. Measured identical build times. I can provide tests if wanted

### DIFF
--- a/can/process_dbc.py
+++ b/can/process_dbc.py
@@ -8,6 +8,7 @@ import jinja2
 from collections import Counter
 from opendbc.can.dbc import dbc
 
+
 def process(in_fn, out_fn):
   dbc_name = os.path.split(out_fn)[-1].replace('.cc', '')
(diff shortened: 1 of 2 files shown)


can/process_dbc.py
```diff
@@ -23,8 +24,7 @@ def process(in_fn, out_fn):
   msgs = [(address, msg_name, msg_size, sorted(msg_sigs, key=lambda s: s.name not in ("COUNTER", "CHECKSUM")))
           for address, ((msg_name, msg_size), msg_sigs) in sorted(can_dbc.msgs.items()) if msg_sigs]
 
-  def_vals = {a: sorted(set(b)) for a, b in can_dbc.def_vals.items()}  # remove duplicates
-  def_vals = sorted(def_vals.items())
+  def_vals = sorted((a, sorted(b)) for a, b in can_dbc.def_vals.items())
```
sshane, line 27: Verified the same by comparing with old

PR #510, opened 2022-01-20
Cython classes cleanup

sshane:


can/parser_pyx.pyx
```diff
@@ -36,42 +36,39 @@ cdef class CANParser:
...
       raise RuntimeError(f"Can't find DBC: {dbc_name}")
     self.vl = {}
     self.ts = {}
-
+    self.can_valid = True
     self.can_invalid_cnt = CAN_INVALID_CNT
 
-    cdef int i
-    cdef int num_msgs = self.dbc[0].num_msgs
-    for i in range(num_msgs):
```
sshane, line 51: Didn't see any noticeable differences with the generated code with/without the cdef: https://www.diffchecker.com/Vmp4EWm2.

adeebshihadeh:
I don't think is worth the time to review, as I'm worried about subtle perf regressions. The cython wrappers are pretty missing, so I want to completely overhaul them soon.

PR #514, opened 2022-01-21
Toyota: sort messages by address

sshane:
Plus minor formatting


adeebshihadeh:
All the listed changes sound good, but can you split them into separate PRs?

PR #526, opened 2022-01-27
Honda: general comment cleanup

vanillagorillaa:
duplicate comments and values left in vehicle specific generator files that are already in the base file (or now added to the base).


adeebshihadeh:
@sshane is this still relevant?

PR #528, opened 2022-01-28
Honda Bosch: move standstill to common bosch DBC

sshane:
- [ ] Verify Insight works with this, a few of its signals are moved around (Insight's BRAKE_ERROR_2 is WHEELS_MOVING). Or find a common signal


sshane:
Would change behavior, best to find where the signals are common and just move them instead for now

PR #530, opened 2022-01-28
Honda: move RADAR_HUD to _bosch_2018

sshane:
The only mismatch is the `ZEROS_BOH3` signal. But the only two signals we use are the same.

Insight: `SG_ ZEROS_BOH : 7|10@0+ (1,0) [0|127] "" BDY`
Other cars: `SG_ ZEROS_BOH3 : 31|32@0+ (1,0) [0|4294967295] "" XXX`

- [ ] Verify this is okay


sshane:
Done in https://github.com/commaai/opendbc/pull/637

PR #531, opened 2022-01-28
Honda: create new DBC for common messages

sshane:
This PR pulls out common messages (duplcates) from the four different base DBCs into a holistic common DBC.

Notes:
- Only moved signals that either existed in all four base DBCs, or existed in the two bosch ones AND didn't exist elsewhere (skipped `KINEMATICS` for example)
- Added `LONG_COUNTER` signal to `ROUGH_WHEEL_SPEED`
- Removed "rpm" unit from miscellaneous `POWERTRAIN_DATA` signals
- Signals added to common DBC only used by `_bosch_20**.dbc` cars: `ACC_CONTROL`, `ACC_CONTROL_ON`, `XXX_16`


sshane:
This is getting split into smaller PRs, like: #533

sshane:
Most of this stuff is already in the base DBC, and the rest now have conflicts with 22 Civic

PR #541, opened 2022-01-30
Honda: move ACC_HUD to common DBC

sshane:
Moved quite a bit of signals around, will need to verify by looking at data.

One BOH signal to PCM_SPEED


sshane:
Done in https://github.com/commaai/opendbc/pull/632

PR #542, opened 2022-01-30
Honda: move messages to common Honda DBC

sshane:


sshane:
Has conflicts with new 22 Civic and doesn't clean up a lot any more

PR #544, opened 2022-02-01
CANParser: remove timestamp dict

sshane:
busTime was removed from the CAN data, so we need to use the logMonoTime for timestamp

### DIFF
--- a/can/common.h
+++ b/can/common.h
@@ -33,7 +33,6 @@ public:
   std::vector<Signal> parse_sigs;
   std::vector<double> vals;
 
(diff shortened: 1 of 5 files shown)


sshane:
Or should we remove ts as it's not used anywhere?

adeebshihadeh:
Let's just remove it

sshane:
Since ts is currently used in openpilot (will not be soon), I merged this into the PR to fix brake pressed mismatches for Honda: https://github.com/commaai/opendbc/pull/548/files

PR #545, opened 2022-02-01
GM dbc: Add ECMPRDNL2.PRNDL2 Message / signal with more detailed "gear" status

JasonJShuler:
The existing signal used to determine the gear shifter position does not differentiate between "D" and "L", nor is Neutral included.

This new signal has a distinct value for each shifter position. Differentiating between "D" and "L" is important especially for EVs and Hybrids where "L" mode is also known as single pedal mode. In this mode the gas pedal will perform regenerative braking.

The message - 501 - is present in all the current fingerprints.
new with L mode

Further analysis will be required to determine if this message is more appropriate than the one currently being used for all vehicles or only Hybrid/EV.


JasonJShuler:
Apologies, this needs further analysis - the values seem to be different on the Volt.

PR #548, opened 2022-02-01
CANParser: add field for all updated values

sshane:
This maintains same behavior of return value for updated message addresses and cp always holding the latest value, while adding a new field that holds all the updated values given to the parser in one cycle (either update strings or string)

To test if a message is updated you can do `len(cp.updated["POWERTRAIN_DATA"]["BRAKE_SWITCH"]) != 0`, or iterate through the values in case something lagged

Alternatively, we can store the updated values in the parser if we received two messages at once, then serve them when the next cycles are missing them, that might be better


can/parser_pyx.pyx
```diff
@@ -121,24 +125,35 @@ cdef class CANParser:
       cv_name = <unicode>cv.name
 
       self.vl[cv.address][cv_name] = cv.value
-      self.ts[cv.address][cv_name] = cv.ts
-
       self.vl[name][cv_name] = cv.value
+      self.ts[cv.address][cv_name] = cv.ts
       self.ts[name][cv_name] = cv.ts
 
+      if not init:
```
sshane, line 132: needed since on init we use this to fill all the values.

can/parser_pyx.pyx
```diff
@@ -30,6 +30,7 @@ cdef class CANParser:
     string dbc_name
     dict vl
     dict ts
+    dict up
```
adeebshihadeh, line 33: just call it updated

sshane:
Timings for packing/unpacking remain similar to before (using https://gist.github.com/sshane/93d1adab1570dbbf1c2c1fadcfd227b4):

Before PR:
```
0.5877773761749268
0.6649384498596191
0.6165750026702881
0.5889937877655029
0.5987982749938965
0.5964610576629639
0.597658634185791
0.645211935043335
0.6080794334411621
0.6069376468658447
```

After PR:
```
0.6619977951049805
0.685664176940918
0.5850427150726318
0.5612833499908447
0.5617363452911377
0.5562198162078857
0.5591819286346436
0.590254545211792
0.5700192451477051
0.5692486763000488
```

sshane:
This PR now also removes the `ts` field since it won't be used any longer, and the bus time/timestamp was always zero

can/parser_pyx.pyx
```diff
@@ -121,27 +121,26 @@ cdef class CANParser:
...
+      self.updated[cv.address][cv_name] = cv.updated_values
+      self.updated[name][cv_name] = cv.updated_values
+
+      if cv.updated_values.size():
+        updated_val.insert(cv.address)
 
     return updated_val
 
-  def update_string(self, dat, sendcan=False):
+  def update_string(self, dat, sendcan=False, update_vl=True):
```
sshane, line 134: Since update_vl returns all values and sets updated vals, we don't want to reset the vals each string if updating multiple

adeebshihadeh:
Seems like nothing uses `update_string`, let's just remove it.

can/common_dbc.h
```diff
@@ -24,9 +24,9 @@ struct MessageParseOptions {
 
 struct SignalValue {
   uint32_t address;
-  uint16_t ts;
   const char* name;
   double value;
+  std::vector<double> updated_values;
```
adeebshihadeh, line 29:
```suggestion
  double value;                                   // latest value
  std::vector<double> latest_values;  // all values the last set of messages
```

sshane:
Reverted removal of ts field, will change to monoLogTime in another PR

can/common.h
```diff
@@ -31,7 +31,8 @@ class MessageState {
   unsigned int size;
 
   std::vector<Signal> parse_sigs;
-  std::vector<double> vals;
+  std::vector<double> last_vals;
```
sshane, line 34: this is equivalent to `SignalValue`, so probably should keep at `vals`

PR #551, opened 2022-02-04
Toyota: add alternate brake pressed bit

sshane:
This bit may be common across all 3 platforms, I checked a Corolla (new_mc) and a HighlanderH (tnga_k), but not nodsu yet. For both routes, their respective `BRAKE_MODULE` brake pressed signal matched up perfectly.

Draft PR until we take a better look at this.

openpilot issue: https://github.com/commaai/openpilot/issues/23706


cydia2020:
I can confirm that this signal exists on my Prius.

sshane:
Checked this signal a while ago and recall it didn't catch every brake pressed case on all Toyotas, will look into it again and close if this signal doesn't work for all cars.

sshane:
Adding another bit in https://github.com/commaai/opendbc/pull/794

PR #553, opened 2022-02-06
remove round that does nothing

GregorKikelj:


adeebshihadeh:
Since round doesn't always return an int, I think making this explicit is more clear.

GregorKikelj:
Looking at https://docs.python.org/3/library/functions.html#round seems like the single argument round always returns an int

PR #554, opened 2022-02-07
Honda: unify CRV Hybrid, Civic Bosch Diesel, Accord/H DBCs

sshane:
We only add new messages by merging these DBCs, and their `GEAR_SHIFTER` definitions are similar to the other cars we merged into the ILX in this PR: https://github.com/commaai/opendbc/pull/539/files

Instead of keeping the Accord GEAR definition here, changed to the ILX gear definition (made sure to verify Accord matches its old parsed gear values as well).  Also gain neutral gear with this change, the previous definition wasn't correct for that.

Need to verify the new definitions match the old ones for:
- [x] Honda CRV Hybrid
- [x] Honda Civic Sedan Diesel (Bosch)
- [x] Honda Accord 2018
- [x] Honda Accord Hybrid 2018

Plots: https://imgur.com/a/VEk6amO

Wanted to make sure that the whole segments where the car is in neutral with the new CAN definitions were real, the first one in the list above is the user in a car wash! So seems to check out: https://imgur.com/a/kVF4H3Y


sshane:
In 2 segments out of 200,000, the Accord Hybrid has two dongles where the new definition I'm using from the Acura ILX becomes inaccurate, and it switches between D and 0 rapidly, which is an undefined gear. While the old gear parsing has the car at D the whole time. The car was on the highway at this time, so this new definition cannot be used. Weirdly, in another segment in the route, the new parsing matches the old exactly.

However using the definition from the Civic Deisel still nets us the neutral gear, while working for all other cars. As it has the least segments at 2, we don't need to worry about definitions changing for it if we use that one now.

sshane:
How mismatches are defined for each segment: `mismatches = sum([int(old != new and not (new == 'N' and old == 'U')) for old, new in zip(old_gears, unified_gears)])`

Total is total data in the database, checked is data available to query.

```
Platform:   HONDA CIVIC SEDAN 1.6 DIESEL 2019
Total:      1 unique dongles, 5838 unique routes
Checked:    1 unique dongles, 19 unique routes, and 123 segments
Mismatches: 0 frames total, 0 segments (any frame)
Segments with mismatches: set()
---
Platform:   HONDA ACCORD 2018
Total:      212 unique dongles, 1622430 unique routes
Checked:    18 unique dongles, 1366 unique routes, and 11377 segments
Mismatches: 0 frames total, 0 segments (any frame)
Segments with mismatches: set()
---
Platform:   HONDA CR-V HYBRID 2019
Total:      21 unique dongles, 111421 unique routes
Checked:    1 unique dongles, 151 unique routes, and 1227 segments
Mismatches: 0 frames total, 0 segments (any frame)
Segments with mismatches: set()
---
Platform:   HONDA ACCORD HYBRID 2018
Total:      93 unique dongles, 740371 unique routes
Checked:    11 unique dongles, 1200 unique routes, and 13915 segments
Mismatches: 0 frames total, 0 segments (any frame)
Segments with mismatches: set()
```

PR #557, opened 2022-02-09
CANParser: remove ts field

sshane:
Was set with 0 before, and setting with logMonoTime doesn't give us any more information as it's guaranteed that we receive messages in the same order here: https://gist.github.com/sshane/93d1adab1570dbbf1c2c1fadcfd227b4


sshane:
Speeds up total time in the CANParser in a certain test route from `0.804498` seconds to `0.605563` seconds. Will be less when #556 is merged

PR #568, opened 2022-02-22
CAN FD support in packer + parser

adeebshihadeh:


Matheus-Garbelini:
Hi @adeebshihadeh thanks for the can-ds support addition.
Do you happen to have sample capture for can-fd so we can test this PR?
I can see many types of CRC checks for several cars.
Regards.

PR #569, opened 2022-02-23
CANParser: expose last seen timestamp of messages

sshane:


adeebshihadeh:
still needed?

sshane:
probably not, until we need it for giving more info about the can errors using test_models

PR #580, opened 2022-03-21
Hongqi: Add DBC for Hongqi HS5

jyoung8607:
Work in progress, we still need to locate several more signals.


jyoung8607:
Merging the WIP DBC alone. Its current state is reasonably functional.

There is a known issue with properly tracking the car's switchable basic/adaptive cruise control modes, and some ancillary signals still need to be tracked down. See the list in commaai/openpilot#24012 for more detail.

The CHECKSUM algorithm is a simple XOR, already supported by opendbc for other cars. I've removed the CHECKSUM and COUNTER support from this PR to avoid adding code paths without usage/test coverage. They can be re-added at such time work on the HS5 port is resumed.

PR #583, opened 2022-04-04
body: counter and crc

briskspirit:
add to sanity check and parser


can/process_dbc.py
```diff
@@ -96,6 +96,12 @@ def process(in_fn, out_fn):
           sys.exit("%s: PEDAL COUNTER is not 4 bits long" % dbc_msg_name)
         if sig.name == "CHECKSUM_PEDAL" and sig.size != 8:
           sys.exit("%s: PEDAL CHECKSUM is not 8 bits long" % dbc_msg_name)
+      # body rules
+      if address == 0x250:
+        if sig.name == "COUNTER" and sig.size != 4:
+          sys.exit("%s: BODY COUNTER is not 4 bits long" % dbc_msg_name)
+        if sig.name == "CHECKSUM" and sig.size != 8:
+          sys.exit("%s: BODY CHECKSUM is not 8 bits long" % dbc_msg_name)
```
adeebshihadeh, line 104: this isn't needed, the pedal only had an exception because the signals have a "_PEDAL" suffix

PR #584, opened 2022-04-05
GM AcceleratorPedal2 adds GasPressed, GasPressedInACC, Speed

qadmus:
Not backtested beyond cabana visual check of test route. ~For @sshane cooking ndog.~


sshane:
This is unrelated to no disengage on gas, this is just to use a gas pressed bit where available. Ran it against the gasPressed signal in openpilot and it was near identical

qadmus:
Fixing the generated files

qadmus:
Replaced by #591

PR #585, opened 2022-04-06
Fix Honda checksum for extended CAN IDs. Add test function.

csouers:
Fixes a regression from https://github.com/commaai/opendbc/pull/568. `address` gets decimated so we can't use it for the extended id check. This also adds a test that checks both Honda checksum styles against a set of known good values.

Verified by Discord user: RedAcid
Vehicle: 2020 Honda Civic Hatchback


can/tests/test_packer_parser.py
```diff
@@ -182,6 +182,43 @@ def test_updated(self):
       if len(user_brake_vals):
         self.assertEqual(vl_all[-1], parser.vl["VSA_STATUS"]["USER_BRAKE"])
 
+  def test_honda_checksum(self):
```
adeebshihadeh, line 185: can you pull this into a separate file called test_checksums?

jyoung8607: +1, been meaning to do this for VW. Will do so once a place/structure is settled on.

csouers: Done. Comments?

adeebshihadeh:
Thanks for the fix, and especially for writing tests :)

csouers:
> Thanks for the fix, and especially for writing tests :)

Any time!

PR #590, opened 2022-04-13
Ford: add OP-style transmission values

incognitojam:
DBC updates to allow for parsing transmission gear in openpilot

Would it be preferred that I create a new DBC file with the OP relevant signals, so that I don't have to delete any of the original information?


incognitojam:
Also this DBC has duplicates for lots of signals with both `VAL_` and `VAL_TABLE_` statements... I'm not sure if that affects how OP parses it, but other brands just seem to use `VAL_`.

adeebshihadeh:
Hmm, this DBC is actually quite nice. I'm cool with leaving the Ford as is if you want.

incognitojam:
If you think so, that's fine with me :smile:

PR #596, opened 2022-04-23
Added new DBC file: mazda_radar.dbc

MoreTore:
This contains Mazda radar tracks as well as cruise state data from the radar. These messages seem to be common to all Mazdas with MRCC.


MoreTore:
Please add this to Plotjuggler release.

gregjhogan:
This seems super sloppy, signals that I assume should be identical for each radar track address have different ordering, value ranges, and even the name is inconsistent sometimes.

for example:
```
SG_ RVEL_OBJ : 31|11@0- (1,0) [0|255] "" XXX
SG_ RVEL_OBJ : 31|11@0- (1,0) [0|63] "" XXX
SG_ RELV_OBJ : 31|11@0- (1,0) [0|63] "" XXX
```

For Hyundai I wrote a script to stamp out all the identical messages to keep things consistent and clean:
https://github.com/commaai/opendbc/blob/master/hyundai_kia_mando_front_radar_gen.py

Also, did you check if these object messages are sent from the radar and not the camera?  It is strange there are so few, so it seems possible they are camera objects or just objects fused with camera objects.

MoreTore:
You are correct. I did catch that after it was already merged. I'm quite dyslexic. I have corrected it on my fork and i am debugging parser errors before I make another PR.

 ```
return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, 2)
  File "opendbc/can/parser_pyx.pyx", line 67, in opendbc.can.parser_pyx.CANParser.__init__
IndexError: tuple index out of range

```

gregjhogan:
If you haven't already verified these messages come from the radar, I would also try covering up the stock camera and make sure you still get data in these messages.   If these objects are dependent on the stock camera, you probably want to ignore them completely and use openpilot longitudinal in vision only mode (because the openpilot vision objects should be much better than the stock camera long-term).

MoreTore:
These are from the radar. I have a wiring harness that plugs directly into the radar to isolate it from the rest of the car. I have tied the isolated radar bus to the camera bus so that I can intercept the radar messages and paste in the OP acceleration command the same way we intercept the lkas command. Since the radar is on the same bus as the camera, I have the signals/checks for the radar as part of the cam_can_parser. This way, I can pass through the messages to mazdacan as cp_cam.vl. I think this might be an issue now because those signals are also part of the radar_can_parser. mazda_radar.dbc and mazda_2017.dbc have the same addresses with different signals within them. It may be something else but I will have to go over the code to make sure.

gregjhogan:
If the objects in these messages are always fused with (and therefore depend on) the vision objects from the stock camera, they could still come out of the radar and what I was saying still applies.  Hopefully they are pure radar objects and the radar doesn't wait for a matching object from the camera, but it is really weird that there are so few and covering the stock camera may be an easy way to determine this.

MoreTore:
Ok I see what you mean. I will test that out.

MoreTore:
I will say the points coming from the radar are a lot more stable than the OP lead estimate. I have seen stable tracks from up to 150 meters away so that is a lot better than the C2 lead estimate. Another thing I saw is a stable radar track for an occluded vehicle in front of the lead. 

You can see from this image that it correctly tracked 3 lead vehicles at once.
Screenshot from 2022-04-26 21-46-44

gregjhogan:
That is pretty cool, but openpilot for quiet a while now identifies all leads using vision first (except at very low speed) and then fuses with radar points second (only to get relative speed/distance).  This means only the immediate lead matters (a car in front of your lead that you cannot see is not used) and the most important thing is you get a radar point for whatever openpilot vision thinks is the lead (which is very reliable with many radar points).  Situations like coming up to a stopped vehicle on a corner with multiple other vehicles around may have problems with so few points (especially if they you are only getting radar objects fused with stock camera objects).

MoreTore:
I would imagine the stock camera is doing something similar, that is, it gets data from the radar and selects a lead based on the lane information. 

Even with the radar points I am having issues with keeping a safe lead distance so I think I need to figure out how to setup live tuning when I get a chance.

PR #597, opened 2022-04-24
build on ubuntu 22.04

MoreTore:


adeebshihadeh:
openpilot and its submodules target Ubuntu 20.04. If the dependency versions need to be bumped, that should be done in openpilot.

PR #598, opened 2022-04-26
2022 Honda Civic DBC

vanillagorillaa:
Redoing after DBC clean up project. Had to move the 829 LKAS_HUD signal out of the Honda_common dbc as the 2022 Civic uses an 8 byte 829 LKAS_HUD signal. Also, 479 ACC_CONTROL is now on 477 on 2022 Civic. We can make it '477 ACC_CONTROL_2' if we want to not remove it from the _bosch_2018 dbc.


sshane:
cleaned up and moved to https://github.com/commaai/opendbc/pull/618

PR #602, opened 2022-05-06
Dynamic DBC parsing

sshane:
Moved and minimalized from deanlee's PR here: https://github.com/commaai/opendbc/pull/376

Before, the parsing speed test took ~0.002 seconds for 25 DBCs, now it takes ~0.09, which is still pretty quick if we're worried about startup time (only ~3.6 ms per DBC).

`test_can_parsing_speed` is also the exact same as master (0.6 to 0.7 seconds)

- [x] add `get_dbcs` back for PJ


can/SConscript
```diff
@@ -3,18 +3,13 @@ Import('env', 'envCython', 'cereal')
...
-    env.Depends(dbc, ["dbc.py", "process_dbc.py"])
-    dbcs.append(dbc)
-
-libdbc = env.SharedLibrary('libdbc', ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]+dbcs, LIBS=["capnp", "kj"])
+envDBC = env.Clone()
+define_path = '-DDBC_FILE_PATH=\'"%s"\'' % (envDBC.Dir("..").abspath)
+if isinstance(envDBC['CXXFLAGS'], list):
+  envDBC['CXXFLAGS'] += [define_path]
+else:
+  envDBC['CXXFLAGS'] += ' ' + define_path
```
sshane, line 11: Should we set an environment variable from PJ instead of this, and then default to $HOME/openpilot like we do for cereal?

adeebshihadeh:
Let's add some tests around this, including DBC parsing time.

can/SConscript
```diff
@@ -3,18 +3,13 @@ Import('env', 'envCython', 'cereal')
...
-    dbcs.append(dbc)
-
-libdbc = env.SharedLibrary('libdbc', ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]+dbcs, LIBS=["capnp", "kj"])
+envDBC = env.Clone()
+dbc_file_path = '-DDBC_FILE_PATH=\'"%s"\'' % (envDBC.Dir("..").abspath)
+if isinstance(envDBC['CXXFLAGS'], list):
+  envDBC['CXXFLAGS'] += [dbc_file_path]
+else:
+  envDBC['CXXFLAGS'] += ' ' + dbc_file_path
+libdbc = envDBC.SharedLibrary('libdbc', ["dbc.cc", "parser.cc", "packer.cc", "common.cc"], LIBS=["capnp", "kj"])
```
sshane, line 12: Think this is the cleanest, then you override the BASEDIR envvar if you want to get the DBCs from somewhere else (like PJ would use). If we use RPATH we'd have to get the path of the shared library in the CANParser and CANPacker Cython libs, then pass as an argument to libdbc

can/tests/test_parser_speed.py
```diff
@@ -0,0 +1,50 @@
...
+      cls.dbcs.append(os.path.basename(dbc).split('.')[0])
+
+  def test_dbc_parsing(self):
+    start_time = time.time()
+    for dbc in self.dbcs:
+      CANParser(dbc, [], [], 0)
+    elapsed = time.time() - start_time
+    self.assertLess(elapsed, 0.15, "Took too long to parse {} DBCs".format(len(self.dbcs)))
+
+  def test_can_parsing(self):
```
sshane, line 29: Same result as master on this test, ~0.6 to 0.7 seconds

can/tests/test_parser_speed.py
```diff
@@ -0,0 +1,50 @@
...
+import unittest
+
+from opendbc import DBC_PATH
+from opendbc.can.packer import CANPacker
+from opendbc.can.parser import CANParser
+from opendbc.can.tests.test_packer_parser import can_list_to_can_capnp
+from opendbc.generator.generator import generated_suffix
+
+
+class TestCANParserSpeed(unittest.TestCase):
```
adeebshihadeh, line 15: just call this TestCANParser. speed is just one test case. packer + parser combo tests can remain separate

can/dbc.cc
```diff
@@ -1,27 +1,251 @@
...
+}
+
+const DBC* dbc_lookup(const std::string& dbc_name) {
+  static std::mutex lock;
+  static std::map<std::string, DBC*> dbcs;
+
+  std::unique_lock lk(lock);
+  set_dbc_file_path();
+
+  auto it = dbcs.find(dbc_name);
```
sshane, line 232: Since parsing DBCs is so quick, do we need this? Not sure how static `dbcs` works, if it caches anything or not

sshane:
On a C3 with the test on this PR:

```
master:
test_dbc_parsing_speed: 2.7299, 2.7812, 2.7356, 2.7723, 2.8465, 2.8937, 2.7382 ms for 25 DBCs. 0.1114 ms per DBC
test_can_parsing_speed: 1255.7981, 1278.3473, 1274.1129, 1250.8554, 1264.3416, 1273.9437, 1253.8807 ms

dynamic-parsing:
test_dbc_parsing_speed: 165.3652, 165.7858, 165.437, 165.437, 165.5698, 165.0653, 165.3037 ms for 25 DBCs. 6.6169 ms per DBC
test_can_parsing_speed: 1255.7981, 1278.3473, 1274.1129, 1250.8554, 1264.3416, 1273.9437, 1253.8807 ms
```

can/tests/test_can_parser.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestCANParser(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+    cls.dbcs = []
+    for dbc in glob.glob(f"{DBC_PATH}/*{generated_suffix}"):
+      cls.dbcs.append(os.path.basename(dbc).split('.')[0])
+
+  def test_detecting_dbcs(self):
+    self.assertGreaterEqual(len(self.dbcs), 20)
```
adeebshihadeh, line 22: this isn't a great test. either check that all the real DBCs are detected or create a tmp directory with a random number of fake DBCs

can/tests/test_can_parser.py
```diff
@@ -0,0 +1,54 @@
...
+
+  def test_dbc_parsing_speed(self):
+    start_time = time.time()
+    for dbc in self.dbcs:
+      CANParser(dbc, [], [], 0)
+
+    elapsed = time.time() - start_time
```
adeebshihadeh, line 29: time.time can jump
```suggestion
    start_time = time.monotonic()
    for dbc in self.dbcs:
      CANParser(dbc, [], [], 0)

    elapsed = time.monotonic() - start_time
```

can/tests/test_can_parser.py
```diff
@@ -0,0 +1,54 @@
...
+    self.assertLess(elapsed, 0.15, "Took too long to parse {} DBCs".format(len(self.dbcs)))
+
+  def test_can_parsing_speed(self):
```
adeebshihadeh, line 32: this doesn't really go in this PR. I'd suggest breaking out this test class into another PR merged before this one and only adding relevant test cases in this PR

sshane: Wanted to test changing from arrays to vectors, but agreed. No change in time here

sshane:
For the `hyundai_kia_generic` DBC, times to parse are:

`BO_ lines`:
- regex parsing: 0.12 ms
- all processing: 0.381620 ms

`SG_ lines`:
- regex parsing: 4.266413 ms
- all processing: 4.700723 ms

`VAL_ lines`:
- regex parsing: 0.031581 ms
- all processing: 0.650602 ms

sshane:
Using [`ctre`](https://github.com/hanickadot/compile-time-regular-expressions), for the `hyundai_kia_generic` DBC, most of the time is spent in the `SG_` section of the loop, taking a total time of 5-6ms just to run the pattern matching (`auto match = ctre::match<sg_regexp_ctre>(line);`)

```python
bo: 0.416959 ms
sg: 5.880782 ms
val: 0.105246 ms

bo: 0.302340 ms
sg: 6.181302 ms
val: 0.074481 ms

loop time: 7.103972 ms
bo: 0.245470 ms
sg: 5.428277 ms
val: 0.049240 ms

loop time: 7.064819 ms
bo: 0.249569 ms
sg: 5.415454 ms
val: 0.073308 ms

loop time: 7.939813 ms
bo: 0.266407 ms
sg: 6.127016 ms
val: 0.083280 ms

loop time: 7.658933 ms
bo: 0.249199 ms
sg: 5.910799 ms
val: 0.049321 ms

loop time: 7.535592 ms
bo: 0.251016 ms
sg: 5.785024 ms
val: 0.089635 ms
```

sshane:
With compiling the regex at import time and an improvement when there's many `VAL_`'s, HKG is about the same as ctre:

```python
loop time: 5.812946 ms
bo: 0.304730 ms
sg: 5.007673 ms
val:  0.157733 ms

loop time: 6.428721 ms
bo: 0.346666 ms
sg: 5.514892 ms
val:  0.139059 ms

loop time: 6.363871 ms
bo: 0.332463 ms
sg: 5.471228 ms
val:  0.170879 ms

loop time: 5.180957 ms
bo: 0.298319 ms
sg: 4.432689 ms
val:  0.136576 ms
```

Toyota is also about the same:
```python
loop time: 1.789508 ms
bo: 0.146846 ms
sg: 0.851223 ms
val:  0.632434 ms
```

can/dbc.cc
```diff
@@ -185,8 +185,7 @@ DBC* dbc_parse(const std::string& dbc_name) {
       val.name = match[2].str();
 
       auto defvals = match[3].str();
-      std::regex regex{R"([\"]+)"};  // split on "
```
sshane: now toyota is much quicker with this gone

can/tests/test_can_parser.py
```diff
@@ -0,0 +1,35 @@
...
+    for dbc, lines in self.dbcs.items():
+      start_time = time.monotonic()
+      CANParser(dbc, [], [], 0)
+      dbc_times.append((dbc, lines, time.monotonic() - start_time))
+
+    # for dbc, lines, t in sorted(dbc_times, key=lambda x: x[-1]):
+    #   print('{:54} ({:5} lines): {} ms'.format(dbc, lines, round(t * 1000, 4)))
+
+    max_ms_per_line = max([t / lines * 1000 for _, lines, t in dbc_times])
+    self.assertTrue(max_ms_per_line < 0.02, f'Max ms/line time too high: {max_ms_per_line} >= 0.02')
```
adeebshihadeh, line 31: move the assert into the loop body and add the DBC name to the error msg

can/tests/test_can_parser.py
```diff
@@ -0,0 +1,35 @@
...
+class TestCANParser(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+    cls.dbcs = {}
+    for dbc in glob.glob(f"{DBC_PATH}/*.dbc"):
+      dbc_name = os.path.basename(dbc).split('.')[0]
+      with open(dbc, 'r') as f:
+        cls.dbcs[dbc_name] = len(f.readlines())
+
+  def test_dbc_parsing_speed(self):
```
adeebshihadeh, line 20: let's also add a check for the overall time that's slightly more than the current max

can/SConscript
```diff
@@ -1,20 +1,14 @@
...
-    env.Depends(dbc, ["dbc.py", "process_dbc.py"])
-    dbcs.append(dbc)
-
-libdbc = env.SharedLibrary('libdbc', ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]+dbcs, LIBS=[common, "capnp", "kj", "zmq"])
+envDBC = env.Clone()
+dbc_file_path = '-DDBC_FILE_PATH=\'"%s"\'' % (envDBC.Dir("..").abspath)
+if isinstance(envDBC['CXXFLAGS'], list):
+  envDBC['CXXFLAGS'] += [dbc_file_path]
+else:
+  envDBC['CXXFLAGS'] += ' ' + dbc_file_path
```
adeebshihadeh, line 10: why isn't this always a list? we assume it is in other SConscripts

can/dbc.cc
```diff
@@ -1,27 +1,251 @@
...
+#include <sstream>
+#include <unistd.h>
 #include <vector>
 
 #include "common_dbc.h"
 
-std::vector<const DBC*>& get_dbcs() {
-  static std::vector<const DBC*> vec;
-  return vec;
+std::string dbc_file_path;
```
adeebshihadeh, line 15: why is this a global?

can/parser_pyx.pyx
```diff
@@ -49,8 +49,7 @@ cdef class CANParser:
     self.can_invalid_cnt = CAN_INVALID_CNT
 
     cdef int i
-    cdef int num_msgs = self.dbc[0].num_msgs
-    for i in range(num_msgs):
+    for i in range(self.dbc[0].msgs.size()):
```
adeebshihadeh, line 52: usually the cdef is to be explicit about being a C loop, did you check if this is this still a C loop?

can/tests/test_can_parser.py
sshane: Sometimes on my PC the ford DBC can take 80 ms, but usually it's 30 ms

can/dbc.cc
sshane: doesn't need to be, thought `set_dbc_file_path` might've been slow but it takes 0.001396 ms

can/SConscript
sshane: ah right, it is. I could have sworn I printed it and it was a string on my PC but list in CI

can/parser_pyx.pyx
sshane: main difference I see is `__Pyx_decode_cpp_string` with this PR vs `__Pyx_decode_c_string` with master for the decode line

sshane: Definitely gets slower without the `cdef i`. avg `0.10887859389185905` ms vs `0.16389694064855576` ms

sshane: Ah there's no more strlen called on the string when decoding, no matter if cdef is there or not. Think it's not a c loop anymore

PR #606, opened 2022-05-10
Toyota: 6MT (manual transmission) signals

bscholer:
Adds `CLUTCH_RELEASED` signal for the clutch pedal on Toyotas with 6MT. Message was previously `untitled`, so this does not change any existing signals.

This is tested on a 2022 Corolla Hatchback 6MT.

I was **not able to identify**:
-  Signal showing the _amount_ the clutch is pressed, so this change is a `boolean` value.
-  Signal providing the gear that the transmission is in (1-6), besides reverse, which is already provided by `GEAR_PACKET.GEAR`.
    -  It is worth noting that `GEAR_PACKET.GEAR` on a 6MT _only_ changes when in reverse, and none of the other signals within `GEAR_PACKET` seem to ever change, which makes sense. I've added a comment to that signal.
  
image


toyota_nodsu_pt_generated.dbc
```diff
@@ -403,14 +389,53 @@ BO_ 1592 DOOR_LOCKS: 8 XXX
  SG_ LOCK_STATUS : 20|1@0+ (1,0) [0|1] "" XXX
  SG_ LOCKED_VIA_KEYFOB : 23|1@0+ (1,0) [0|1] "" XXX
 
+BO_ 865 CLUTCH: 8 XXX
+ SG_ CLUTCH_PRESSED : 38|1@0+ (1,0) [0|1] "" XXX
```
cydia2020, line 393: add this to `opendbc\generator\toyota\_toyota_2017.dbc`, don't change the rest of the file, and run `generator.py` with python

cydia2020 (changes requested):

PR #610, opened 2022-05-13
CANParser: parse all signals of a message

sshane:
This might be pretty slow actually.

~80 ms vs 184 ms to parse 10,000 `ACC_CONTROL` messages from Toyota when going from just one message to all of them.

most of it might be in cython though!


can/parser.cc
```diff
@@ -292,6 +275,48 @@ void CANParser::UpdateValid(uint64_t sec) {
...
 }
 
+double CANParser::get_value(uint32_t msg_addr, std::string &signal) {
+  auto state_it = message_states.find(msg_addr);
+  if (state_it == message_states.end()) {
+    fprintf(stderr, "CANParser: could not find message 0x%X in DBC\n", msg_addr);
+    assert(false);
+  }
+
+  MessageState state = state_it->second;
```
deanlee, line 285: this will make a deep copy. should be `const MessageState &state = state_it->second;`?

sshane:
Taken over in https://github.com/commaai/opendbc/pull/795

sshane:
Delete me when new PR is open!

sshane:
done in https://github.com/commaai/opendbc/pull/828

PR #616, opened 2022-05-14
Ignore checksum and counter from body

sshane:
TODO: fill these out

https://github.com/commaai/body/issues/2


sshane:
Added new test route instead

PR #618, opened 2022-05-14
2022 Honda Civic DBC

sshane:
- Created `_honda_common_2018.dbc` that houses older `LKAS_HUD` and `ACC_CONTROL` @vanillagorillaa do we have a better name, or is this good?
  - No conflicts or major changes to other files
- Files that imported `_honda_common.dbc` now also import `_honda_common_2018.dbc` for `ACC_CONTROL` and `LKAS_HUD`, just only Bosch uses `ACC_CONTROL`


sshane:
Bit simpler

PR #619, opened 2022-05-16
CANParser: display name for missing messages

sshane:
...instead of address 0x0, helps with car ports.

parser.cc handles the missing messages if address is provided instead of name. Ideally parser.cc handles all of this, but passing in message name required a little bigger of a change in pyx than I'd like, and we should refactor it all at once if we do that.


sshane:
This should be done in parser.cc when we move more of the Cython code into it

PR #620, opened 2022-05-17
2022 Honda Civic DBC

sshane:
- `LKAS_HUD` is the same address as the one in the root common DBC, however the size is different with signals now taking up the new 3 bytes. Move LKAS_HUD from `_honda_common` to `_bosch_2018_adas` and `_nidec_common`.
- (Suspected) `ACC_CONTROL` message is different address, but so we don't have to add a suffix of `_2` to the new one and include the old ones in the new DBC needlessly, make a new DBC called `_bosch_adas_2018` for the old ACC_CONTROL message.

Diff before generating is +154 -53


sshane:
ACC_CONTROL (477) looks wrong for this car, the signals in it don't really look like you expect them too (wrong definitions, so the accel cmd is going from pos to neg, CONTROL_ON not mimicking POWERTRAIN's signal, etc). However from @vanillagorillaa's list of messages the camera sends, think I found the one the camera actually sends for longitudinal!

Old ACC_CONTROL (sent by car):
Screenshot from 2022-05-16 21-01-17

New message (sent by camera): 
Screenshot from 2022-05-16 20-57-43

PR #621, opened 2022-05-17
Fix DBC parsing time test

sshane:


can/tests/test_can_parser.py
```diff
@@ -18,14 +19,19 @@ def setUpClass(cls):
...
+        CANParser(dbc, [], [], 0)
+        total_time_ms = (time.monotonic() - start_time) * 1000
+        dbc_parsing_times[dbc]['total'].append(total_time_ms)
+        dbc_parsing_times[dbc]['lines'].append(total_time_ms / lines)
 
-      self.assertTrue(total_time_ms < 45, f'{dbc}: total parsing time too high: {total_time_ms} ms >= 45 ms')
-      ms_per_line = total_time_ms / lines
-      self.assertTrue(ms_per_line < 0.02, f'{dbc}: max ms/line time too high: {ms_per_line} ms >= 0.02 ms')
+    for dbc, dbc_times in dbc_parsing_times.items():
+      # FIXME: this doesn't really test anything since it caches the parsed dbc after the first run
```
sshane, line 32: Tried to del CANParser, pop it from sys.modules, reload it using importlib, and only import it using importlib but it's keeping its cache around

PR #622, opened 2022-05-21
Add Toyota LTA state signal

echo1001:
Steering fault feedback for TSS2 LTA steering mode.


sshane:
Want to add some definitions for what you found the values mean in the DBC?

echo1001:
I was trying to pinpoint exactly what they meant, but I've added what I found so far. This should be good enough to add in fault messages.

sshane:
One more thing -- can you share a route where it faults with various codes?

echo1001:
With faults - `7e8e854eb2c6127c|2022-06-01--10-42-36--11`
Just to clarify, did you mean LTA staying active? If so: `7e8e854eb2c6127c|2022-06-01--10-42-36--5`
The LKA state is normally at 1 regardless of LTA's state.

generator/toyota/toyota_nodsu_pt.dbc
```diff
@@ -52,3 +53,4 @@ CM_ SG_ 1014 APPROACHING_ENABLED "when BSM is enabled in settings, this is on al
 
 VAL_ 610 IPAS_STATE 5 "override" 3 "enabled" 1 "disabled";
 VAL_ 610 LKA_STATE 25 "temporary_fault" 17 "permanent_fault" 9 "temporary_fault2" 5 "active" 1 "standby";
+VAL_ 610 LTA_STATE 6 "temporary_fault" 4 "temporary_fault2" 2 "temporary_fault3" 1 "active" 0 "disabled";
```
sshane, line 56: you can duplicate the definition for LKA state above

echo1001:
Thanks Shane!

PR #626, opened 2022-06-01
tests: remove DBC parsing timing test

sshane:
As discusses, we shouldn't have timing tests in CI as it's too flaky. Can possibly add back when Cython code is moved to C++ and caching is removed so we can test same DBC multiple times


can/tests/test_dbc_parser.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class TestDBCParser(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+    cls.dbcs = []
+    for dbc in glob.glob(f"{DBC_PATH}/*.dbc"):
+      cls.dbcs.append(os.path.basename(dbc).split('.')[0])
+
+  def test_parse_all_dbcs(self):
```
sshane, line 17: Should add tests for catching common issues like duplicate addresses and signal out of bounds errors here too

PR #627, opened 2022-06-05
Toyota: Follow Distance Signal

cydia2020:
Used by the stock PCM adaptive cruise control to determine how close the lead should be followed. This changes with each press of the distance button, and cycles from 1 to 2 to 3.


cydia2020:
This is different to `DISTANCE_LINES` in `0x399`, this signal always reflects the follow distance no matter if the distance setting display pops-up or not. `DISTANCE_LINES` is not 0 when the distance button is pressed.

sshane:
Nice, so I'm guessing this is available on all TSS cars?

cydia2020:
Still need to verify this, but it should be available on all Toyotas.

sshane:
Thanks for the PR! Looks good against a TSSP Corolla and TSS2 Camry Hybrid

PR #629, opened 2022-06-07
Change GM PRNDL2 transition value (0) to U

JasonJShuler:
The new ECMPRDNL2.PRNDL2 has a value of zero while the transmission is shifting between modes. This maps most appropriately to the "unknown" GearShifter value in car.capnp.

Changed 0 from "Shifting" to "U"

Part of implementation for issue https://github.com/commaai/openpilot/issues/24763


JasonJShuler:
Realized this is not necessary as the mapping function defaults to unknown.

PR #633, opened 2022-06-09
VW PQ: Message updates, checksum and counter support

jyoung8607:
Slight refactor to MQB CRC/counter handling to make way for PQ, and then handle PQ checksums and counters. Holding in draft until I can verify against a couple different PQ vehicles.


jyoung8607:
Checks out against a Golf Mk6, ready for consideration.

vw_golf_mk4.dbc
```diff
@@ -1084,7 +1084,7 @@ BO_ 872 ACC_System: 8 XXX
  SG_ ACS_max_AendGrad : 48|8@1+ (1,0.02) [0.02|5.06] "Unit_MeterPerSeconSquar" XXX
 
 BO_ 1386 ACC_GRA_Anziege: 8 XXX
- SG_ ACA_Checksum : 0|8@1+ (1,0) [0|255] "" XXX
```
adeebshihadeh, line 1087: think it would be nice to keep these names. how do you feel about using the unit field for this instead?

can/common.cc
```diff
@@ -188,6 +188,22 @@ unsigned int volkswagen_crc(uint32_t address, const std::vector<uint8_t> &d) {
...
+  uint8_t checksum = 0;
+  int checksum_byte = checksum_start_bit / 8;
+
+  // Simple XOR over the payload, except for the byte where the checksum lives.
+  for (int i = 0; i < d.size(); i++) {
+    if (i == checksum_byte) {
+      continue;
+    } else {
+      checksum ^= d[i];
+    }
```
adeebshihadeh, line 201:
```suggestion
    if (i != checksum_byte) {
      checksum ^= d[i];
    }
```

can/common.h
```diff
@@ -26,7 +26,8 @@ unsigned int honda_checksum(uint32_t address, const std::vector<uint8_t> &d);
 unsigned int toyota_checksum(uint32_t address, const std::vector<uint8_t> &d);
 unsigned int subaru_checksum(uint32_t address, const std::vector<uint8_t> &d);
 unsigned int chrysler_checksum(uint32_t address, const std::vector<uint8_t> &d);
-unsigned int volkswagen_crc(uint32_t address, const std::vector<uint8_t> &d);
+unsigned int volkswagen_mqb_crc(uint32_t address, const std::vector<uint8_t> &d);
+unsigned int volkswagen_pq_checksum(int checksum_start_bit, const std::vector<uint8_t> &d);
```
adeebshihadeh, line 30: I'd like to keep the signatures for all the checksum functions the same so we can eventually do something nicer than the if-else-spaghetti. think this would be best:
```suggestion
unsigned int volkswagen_pq_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d);
```

can/common.cc
jyoung8607: LOL, derp... stared at that when I wrote it for a couple minutes because I felt like there had to be a more concise way, namely, this. Don't know where my mind was. Thanks!

can/common.h
jyoung8607: Reasonable. It's a lot of touches, so early next week I'll propose a null-effect refactor of the existing code to work as per your suggestion, then once that passes review, rebase this on top.

vw_golf_mk4.dbc
jyoung8607: This has come up in discussion before; my first cut at MQB back in the day kept the signal names intact. I'm amenable to any good solution that lets us tag an existing canonically-named signal as "the checksum signal" or "the counter signal". I think last time I looked at it, there was some other means within the DBC file spec I wanted to use, but (ab)using units is possible too. I will dig up my old conversations and think about it.

can/common.h
jyoung8607: I took a pass at this in #650, see what you think of it.

vw_golf_mk4.dbc
jyoung8607: For DBCs where we want to keep the canonical signal names, I think the best thing is a DBC attribute that maps those signal names to their special usage. For DBCs that were created by hand and use the conventional CHECKSUM and COUNTER naming, we could fall through to using the name when the signal type attribute is entirely absent.

```
BA_DEF_ SG_ "opendbc_SignalType" ENUM "DEFAULT", "CHECKSUM", "COUNTER";
...
BA_DEF_DEF_ "opendbc_SignalType" "DEFAULT";
...
BA_ "opendbc_SignalType" SG_ 906 GRA_Checksum 1;
BA_ "opendbc_SignalType" SG_ 906 GRA_Neu_Zaehler 2;
(etc)
```

We could also get rid of the filename-based magic for selecting the checksum algorithm, even tweaking the algorithm per-message if we need to for situations like GM. I've even seen MQB DBCs with our annoying [pad byte lookup tables](https://github.com/commaai/opendbc/blob/47b79c4d5ab5adc0bdd9d228c3d9594da0355c49/can/common.cc#L123-L186) stored in the DBC in a similar way, which is something we could consider.

```
BA_DEF_ BO_ "opendbc_ChecksumAlgorithm" STRING;
...
BA_DEF_DEF_ "opendbc_ChecksumAlgorithm" "VOLKSWAGEN_PQ";
```

Something like this is a worthy goal, and I'm even willing to work on it at some point, but I'd rather not add a complex blocker for Volkswagen PQ dashcam upstream if that's okay. If you agree with the approach, I'd want to go back and unwind the signal names we changed years ago for MQB, and I can fix PQ at the same time.

jyoung8607:
This is ready for consideration. Panda tests will need updates to match, so when this is merged, I request the openpilot submodule reference be bumped right away, so that commaai/panda#964 can be released for consideration as well.

PR #634, opened 2022-06-09
remove checksum and counter alignment checks

jyoung8607:
Cherrypicked from #633 for independent consideration.

There's no reason we have to do these checks, other than we've always done them. They are now interfering with VW PQ, which has a few four-bit aligned counters, and a genuinely unaligned counter in Getriebe_1. Checksums and counters start at the bit the vehicle manufacturer chooses, we can't change them if we don't like them.


sshane:
I like having this check, can you just set the counter and checksum start bit to -1?

PR #178, opened 2019-08-20
Add DBC for JDM Honda Fit Hybrid 2018

AravindaDP:
_dual_can_nidec_2018.dbc added as modified version of _honda_2017.dbc
since JDM Fit Hybrid use different brake message (Can be use for
Clarity as well)

PR #635, opened 2022-06-10
Honda: merge Nidec DBCs

sshane:
These two DBCs had duplicate messages, (one or the other had) extra messages, and just a single conflicting message [thought to be different completely different](https://github.com/commaai/opendbc/pull/178/files) (`BRAKE_COMMAND`), but could be merged together cleanly with only two signals being alternates (`BRAKE_PUMP_REQUEST_ALT`, and `COMPUTER_BRAKE_ALT`).

The two DBCs that imported dual_can_nidec aren't even being used in openpilot, so for every other signal to match except these two, I think this is a good change.


sshane:
Since you've been working on the Clarity which uses the other signals, does this look good to you?

Diff in messages:

Screenshot from 2022-06-09 23-36-54

PR #636, opened 2022-06-11
Add back checksum start bit and length checks

sshane:


sshane:
@jyoung8607 look good?

jyoung8607:
Tested this locally, works for me!

PR #639, opened 2022-06-14
Hyundai canfd

shinkansan:
Hyundai CAN FD

confirmed with hyundai tuscon 2022


adeebshihadeh:
Looks like this is mostly the same as the EV6. I'd like to hold off on merging this until we have a matching car port PR.

sshane:
This accelerator signal doesn't look quite right for your car. The one added here does look correct though: https://github.com/commaai/opendbc/pull/680/files

Screenshot from 2022-11-09 14-12-21

sshane:
We already have alternate signals in our current DBC for your car, and some of these are incomplete or contradict what we currently have and what your car sends. Closing, but thanks for the PR!

PR #642, opened 2022-06-18
Chrysler: some unified signal names/definitions

sshane:
- Steering_Column_Angle_Status: Nicer scaling factors, looked identical on a 19 Pacifica
- BCM_1: Just renames message/signals
- DAS_3: Adds engine torque request, standstill bit, and splits ACC_STATUS_2 into two bits and adds available (if you read the SG_ comment, there's no behavior change)


sshane:
@adeebshihadeh should be good, feel free to merge with the openpilot PR once it passes and update the RAM branch.

PR #646, opened 2022-06-21
Add typing stubs for commonly used Cython classes, structures, methods and important attributes

JasonJShuler:
These stubs provide typing support for CANParser and CANDefine. Stubs do not impact the runtime behavior, but they provide the opportunity for type validation, dramatically simplify the development process and also aid dramatically in trying to understand the code.

Cython / CPython files, or even headers are not parsed by common type inferencing engines (like PyLance), causing them to show up as un-typed in editors like VSCode. The type hints allow not only presenting the objects as though they were first-class python objects, but docstrings can be added to aid developers in using the classes.

Remaining Tasks:
- [ ] Determine the appropriate python type to use for SignalValue.value (also the vl dicts)
- [ ] Implement remaining classes / data structures that are commonly used


adeebshihadeh:
Merged everything except the pyi files in #774. I support the intent behind them, but maintaining a separate file for them isn't worth it in my opinion.

PR #650, opened 2022-06-22
refactor per-manufacturer checksum calculation

jyoung8607:
Prep for #633:

* volkswagen -> volkswagen_mqb and associated PQ prep
* harmonize checksum function signatures
* reduce if-spaghetti by calling checksum functions through a saved pointer


jyoung8607:
Note: I haven't updated the Cython definition of Signal with the new function pointer. I'm not quite sure how that would look, or if it's really necessary since it's the last field in the struct and won't be referenced directly from Python. This seems to work fine with the opendbc built-in Honda checksum test, and with Panda's across-the-board safety tests.

https://github.com/commaai/opendbc/blob/e7cd3ebc893047bf6eb947c60d8e3196b506e8d3/can/common.pxd#L27-L33

jyoung8607:
I'm not in a huge rush, but FYI this is head-of-line blocking further work toward commaai/openpilot#24768. If the checksum function call refactor isn't to your taste, feel free to revert 3acd049a0b05c29c3094950f88253bf108bad547. The prior commit has just the requested signature changes and should be pretty uncontroversial.

can/common.pxd
```diff
@@ -9,6 +9,8 @@ from libcpp.vector cimport vector
 from libcpp.unordered_set cimport unordered_set
 
 
+ctypedef unsigned int (*calc_checksum_type)(uint32_t, const Signal&, const vector[uint8_t] &)
```
jyoung8607, line 12: Well that was easy, not sure how I got so turned around thinking it was harder. Thanks for the assist!

PR #651, opened 2022-06-22
Honda Civic 2022: add FCW bits

sshane:
AEB_PREPARE also seemed to be correct. Thanks to First for the helpful route with two FCWs!: `1a5d045d2c531a6d|2022-06-22--01-20-39`

Screenshot from 2022-06-22 14-44-58


sshane:
`AEB_PREPARE` from the current DBC is always the first to rise.

Can rule these out:
- Bit 42 rises same frame, but fell much earlier than AEB_PREPARE. Second FCW it fell same time as the last FCW bit but rose 0.5 seconds after AEB_PREPARE.
- Bit 17 and bit 49 ignored first FCW
- Bit 29 rose 10 frames after AEB_PREPARE for first FCW.

PR #654, opened 2022-06-24
vw_golf_mk4: translate signals to english as comments episode 2

rusefillc:
A bit more of the same

### DIFF
--- a/vw_golf_mk4.dbc
+++ b/vw_golf_mk4.dbc
@@ -1433,38 +1433,38 @@ BO_ 2000 Diagnose_1: 8 XXX
  SG_ DI1_KM_Stand_alt : 62|1@1+ (1,0) [0|1] "" XXX
  SG_ DI1_Zeit_alt : 63|1@1+ (1,0) [0|1] "" XXX
 
-CM_ SG_ 80 Checksumme_Airbag_1 "Checksum Airbag_1"
-CM_ SG_ 80 Zaehler_Airbag_1 "Counter Airbag_1"
+CM_ SG_ 80 Checksumme_Airbag_1 "Checksum Airbag_1";
+CM_ SG_ 80 Zaehler_Airbag_1 "Counter Airbag_1";
 CM_ SG_ 210 LM_Offset "centiNewton-meters for ease of calculation without FP math in Panda";
 CM_ SG_ 210 Vib_Amp "Steering wheel haptic, amplitude";
 CM_ SG_ 210 Vib_Freq "Steering wheel haptic, frequency";
-CM_ SG_ 416 Zaehler_Bremse_1 "Counter Bremse_1"
-CM_ SG_ 640 inneres_Motor_Moment "Engine Indicated Torque"
-CM_ SG_ 640 Fahrerwunschmoment "Driver Requested Torque"
-CM_ SG_ 640 mechanisches_Motor_Verlustmomen "Mechanical Torque Loss"
+CM_ SG_ 416 Zaehler_Bremse_1 "Counter Bremse_1";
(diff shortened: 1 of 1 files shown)


sshane:
I think the convention so far is not to include spaces in between messages

rusefillc:
@sshane huh? https://github.com/commaai/opendbc/pull/653/files was merged a few hours ago

sshane:
Right, we should revert that. We should do all DBCs or none, and it'd probably be its own PR if it's a standard.

rusefillc:
I do not follow since spaces were used prior to my merge but I hope you guys will figure out and document your convention if you have one.

https://github.com/commaai/opendbc/blob/5ce12a2558e5657b62936d6346b7b05baecf6646/vw_golf_mk4.dbc#L1437

jyoung8607:
I've seen several DBCs written by commercial tools with an extra newline between blocks of signal comments. I'm not married to either way of doing it, but neither way seems like a violation to me, and I like it if there's going to be a lot of comments.

rusefillc:
So when @sshane says "spaces" he means "empty lines between content lines"? I am not a native speaker that was confusing.

jyoung8607:
While any translation work is interesting, the most helpful would be translations and descriptions of the signals used by openpilot (commaai/openpilot#24768). We use a variety of signals within the following messages:

* Bremse_1
* Bremse_3
* Lenkhilfe_3
* Lenkwinkel_1
* Motor_3
* Airbag_1
* Bremse_5
* GRA_Neu
* Kombi_1
* Motor_2
* Lenkhilfe_2
* Gate_Komf_1
* ACC_GRA_Anziege
* SWA_1
* LDW_1
* HCA_1

Within those messages, the signals actively used in openpilot carstate and carcontroller are the most helpful.

jyoung8607:
BTW, if you are still actively working on a given PR (as seems the case here) use the "convert to Draft" function to let other potential contributors and reviewers know it's not yet ready to be reviewed and accepted.

vw_golf_mk4.dbc
```diff
@@ -1433,19 +1433,51 @@ BO_ 2000 Diagnose_1: 8 XXX
  SG_ DI1_KM_Stand_alt : 62|1@1+ (1,0) [0|1] "" XXX
  SG_ DI1_Zeit_alt : 63|1@1+ (1,0) [0|1] "" XXX
 
+CM_ SG_ 80 Checksumme_Airbag_1 "Checksum Airbag_1"
```
sshane, line 1436: Can you make sure add semicolons and remove the newlines?

rusefillc: @sshane do you mean https://github.com/commaai/opendbc/pull/654/commits/984a9ab3668812d494b3ad9e8c02ae47e8f8815a when you request new lines to be removed? Seems to be a negative value but it's your call.

@sshane please clarify what semicolors are you referring to. I am a bit confused if there is some pattern/style/convention I am missing or if that's a random subjective preference?

jyoung8607: The [DBC specification](http://mcu.so/Microcontroller/Automotive/DBC_File_Format_Documentation.pdf) calls for a trailing semicolon on the comment lines:

```suggestion
CM_ SG_ 80 Checksumme_Airbag_1 "Checksum Airbag_1";
```

rusefillc:
ooops, my bad. fixed.

vw_golf_mk4.dbc
adeebshihadeh: @rusefillc we generally don't have strict stylistic guidelines. it's up to the author to make their code tasteful, and we'll review as needed. in this case, I think you were right with the newlines. @sshane why did you want them removed?

PR #656, opened 2022-07-04
hyundai: fix FCA11 checksum and counter

gregjhogan:
required for https://github.com/commaai/openpilot/pull/25027


gregjhogan:
note that some vehicles most likely use the below checksum/counter from the old definition, but since everything was broken due to the mixed definition used by openpilot (which didn't generate valid counters or checksums for either) we are just adding support for the newer vehicles for now
```
 SG_ CR_FCA_Alive : 56|4@1+ (1,0) [0|15] "" ESC
 SG_ CR_FCA_ChkSum : 60|4@1+ (1,0) [0|15] "" ESC
 ```

PR #670, opened 2022-07-22
Chrysler: add Ram HD

realfast:
This adds the needed messages to incorporate the Ram HD trucks. 

I have changed the names of 678 and 658 to more accurate names. 

Updated signals to uppercase to be uniform.


adeebshihadeh:
Nice. In the future, a PR like this should be split up into two different ones: one for Ram HD support and the other for the style/name changes.

adeebshihadeh:
I removed the other changes, and moved the HD signals to use the generator.

realfast:
I think this was mixed up somehow. Here is a corrected version. 

https://github.com/realfast/opendbc/commit/2a79fb1290fd9a7a7b10372371d7bda7f6cbd5a5

PR #674, opened 2022-07-26
body: add knee and few other messages

briskspirit:
it is not backwards compatible


briskspirit:
reverted msg structure change

PR #675, opened 2022-07-29
Chrysler LONG Prep

realfast:
This includes more messages from the DAS_3 message to control acceleration and deceleration for long control. 

Engine rpm and torque data added.


adeebshihadeh:
I couldn't push to your branch for some reason, so I made #676 to use the generator.

PR #677, opened 2022-08-05
new generated file for Toyota Hilux 2017

ramsieber:
New file generated out of [toyota_new_mc_pt.dbc]. Adds additional messages applicable to Toyota Hilux 2015-2020.

No messages of the generating file were deleted. These messages may not apply to the Toyota Hilux.

### DIFF
--- a/toyota_hilux_2017_generated.dbc
+++ b/toyota_hilux_2017_generated.dbc
@@ -1,0 +1,554 @@
+VERSION ""
+
+
+NS_ : 
+	NS_DESC_
+	CM_
+	BA_DEF_
+	BA_
+	VAL_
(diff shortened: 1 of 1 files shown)


cydia2020:
I'll try to verify all the changes on my Prius, if the extra signals are present then we should add them to the general Toyota file.

cydia2020:
Looks like the new signals are somewhat limited to the IMV platform, maybe we can add a new generator file for these new signals

ramsieber:
I just tested tested the signals on the Toyota Hilux model GUN125L-DTFLHV with the engine 2GD-FTV. Don't really know which vehicles share the same platform. A new generator file would probably be good.

PR #678, opened 2022-08-10
GM: add counter and checksum for ASCMSteeringButton

JasonJShuler:
Added signals for Rolling Counter and Checksum to ASCMSteeringButton message.

Needed if OP is going to send button presses to the camera (To hopefully disengage the ACC)


sshane:
Will merge once we confirm we need this and it works

PR #681, opened 2022-08-14
Add 2021+ Honda Odyssey

csouers:
This car is a bit more modern but continues to use the older EPS signal styling.


generator/honda/honda_odyssey_touring_2021_can.dbc
```diff
@@ -0,0 +1,25 @@
+CM_ "IMPORT _honda_common.dbc";
+CM_ "IMPORT _bosch_2018.dbc";
+CM_ "IMPORT _bosch_adas_2018.dbc";
+CM_ "IMPORT _steering_sensors_b.dbc";
+
+BO_ 419 GEARBOX: 8 PCM
+ SG_ GEAR : 7|8@0+ (1,0) [0|256] "" EON
+ SG_ GEAR_SHIFTER : 35|4@0+ (1,0) [0|15] "" EON
```
adeebshihadeh, line 8: are the gear shifter signals really slightly different between all the cars? @sshane did you already look into this when you did the Honda DBC cleanup?

sshane: We can't easily make this common, because the `419 GEARBOX` message is used in DBCs that import both Nidec and Bosch common messages. And of all the DBCs that import those common messages, there exists different GEARBOX messages (not just definitions, messages).

sshane: Actually this is functionally identical to the acura_rdx_2020 DBC. Both the definitions work

PR #683, opened 2022-08-14
Ford: fix steering command signal ranges

incognitojam:
The signal ranges seem to be "off center" in the Ford DBC... This might be causing the steering control issues.

Screenshot from 2022-08-14 20-55-34
Show above is the LCA message when the system can't detect any lane lines. You can see that the path angle signal is non zero.

Example segments:
`59a107f5793d9cc0|2022-08-06--14-11-17--0` at 2.5 seconds
`59a107f5793d9cc0|2022-08-07--00-13-33--0` at 2.9 seconds
`86d00e12925f4df7|2022-08-11--16-45-58--0` at 10.6 seconds


incognitojam:
Just noticed that the DBC is invalid, I changed the [ranges] but not the (step, min) part... Maybe there should be a CI test for this?

incognitojam:
Example segments:
`59a107f5793d9cc0|2022-08-06--14-11-17--0` at 2.5 seconds

PR #687, opened 2022-08-19
Enable Toyota counter checks

sshane:
It looks like 4 of our supported platforms use the PCS message for cruise control, and thus set the counter:
- LEXUS RC 2020
- LEXUS IS 2018
- LEXUS IS 2018
- LEXUS RC 2020


sshane:
It appears that the 2022 RAV4's set their STEERING_LKA checksum correctly, it increments by 1 unlike all other Toyota (checked with test_models).

generator/toyota/_toyota_2017.dbc
```diff
@@ -117,7 +117,7 @@ BO_ 614 STEERING_IPAS: 8 IPAS
  SG_ CHECKSUM : 63|8@0+ (1,0) [0|255] "" XXX
 
 BO_ 643 PRE_COLLISION: 7 DSU
- SG_ COUNTER : 7|8@0+ (1,0) [0|255] "" XXX
+ SG_ _COUNTER : 7|8@0+ (1,0) [0|255] "" XXX
```
sshane, line 120: Expected to only be set on cars that use this message for cruise control

PR #688, opened 2022-08-19
CANParser: message counters affect can_valid

sshane:
The MessageState counters currently only block parsing signals after 5 consecutive expected counter failures. This PR sets can invalid at the same time, triggering an immediate disable in openpilot. At the same time, in panda, it also sets controls allowed to 0 after 5 invalid counters in a row.

Both panda and CANParser also have the same behavior of incrementing/decrementing the valid counter cnts, not setting to 0 immediately.

### DIFF
--- a/can/parser_pyx.pyx
+++ b/can/parser_pyx.pyx
@@ -26,8 +26,6 @@ cdef class CANParser:
   cdef readonly:
     dict vl
     dict vl_all
-    bool can_valid
-    bool bus_timeout
     string dbc_name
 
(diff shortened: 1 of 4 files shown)


can/parser_pyx.pyx
```diff
@@ -15,8 +15,6 @@ import os
 import numbers
 from collections import defaultdict
 
-cdef int CAN_INVALID_CNT = 5
-
```
sshane, line 19: Need to figure out why this was added and if we can remove it, this is a threshold on the parser's valid which is already using a threshold. This should be in the C++ code anyway.

adeebshihadeh (approved):

PR #690, opened 2022-08-19
CANParser: move can valid threshold into parser

sshane:
Just a refactor, does not change behavior

### DIFF
--- a/can/parser.cc
+++ b/can/parser.cc
@@ -201,11 +201,7 @@ void CANParser::update_string(const std::string &data, bool sendcan) {
   auto cans = sendcan ? event.getSendcan() : event.getCan();
   UpdateCans(last_sec, cans);
 
-  can_invalid_cnt++;
-  if (CanValid(last_sec)) {
-    can_invalid_cnt = 0;
-  }
-  can_valid = can_invalid_cnt < CAN_INVALID_CNT;
+  UpdateValid(last_sec);
 }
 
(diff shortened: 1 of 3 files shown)


can/parser_pyx.pyx
```diff
@@ -110,10 +106,7 @@ cdef class CANParser:
     cdef unordered_set[uint32_t] updated_addrs
 
     # Update invalid flag
-    self.can_invalid_cnt += 1
-    if self.can.can_valid:
-      self.can_invalid_cnt = 0
-    self.can_valid = self.can_invalid_cnt < CAN_INVALID_CNT
+    self.can_valid = self.can.can_valid
```
adeebshihadeh, line 109: make this a property and return `self.can.can_valid`

can/parser.cc
```diff
@@ -275,7 +275,7 @@ void CANParser::UpdateCans(uint64_t sec, const capnp::DynamicStruct::Reader& cms
 void CANParser::UpdateValid(uint64_t sec) {
   const bool show_missing = (last_sec - first_sec) > 8e9;
 
-  can_valid = true;
+  bool _valid = false;
```
adeebshihadeh, line 278: this is always false

adeebshihadeh (changes requested):
Does this even work? Always make sure the tests are passing before requesting a review.

sshane:
typo, my bad

can/parser.cc
sshane: typo, my bad

adeebshihadeh (approved):

PR #694, opened 2022-08-23
Hyundai Genesis GV70 2022 HDA1 Port 

zunichky:


adeebshihadeh:
Thanks, we should have everything needed after #680

PR #695, opened 2022-08-24
Toyota: add universal units signal

sshane:
Some notes:

- Message exists on all Toyota and Lexus. Signal on all Toyota and Lexus is never 0 (unlike `UI_SETTING->UNITS` or `RSA2->SPUNT`, which then fall back to msg `AVN1S21`)
- Signal seems to be identical to `AVN1S21->A_SPDUNT` on cars that have 0x435 on the bus (not all)
- Describes not only units, but gas mileage represenation preference (for ex. `litres per 100km vs km per litre`, why there's four values)


adeebshihadeh:
> * Describes not only units, but gas mileage represenation preference (for ex. `litres per 100km vs km per litre`, why there's four values)

This should be reflected in the VAL lines.

PR #696, opened 2022-08-25
add script to generate mapping for each car fingerprint to dbc name

incognitojam:


adeebshihadeh:
This should go in cabana and just use the prebuilt openpilot container.

PR #698, opened 2022-08-28
vw_golf_mk4: a bit of progress

rusefillc:
A bit more of the same - English translation as comments

### DIFF
--- a/vw_golf_mk4.dbc
+++ b/vw_golf_mk4.dbc
@@ -1541,14 +1541,12 @@ CM_ SG_ 1152 Vorgluehlampe__Motor_5_ "Glow light";
 
 CM_ SG_ 1160 Zaehler_Motor_6 "Counter Motor_6";
 CM_ SG_ 1160 Hoeheninfo__Motor_6_ "Altitude Correction";
-CM_ SG_ 1160 Istmoment_f_r_Getriebe "Actual Torque";
-CM_ SG_ 1160 Sollmoment_f_r_Getriebe "Target Torque";
+CM_ SG_ 1160 Istmoment_f_r_Getriebe "Actual torque for gear";
+CM_ SG_ 1160 Sollmoment_f_r_Getriebe "Target torque for gearbox";
 CM_ SG_ 1160 Checksumme_Motor_6 "Checksum Motor_6";
 CM_ SG_ 1160 GRA_Sollbeschleunigung "GRA target acceleration";
 CM_ SG_ 1160 Ruckmeldung_Momenten "Feedback torque-integral gear intervention";
 CM_ SG_ 1160 Hoeheninfo__Motor_6_ "Altitude info";
-CM_ SG_ 1160 Istmoment_f_r_Getriebe "Actual torque for gear";
-CM_ SG_ 1160 Sollmoment_f_r_Getriebe "Target torque for gearbox";
 
 CM_ SG_ 1344 Zahler_Getriebe_2 "Counter Getriebe_2";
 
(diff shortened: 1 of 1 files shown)


rusefillc:
@adeebshihadeh @sshane poke :)

sshane:
@jyoung8607 does this look good?

vw_golf_mk4.dbc
```diff
@@ -303,7 +303,8 @@ BO_ 1408 Motor_Flexia: 8 XXX
  SG_ Multiplex_Schalter_Motor_Flexia M : 0|1@1+ (1,0) [0|0] "" XXX
 
 BO_ 1416 Motor_7: 8 XXX
- SG_ Frei_Motor_7_3 : 40|24@1+ (1,0) [0|0] "" XXX
+ SG_ Oltemperatur : 56|8@1+ (1,0) [0|0] "" XXX
+ SG_ Frei_Motor_7_3 : 40|16@1+ (1,0) [0|0] "" XXX
```
jyoung8607, line 307: Let's revert this change. If we're going to clean up sorting of signals, they would go in ascending order rather than descending, and it should be done in a standalone PR for that purpose rather than intermixed with translation/commentary updates. And I'm not in a great hurry to perform/review that sort at this particular time.

vw_golf_mk4.dbc
```diff
@@ -321,7 +322,8 @@ BO_ 1160 Motor_6: 8 XXX
  SG_ Frei_Motor_6_4 : 58|2@1+ (1,0) [0|0] "" XXX
  SG_ ltemperaturschutz : 57|1@1+ (1,0) [0|0] "" XXX
  SG_ GRA_Bremseingriff_Freigabe : 56|1@1+ (1,0) [0|0] "" XXX
- SG_ Frei_Motor_6_3 : 40|16@1+ (1,0) [0|0] "" XXX
+ SG_ Frei_Motor_6_3 : 48|8@1+ (1,0) [0|0] "" XXX
+ SG_ Ruckmeldung_Momenten : 40|8@1+ (0.39,0) [0|100] "" XXX
```
jyoung8607, line 326: Same, revert.

jyoung8607 (changes requested):

vw_golf_mk4.dbc
rusefillc: @jyoung8607 let's confirm it's not a misunderstanding. I belive this is NOT a sorting change here.
"Frei_Motor_XX_XX" in both cases seems like "unused unmapped areas" and in both cases I am reducing the lenght of "unused unmapped Frei_Motor" while adding a new field definition which takes a part of what was "unmapped"

like here I believe that I am ading Oltemperatur and reduce unmapped placeholder from 24 bits to 16 bits

Question: shall I revent addition of the new field?

vw_golf_mk4.dbc
rusefillc: Please confirm if I should revert new field "Ruckmeldung_Momenten"

rusefillc:
@jyoung8607 did you have a chance to revisit?
@sshane @adeebshihadeh poke :)

rusefillc:
@jyoung8607 @sshane @adeebshihadeh poke

sshane:
@jyoung8607 gtm?

jyoung8607 (approved):
Eventually the filler/placeholder signals tweaked in this PR (such as `Frei_Motor_7_3`) should really be discarded, and we should replace the whole of Motor_7 (among others) with its current/modern PQ35/PQ46 equivalent as we've done with other messages, but this is close enough for now.

rusefillc:
@sshane poke?

PR #707, opened 2022-09-09
Toyota: radar dsu tss-p radar support, aka nodsu

Smartype:
Add support for the continental radar used in CAMRY-18, CHR.


cydia2020:
Can we get this merged? @adeebshihadeh

Users have been testing this extensively on my fork and it seems to work well besides the scale factor of `LONG_DISTANCE` which has been recently changed to .03

@Smartype wanna change that here as well?

Smartype:
@adeebshihadeh am I supposed to submit the radar support PR to the openpilot repo at the same time?

PR #709, opened 2022-09-19
civic22: add standstill bit

royjr:
@sshane Sorry I wiped out the last standstill bit fix, does this look good?

Screen Shot 2022-09-19 at 10 18 32 AM


sshane:
Yep, added there

PR #712, opened 2022-09-25
Hyundai corner radar DBC

gregjhogan:
The layout is based on the 2020 Palisade rear blind spot firmware.  The CAN messages are on a private bus between the radars which is [CAN-FD 1.0](https://github.com/commaai/panda/issues/916) with arbitration at 1 MHz and data at 2 MHz.

This isn't similar to other radar messages parsed by openpilot
* `0x100`/`0x200` seems to be summary info about `0x101`/`0x201` messages
* `0x101`/`0x201` sends groups of 1 to 13 messages at 20 Hz (`MESSAGE_ID` is the index inside the group, starts at 1)
  * each group is a set of radar points for a given time step
  * the number of messages in a group varies based on how many points are being tracked
  * each message has 5 points so there is a maximum of 65 points per time step
  * the points move around between messages in `0x101`/`0x201` as more or less points are tracked, so it isn't trivial to follow a single point through time
* `0x104`/`0x204` is a CRC16 that covers `0x100`/`0x200` and all `0x101`/`0x201` messages for a time step

There is probably a signal strength/quality in there somewhere, but I haven't found it.

Note that the alignment of the start of each point in the `0x101`/`0x201` messages is wacky because each point is 101 bits long

example from `5decac85ab437ed1|2022-04-16--12-43-15/18`

palisade_bsm_points


gregjhogan:
@adeebshihadeh this is good to merge if it looks good to you

sunnyhaibin:
Would this PR also apply to the 2020 Hyundai Sonata, or it has not been investigated yet?

PR #723, opened 2022-10-20
HKG CAN-FD: rename brake pressure signal, add position signal

sshane:


sshane:
This signal definitely looks meaningfully different, will try to find out what uses what

Screenshot from 2022-10-20 13-17-29

hyundai_canfd.dbc
```diff
@@ -84,6 +84,7 @@ BO_ 101 BRAKE: 32 XXX
  SG_ CHECKSUM : 0|16@1+ (1,0) [0|65535] "" XXX
  SG_ COUNTER : 16|8@1+ (1,0) [0|255] "" XXX
  SG_ BRAKE_PRESSED : 57|1@1+ (1,0) [0|3] "" XXX
+ SG_ BRAKE_PRESSURE : 40|16@1+ (1,0) [0|695] "" XXX
```
sshane, line 87: Only used 10 bits at max force/position, but the rest were zeros, so map the full signal?

sshane:
Seems to be correct. The signal in EPS_STATUS shows the computer-applied pressure when we cancel cruise while the new position signal stays constant (actual user position of brake pedal) https://imgur.com/a/4sfR7Al

PR #727, opened 2022-10-24
CAN-FD HKG - Speed Limits from Dashboard

ioniq5-cz:
There are 3 speed limit bytes in 5:1fa for HKG CAN-FD cars which match the indicated speed in the cars built in dashboard:

image


zunichky:
Confirmed signals work on Genesis HDA1. It looks like SPEED_LIMIT_3 is the speed limit + offset that you configure in the settings. Mines set to +3, and signal 3 is always +3 from signal 1/2. I'll change this this evening to +5 to confirm.

ioniq5-cz:
correct. SPEED_LIMIT_# is sped limit + offset in HKG dash - this was already confirmed by a user in Sunny's discord. Thank you for the confirmation.

sunnyhaibin:
This is also confirmed to work on the 2022 Hyundai Tucson Hybrid **non-HDA**, CAN-FD. Great find!
- `36e10531feea61a4|2022-08-16--18-14-13` (preserved when the port was done)

image

PR #728, opened 2022-10-26
CAN-FD HGK - additional icons and chime from cars dash

ioniq5-cz:
image
image
image

the chime, speed change indicator and sound will be used downstream with SunnyPilot. The icons (arrow up and down) might be useful in the future


adeebshihadeh:
Nice finds. Feel free to even make PRs for things that you're not completely sure about; for example, it's very useful to just know which ECU is sending a message even if you don't know what any of the signals are.

PR #673, opened 2022-07-25
Hyundai: Car Port for Tucson Hybrid 2022

sunnyhaibin:
Prerequisite of:
- https://github.com/commaai/openpilot/pull/25276

PR #729, opened 2022-10-26
Hyundai CAN-FD: fix set speed signal size

ioniq5-cz:
Car owners using kph instead of mph are having a major bug with the current implemtation:

https://user-images.githubusercontent.com/111085918/198109972-ca69a3d5-3069-4a13-af6e-5480f580e3fe.mov

In EU the allowed highway speed limit is 130 kph and the current SET_SPEED signal is only 7 bits. The max value that 7 bits could fit is 127, which is why it was labeled as the max value.

We need to change this to 8 bits in order to facilitate cars with kph instead of mph (new max 255).


hyundai_canfd.dbc
```diff
@@ -212,7 +212,7 @@ BO_ 416 CRUISE_INFO: 32 ADRV
  SG_ ZEROS_4 : 183|4@0+ (1,0) [0|63] "" XXX
  SG_ ZEROS_6 : 119|16@0+ (1,0) [0|65535] "" XXX
  SG_ ZEROS_7 : 63|8@0+ (1,0) [0|255] "" XXX
- SG_ SET_SPEED : 102|7@0+ (1,0) [0|127] "km/h or mph" XXX
+ SG_ SET_SPEED : 102|8@0+ (0.5,0) [0|255] "km/h or mph" XXX
```
incognitojam, line 215: I don't think you need to change the scaling to 0.5

ioniq5-cz: from looking at it in Cabana I think the values are else double then what they should be (the speed limit):

image

vs

image

incognitojam: In the video you changed the set speed to 180km/h, where is that in the graph/cabana?

I don't think the factor should be 0.5 because if you have an 8 bit signal (2^8) then 2^8 * 0.5 is only 128 again.

ioniq5-cz: here is the chart from cabana from the video:

image

see the orange house in the background and the grey car to the left. The cart is going only till 127 as you can see. 

Thank you for explaining the math to me. I will check further until then.

ioniq5-cz: above is the original dbc file. When i do the following change its reflecting reality:

image

 SG_ SET_SPEED : 103|8@0+ (1,0) [0|127] "km/h or mph" XXX

No factor anymore used and I believe we have the solution. I have updated the PR. Only potential problem I can see is that bit 103 is occupied now by 2 signals:

 SG_ NEW_SIGNAL_4 : 103|1@1+ (1,0) [0|1] "" XXX

What do you suggest?

ioniq5-cz: It looks like NEW_SIGNAL_4 is actually nowhere used. I remove it in this PR. Let me know if I missed something.

incognitojam: @sunnyhaibin does this look okay? I see you added that in https://github.com/commaai/opendbc/pull/673

ioniq5-cz: here is a drive with the new DBC values/changes: 2833de0636d26d55|2022-10-28--10-39-21--0 

No issues whatsoever!

incognitojam: just wondering what NEW_SIGNAL_4 might have been used for, it looks like it's in [`carstate.py`](https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/carstate.py#L510) but not used for anything right now, but that would have to be removed too

sunnyhaibin: @incognitojam Looks good to me! I believe when the DBC for HKG CAN-FD was initially generated, bit 103 on 0x1A0 might have been accidentally labeled as its own signal. I checked the route attached and it matches with what's showing on the dashboard and openpilot's "MAX" value, which previously `CRUISE_INFO | SET_SPEED` would "reset" back to 0 and incremented when the set speed on the car's dashboard continued to increase.

image

sunnyhaibin:
@ioniq5-cz You may want to also open another PR to correspond with the definition changes in this PR, specifically removing `CRUISE_INFO | NEW_SIGNAL_4` in `carstate.py`: 

https://github.com/commaai/openpilot/blob/8de9bbaa73ea94b966b1161fb630bf3cac72c764/selfdrive/car/hyundai/carstate.py#L510

ioniq5-cz:
opened PR here https://github.com/commaai/openpilot/pull/26276

PR #730, opened 2022-10-26
Chrysler Auto_Start_Stop button and dashlight

SupaTWC:
Button: 0:355 changes from 0000 to 1001 when the button is pressed.
Dashlight (the light on the button): 0:278 changes to 1 when Auto Start is turned OFF.

Confirmed on 2 different Pacificas.
image
image
image


SupaTWC:
Not sure why it failed. Copied and pasted from "Save DBC"

incognitojam:
You are modifying the `_generated.dbc`. You want to add this message to either the `_common.dbc` or the extras DBC for this car, depending on whether this signal is also on other cars
https://github.com/commaai/opendbc/tree/master/generator/chrysler

PR #732, opened 2022-10-27
hyundai_2015_ccan.dbc: fix signal scale

seonghoony:
The scale of signal `CF_Ems_IsgStat2` should be 1.0 instead of 2.0 because two bits are assigned for the signal. Otherwise, the range of the signal must be [0.0|6.0] rather than [0.0|3.0].


sshane:
Do you know if there was any reason for the 2x?

seonghoony:
By the way, the CAN database is quite accurate and it helps my research a lot. OpenDBC is a great project for me.

PR #736, opened 2022-11-02
GM: add new undefined brake-related signals

sshane:
used a script to find all bits similar to CS.brakePressed, ended up finding bits and continuous signals. may be helpful in resolving some more faults


sshane:
You can see the new brake signals are close, but not insignificantly different to the current signals. Might be something useful here!

Screenshot from 2022-11-02 15-13-30

PR #738, opened 2022-11-03
Created XV40 Camry base file, reversed one signal.

cdeyong2117:
Just has the one signal for now, but I'll see if I can get other signals added moving forward. Tested using data from a 2011 Camry (2AR-FE), and cluster from a 2009 Camry (2AZ-FE).

### DIFF
--- a/toyota_camry_XV40.dbc
+++ b/toyota_camry_XV40.dbc
@@ -1,0 +1,42 @@
+VERSION ""
+
+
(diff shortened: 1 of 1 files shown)


sshane:
Do any of our current DBCs match up with the CAN data on the car? `_toyota_2017` or `toyota_2017_ref_pt` perhaps?

PR #744, opened 2022-11-09
Update bmw_e9x_e8x.dbc

Joel-S-Prodrive:
Extra Tx messages found and mapped to Kombi meter, as discovered on bench test when replaying CAN Log from full vehicle.

### DIFF
--- a/bmw_e9x_e8x.dbc
+++ b/bmw_e9x_e8x.dbc
@@ -1,198 +1,200 @@
 VERSION ""
 
 
-NS_ :
-    NS_DESC_
-    CM_
-    BA_DEF_
-    BA_
-    VAL_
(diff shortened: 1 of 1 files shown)


rusefillc:
@Joel-S-Prodrive with massive formatting change this PR seems impossible to review? how does a human separate formatting change from content change?!

Joel-S-Prodrive:
Hi @rusefillc,

I don't exactly understand your question, but I used Vector CANdB++ to edit the DBC file. If you already have this or an equivalent tool then you should be able to see the changes I made, most of which to 0x130. This is still an ongoing process so there are likely more commits from my end once I get back to it.

I hope this helps,
Joel

PR #745, opened 2022-11-09
More magic bytes for the VAC CRC calculation

crasbe:
Hello everyone,

I found the OpenDBC project and the documentation on the CRC calculation really helped with reverse engineering the protocol for the VAG MEB chargers.
Some information about that can be found in the Openinverter Forum: https://openinverter.org/forum/viewtopic.php?p=48154#p48154 (CAN logs are in the post following mine)

In the process of reverse engineering, we found 6 addresses (and 3 more addresses from the Starter Generator https://openinverter.org/forum/viewtopic.php?t=1520 CAN logs) that were not in this file yet, so I wrote a Python Script (which you can find attached (Github doesn't like .py files, so it's compressed in a .zip)) which brute forces the magic bytes for each address and verifies it against the log. The logs are in SavvyCAN format in .csv files, the script probably needs to be adapted for other formats. The script is free to use with attribution, so you can add it to the repository if you want to or change it in any way or form you want.
The script doesn't work with extended CAN IDs, but I don't know if they have CRC checksums in the same format either.

[Magic Byte Cracker.zip](https://github.com/commaai/opendbc/files/9972328/Magic.Byte.Cracker.zip)

I do not know what the addresses do or what a meaningful name would be, so please excuse the "????" annotations in the comments.

Best regards,
Chris


adeebshihadeh:
Thanks for the PR, however we can't spend the time to validate this since it's not used in openpilot.

PR #757, opened 2022-12-04
Update honda_civic_ex_2022_can_generated.dbc

tapperoo42:
I'm working with @sunnyhaibin to create a port for my 2023 Acura Integra 6MT. Using the honda_civic_ex_2022 profile, I have a CAN error message with the primary messages from "tmux a" as 0x191 and 0x158 are not seen.

Currently, the clutch position & gear / gear "type" differences have been identified plus how they interplay as bit pairs vs individual signals. 

6MT_SHIFTER and CLUTCH_PEDAL can define all major driving states (Shifter: D/N/R as 0,1,2 & Clutch pedal state: slipping, released, or depressed as 0,1,2)

6MT_GEARBOX can define all the gears / transmission state BUT it does need SHIFTER value to differentiate "N" vs 1st Gear.  Value 0 is the clutch pedal being pressed. 
Value 1 is "N" or "1st gear". 
Values 2-6 are the respective gears.
Value 14 is "R"

Reference route: 1b7349c28512586d|2022-12-01--17-02-47--3

Acura 6MT Clutch-Shifter-Gear_Select

image


vanillagorillaa:
You won't want to actually add these new signals directly into the "honda_civic_ex_2022_can_generated.dbc" file. You want to add them to the "honda_civic_ex_2022_can.dbc" (or if they apply to more vehicles, find the appropriate "base" file and add them there), then run the generator script. Look in /generator/honda for more information.

tapperoo42:
Thanks for explaining, I'll add the signals into the appropriate file.

vanillagorillaa:
Naming convention should follow how the other ones are. For example, honda_fit_ex_2018_can_generated.dbc. So you would want something like acura_integra_(insert trim level here)_2023_can_generated.dbc. No need to put 6MT in there. 

Also, not sure if needed, but unless there are a changes in messages between this vehicle and another dbc then you won't actually need a whole new DBC, just add the new signals (as long as they don't change).

tapperoo42:
The Manual Transmission is distinct from the rest of Honda's line up. It does not have 0x158 and 0x191 signals that is found on in the CIVIC_EX dbc. So, it is missing the typical drive gear signal and particularly the XMISSION speed. The Manual Transmission has its own signal group for the 6MT to identify the specific gear and some form of transmission RPM. It might be related to how the car handles rev matching. 

The 6MT only comes in the A-Spec + Tech trim. The Auto trims are base, A-Spec, and A-Spec+Tech. So, naming the DBC using 6MT made sense based on the signal differences vs the AT and that the 6MT is only offered in the single trim.

sshane:
0x1dd (`477 6MT_GEAR`) has no address conflicts with honda_civic_ex_2022_can, can you not just add it to that DBC? Everything else looks identical.

tapperoo42:
The manual transmission trim doesn't have the transmission speed signal. Specifically, 344 ENGINE_DATA
["ENGINE_DATA"]["XMISSION_SPEED"] is not available. Would it be a case of saying, if it's a manual transmission vehicle, use these parameters from the Civic DBC or to use a DBC that omits the signals that aren't present?

PR #758, opened 2022-12-04
Add Subaru Global 2022 DBC

martinl:
This PR adds Subaru Global 2022 DBC with new message ES_LKAS_2, currently only Forester 2022 is using it instead of ES_LKAS


generator/subaru/subaru_global_2022.dbc
```diff
@@ -0,0 +1,14 @@
+CM_ "IMPORT subaru_global_2017.dbc";
```
sshane, line 1:
```suggestion
CM_ "IMPORT subaru_global.dbc";
CM_ "IMPORT subaru_global_2017.dbc";
```

sshane:
I see we use the convention of making new DBCs with new signal additions for Subaru. I don't feel too strongly, but I don't see a big problem with combining everything to one DBC if there's no signal/address conflicts. Especially since we aren't naming this new signal the same name as the current one since it's in the base DBC already

generator/subaru/subaru_global_2022.dbc
```diff
@@ -0,0 +1,14 @@
...
+
+BO_ 292 ES_LKAS_2: 8 XXX
```
sshane, line 3: Is `ES_LKAS_State` (whatever else UI msgs) the same?

sshane, line 3:
```suggestion
BO_ 292 ES_LKAS_GEN2_5: 8 XXX # or _ALT?
```

neither are great names, but mark this as resolved if we don't yet know what is new about these cars. we can also make `_subaru_global` truly global (remove `ES_LKAS`) and put it into subaru_global_2017, so we can name this one `ES_LKAS` and require no changes in openpilot for the message name

generator/subaru/subaru_global_2022.dbc
```diff
@@ -0,0 +1,14 @@
...
+ SG_ COUNTER : 8|4@1+ (1,0) [0|1] "" XXX
+ SG_ LKAS_Request : 12|1@1+ (1,0) [0|1] "" XXX
+ SG_ LKAS_Output : 40|17@1- (-1,0) [0|1] "" XXX
+ SG_ SET_3 : 60|2@1+ (1,0) [0|1] "" XXX
+
+BO_ 282 EPS_LKAS: 8 XXX
+ SG_ CHECKSUM : 0|8@1+ (1,0) [0|1] "" XXX
+ SG_ COUNTER : 8|4@1+ (1,0) [0|1] "" XXX
+ SG_ LKAS_OUTPUT_SENSOR_1 : 24|17@1- (1,0) [0|1] "" XXX
+ SG_ LKAS_OUTPUT_SENSOR_2 : 47|17@1- (1,0) [0|1] "" XXX
```
sshane, line 14: does `281 Steering_Torque` not exist? one of these is probably driver torque and the other EPS output torque

generator/subaru/subaru_global_2022.dbc
jnewb1: I like making _subaru_global truly global. I'll do that.

generator/subaru/subaru_global_2022.dbc
jnewb1: I'll remove this for not as its not used, but I found what these are while investigating steering faults, I'll make a separate PR for that.

generator/subaru/subaru_global_2022.dbc
jnewb1: as far as we know, yes

sshane: You can do that cleanup later if you want

jnewb1: I'll do _ALT for now. Once I get more info about other cars that may use this message, I'll revisit this.

generator/subaru/subaru_global_2022.dbc
```diff
@@ -0,0 +1,9 @@
+CM_ "IMPORT _subaru_global.dbc";
+CM_ "IMPORT subaru_global_2017.dbc";
+
+BO_ 292 ES_LKAS_ALT: 8 XXX
+ SG_ CHECKSUM : 0|8@1+ (1,0) [0|1] "" XXX
+ SG_ COUNTER : 8|4@1+ (1,0) [0|1] "" XXX
+ SG_ LKAS_Request : 12|1@1+ (1,0) [0|1] "" XXX
+ SG_ LKAS_Output : 40|17@1- (-1,0) [0|1] "" XXX
+ SG_ SET_3 : 60|2@1+ (1,0) [0|1] "" XXX
```
sshane, line 9: I think it's fine to add to the existing DBC now

jnewb1: done

PR #762, opened 2022-12-13
HKG CAN-FD: add brake error signal

sshane:


hyundai_canfd.dbc
```diff
@@ -173,6 +173,8 @@ BO_ 373 TCS: 24 XXX
  SG_ DriverBrakingLowSens : 84|1@1+ (1,0) [0|1] "" XXX
  SG_ ACC_REQ : 68|1@0+ (1,0) [0|1] "" XXX
  SG_ ACCEL_REF_ACC : 48|11@1- (1,0) [0|1023] "" XXX
+ SG_ BRAKE_ERROR : 67|2@0+ (1,0) [0|1] "" XXX
+ SG_ BRAKE_ERROR_2 : 98|2@1+ (1,0) [0|3] "" XXX
```
sshane, line 177: Confimed with 1 route out of 4 that BRAKE_ERROR was set while BRAKE_ERROR_2 was not

sshane: See PR https://github.com/commaai/openpilot/pull/26781 for more info

PR #764, opened 2022-12-15
Bump pylint version

sshane:
A bug (https://github.com/PyCQA/pylint/issues/3098) was fixed in 2.5.0 relating to detecting abstract subclasses, bump to 2.5.2 like panda.

@adeebshihadeh somehow the panda docker is installing 2.4.3 in CI, causing a PR to fail static analysis which shouldn't, can that be affected by opendbc's pylint version? That's why I'm bumping it

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,6 @@ flake8==3.7.9
 Jinja2==3.0.3
 numpy==1.21.0
 pycapnp==1.0.0
-pylint==2.4.3
+pylint==2.5.2
 pyyaml==5.4
 scons


incognitojam:
it does look like the panda docker installs the opendbc requirements.txt *after* the panda requirements.txt
https://github.com/commaai/panda/blob/master/Dockerfile#L90

sshane:
Yes I see that now, looks like this should fix it! It seems not great that opendbc can affect the panda docker. They shouldn't be mismatched causing subtle bugs like this

PR #777, opened 2023-01-19
CANpacker/parser: support muxed signals

pd0wm:
DBC specification is as follows https://pisnoop.s3.amazonaws.com/snoop_help_dbc.htm:

> the red capital M defines the special multiplexor signal which defines the "version" of the message;
> the small red m3 defines that the signal TemperatureOutdoorsMultiplexed is only to be used when MultiplexIndexSignal has value 3;
> when transmitting, the sender should cycle through all of the index values used (0, 3, 11), filling the overlapped signal location with the appropriate data at each send.
> Normally consecutive index values are used, but this example shows that non-consecutive values are supported.

```
BO_ 1845 TemperatureMsg: 8 Vector__XXX
 SG_ MultiplexIndexSignal M : 39|4@0+ (1,0) [0|0] "" Vector__XXX
 SG_ NormalSignalAlwaysPresent : 16|4@1- (1,0) [0|0] "" Vector__XXX
 SG_ TemperatureIndoorsMultiplexed m0 : 7|16@0- (0.1,-40) [0|0] "degC" Vector__XXX
 SG_ TemperatureOutdoorsMultiplexed m3 : 7|16@0- (0.1,-40) [0|0] "degC" Vector__XXX
 SG_ TemperatureUndergroundMultiplexed m11 : 7|16@0- (0.1,-40) [0|0] "degC" Vector__XXX
```

Example message in a DBC file: https://github.com/commaai/opendbc/blob/master/vw_golf_mk4.dbc#L1412 and https://github.com/commaai/opendbc/blob/master/tesla_can.dbc#L422

Supposedly Tesla makes big use of these multiplexed signals.

Implementation details I'm not yet sure about. Suggestions are welcome!
 - How do frequency checks work for multiplexed signals? Separate frequency for each "sub message"? Does it matter if they are not used in openpilot yet?
 - How do you tell the packer which version of the message to build. I see three options:
   1) Specify the multiplexor signal manually. E.g `packer.make_can_msg("TemperatureMsg", 0, {'MultiplexIndexSignal': 0, ...})`
   2) Add the multiplexor signal as a suffix to the message name. E.g. `TemperatureMsg_0`, `TemperatureMsg_1`...
   3) Figure out the correct multiplexor based on the passed signals.


jnewb1:
There is also extended multiplexing, i.e. a multiplexed signal that can also multiplex other signals:


> But CAN DBC files also support extended multiplexing, where a multiplexed signal (in this case S1) can also be a multiplexor and thus control the interpretation of other parts of the data payload.


running into this right now while doing UDS stuff for Subaru.

https://www.csselectronics.com/pages/can-dbc-file-database-intro

jnewb1:
@sshane do you have any thoughts on this? probably going to pick this up for the subaru UDS stuff:

> How do you tell the packer which version of the message to build. I see three options:
Specify the multiplexor signal manually. E.g packer.make_can_msg("TemperatureMsg", 0, {'MultiplexIndexSignal': 0, ...})
Add the multiplexor signal as a suffix to the message name. E.g. TemperatureMsg_0, TemperatureMsg_1...
Figure out the correct multiplexor based on the passed signals.

jyoung8607:
Obsoleted by #2539 / #2545

PR #779, opened 2023-02-08
CANParser: add timestamp field

sshane:
Reverts https://github.com/commaai/opendbc/pull/557 and fills out field with logMonoTime of `can` service packet


adeebshihadeh (approved):

sshane:
Took a random Bolt segment, 0.3 seconds to parse before and after

PR #780, opened 2023-02-10
GM: add cluster speed signal

sshane:
Message's rate is very close to what's shown on the Bolt!

Screenshot from 2023-02-09 20-05-14


sshane:
Thanks to @deanlee for this wonderful tool! 
Screenshot from 2023-02-09 20-06-54

PR #793, opened 2023-03-15
VW MLB: Audi A4/S4

jyoung8607:


can/common.cc
```diff
@@ -108,7 +108,7 @@ void init_crc_lookup_tables() {
   gen_crc_lookup_table_16(0x1021, crc16_lut_xmodem);    // CRC-16 XMODEM for HKG CAN FD
 }
 
-unsigned int volkswagen_mqb_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d) {
+unsigned int volkswagen_mxb_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d) {
```
adeebshihadeh, line 111: what do you think of mxb -> mqb_mlb for better grep-ability?

jyoung8607: I'm okay with it. The whole set of abstractions needs another look, was focused on passing tests tonight, will look at it with fresh eyes later.

rusefillc:
@jyoung8607 I am interested in this PR, how is it going?

jyoung8607:
> @jyoung8607 I am interested in this PR, how is it going?

Started to reply, and it turned into a general update which I put in [the openpilot PR](https://github.com/commaai/openpilot/pull/27586#issuecomment-1939400855). Let me know if you have other questions, or if you have a supportable vehicle and would like to help.

rusefillc:
> Let me know if you have other questions, or if you have a supportable vehicle and would like to help.

I do not actually use openpilot at all, my only interest is the nice DBC which is part of this larger change here. Thus https://github.com/commaai/opendbc/pull/1008 as an attempt to help with just the DBC merge.

PR #795, opened 2023-03-26
refactor CANParser to improve performance

deanlee:
1. pass by reference and use vector::emplace_back to reduce memory copies
2. Added new function `CANParser::update_strings` to parse all can events in the strings array, and only call `query_latest` once to update the latest data.  the code in cython also only needs to parse the latest message once. this should greatly improve performance when dealing with multiple can messages.
3. Only expose function `CANParser::update_strings` to cython.


deanlee:
looks like the following code is the performance killer.
```
for v in self.vl_all.values():
      v.clear()

```

deanlee:
It looks like half of the time is spent executing the following statement:

> 
>     for v in self.vl_all.values():
>       v.clear()

deanlee:
the time used for parse all signals in message:

|   | before | after |
| ------------- | ------------- | ------------- |
| update_string  | 687.6ms to parse 50000, avg: 13751ns  | 543.6ms to parse 50000, avg: 10871ns |
| update_strings  | 409.5ms to parse 50000, avg: 8190ns  | 85.0ms to parse 50000, avg: 1699ns |

update_string doesn't improve much because we calls update_strings([m]) in loop to  implement the testing for update_string. this will create a new list every time.

deanlee:
@sshane I have not changed the parameters for `CANParser::CANParser`. you can change it to what you want.

sshane:
We can do that in another PR if this one is good to go standalone!

can/parser_pyx.pyx
```diff
@@ -101,47 +102,34 @@ cdef class CANParser:
       message_options_v.push_back(mpo)
 
     self.can = new cpp_CANParser(bus, dbc_name, message_options_v, signal_options_v)
-    self.update_vl()
+    self.update_strings([])
 
-  cdef unordered_set[uint32_t] update_vl(self):
+  def update_string(self, dat, sendcan=False):
```
adeebshihadeh, line 107: let's get rid of the cython update_string. it's not used anywhere

adeebshihadeh (commented):
Nice speedup and cleanup!

can/parser_pyx.pyx
deanlee: it's still used in can/tests/test_packer_parser.py and can/tests/test_parser.py.

should we replace it with `update_strings` in these two files? or keep update_string as it is?

adeebshihadeh: let's replace it

.github/workflows/tests.yml
```diff
@@ -14,6 +14,10 @@ jobs:
   unit-tests:
     name: unit tests
     runs-on: ubuntu-20.04
+    #strategy:
```
adeebshihadeh, line 17: @deanlee in the future, you can uncomment this to run it several times

deanlee: ok!

PR #801, opened 2023-03-27
CANParser: add performance test

deanlee:
run `can/tests/benchmark.py`:
```
All Signals: 164.019 ms total: 10000 avg: 0.016402
One Signal: 75.689 ms total: 10000 avg: 0.007569


```


can/tests/test_parser.py
```diff
@@ -0,0 +1,60 @@
...
+      ("MINI_CAR", "ACC_CONTROL"),
+      ("CANCEL_REQ", "ACC_CONTROL"),
+      ("ACC_CUT_IN", "ACC_CONTROL"),
+      ("PERMIT_BRAKING", "ACC_CONTROL"),
+      ("RELEASE_STANDSTILL", "ACC_CONTROL"),
+      ("ITS_CONNECT_LEAD", "ACC_CONTROL"),
+      ("ACCEL_CMD_ALT", "ACC_CONTROL"),
+      ("CHECKSUM", "ACC_CONTROL"),
+    ]
+    self._benchmark(signals, [('ACC_CONTROL', 10)], (18000, 20000))
```
adeebshihadeh, line 56: @deanlee I rewrote this as a pass/fail test, but the timings are quite different in CI than my machine. can you come up with a good way to canonicalize this across machine?

deanlee: Can't think of a good way right now:(

adeebshihadeh: let's do this for now:
* increase the test runs until the average time is fairly stable
* set the threshold based on what CI outputs

you can still use the local test output to compare for your PRs and we'll have a bit of CI coverage around it

PR #802, opened 2023-03-27
CANParser: improve cython performance

deanlee:
Clear `defaultdict` and extend it in `update_string` is very expensive. they almost consumes half of cython's running time,

time reduced from ~75-180ms (https://github.com/commaai/opendbc/pull/801)  to 45-130ms after refactor `vl_all ` to normal dict 
```
All Signals: 129.830 ms total: 10000 avg: 0.012983
One Signal: 45.953 ms total: 10000 avg: 0.004595
```


adeebshihadeh:
Looks like CI has confirmed the speedup and test threshold will need to be updated.

deanlee:
the new threshold should be ok

deanlee:
the new threshold should be ok

PR #810, opened 2023-04-07
Added dbcview.py. A quick way to view message bitfields

MoreTore:
Usage example.

`python .\dbcview.py mazda_2017.dbc ENGINE_DATA`

or 

`python .\dbcview.py mazda_2017.dbc 514`

image

### DIFF
--- a/dbcview.py
+++ b/dbcview.py
@@ -1,0 +1,86 @@
+import argparse
+import cantools
+import matplotlib.pyplot as plt
+import numpy as np
+import mplcursors
+
+def visualize_bitfield_8x8(message):
+    fig, ax = plt.subplots()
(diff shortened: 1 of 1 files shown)


sshane:
Is there a use-case loading up desktop Cabana with a dummy route doesn't cover?

MoreTore:
No cabana is better for deep RE. But, if you just want to visualize the bits in a field when modifying panda safety code it's quick and easy. 

If cabana could be installed on windows easily, cabana would be better. Also I don't think you can view bit fields unless you have a drive where it is recieved. 

Anyways I got ChatGPT to make this is 10 seconds and I thought it might be useful.

PR #815, opened 2023-04-15
Mazda:  update BSM signals and add comments

Jafaral:


mazda_2017.dbc
```diff
@@ -770,6 +772,10 @@ CM_ SG_ 552 GEAR_BOX "0 P, 14 R, 1 though 6 D for speeds, 15 Shift";
 CM_ SG_ 540 HANDS_ON_STEER_WARN "0 no warning, b warning";
 CM_ SG_ 540 DISTANCE_SETTING "Radar distance 0: disabled, 1: 4 bars, 2: 3 bars, 3: 2 bars, 4: 1 bar";
 CM_ SG_ 1143 REAR_CT_ALERT "Rear Cross Traffic Alert";
+CM_ SG_ 1143 LEFT_BS1 "Left BlindSpot without blinker";
+CM_ SG_ 1143 RIGHT_BS1 "Right BlindSpot without blinker";
+CM_ SG_ 1143 LEFT_BS2 "Left Blindspot with left blinker activated - warning activated";
+CM_ SG_ 1143 RIGHT_BS2 "Right Blindspot with right blinker activated - warning activated";
```
sshane, line 778:
```suggestion
CM_ SG_ 1143 LEFT_BS_BLINKER "Left Blindspot with left blinker activated - warning activated";
CM_ SG_ 1143 RIGHT_BS_BLINKER "Right Blindspot with right blinker activated - warning activated";
```

seems like more appropriate names to me. would it make sense to combine into two 2 bit signals though since they are adjacent?

sshane:
Can you provide a route I can check with both signals activating? @Jafaral

mazda_2017.dbc
Jafaral: We can, though I don't see any advantage either way

Jafaral:
hardly driving more than a few minutes in the next few days and mostly 2-way streets, then traveling for a few days. I have been driving with that fix for a few months though, and I can confirm that it works reliably.

adeebshihadeh:
We'll need a route to confirm this. Perhaps someone from the community can provide one? Doesn't have to be recent.

sshane:
I'll check our data

Jafaral:
8c521516df3527d8|2023-05-09--08-32-12--9

Jafaral:
I got the warning to trigger a couple of times too.

mazda_2017.dbc
sshane: Makes sense to combine them, it's likely just a 2 bit signal in the original DBC and we can check != 0 instead of ORing the two bits

image

sshane:
Checked last 180 days:

```python
Routes: 4177
Dongles: 31
Segments for platform: {'MAZDA CX-5 2022': 3381, 'MAZDA CX-9': 2, 'MAZDA CX-9 2021': 741, 'MAZDA CX-5': 53}
```

if the new bit was 1, then that blinker was always on

PR #824, opened 2023-04-19
VW PQ: added AWV (AEB) signals

FLcruising:
Signals useful for AEB were missing.


jyoung8607:
You'll need to bump the message length from 5 to 8, to accommodate these newer signals.

If we're to make proper use of these later, it would help to include the VAL tables. Can you do that as well?

FLcruising:
Ah yes. Thanks for catching the message length.
Please explain what a VAL table is.

> You'll need to bump the message length from 5 to 8, to accommodate these newer signals.
> 
> If we're to make proper use of these later, it would help to include the VAL tables. Can you do that as well?

jyoung8607:
Toward the bottom of the DBC, you'll see existing VAL entries like this:

`VAL_ 872 ACS_Sta_ADR 2 "ADR_passiv" 0 "ADR_nicht_aktiv" 1 "ADR_aktiv" 3 "irrev_Fehler" ;`

Add the corresponding ones for all the AWV signals. As an example,

`VAL_ 870 AWV_2_SU_Warnzeit 0 "frueh" 1 "normal" 2 "spaet" 3 "adaptiv" ;`

This will be useful for fields like `AWV_Text` where we may want to decode or even rewrite what's being displayed/sent. I'm pretty sure you have the data to put into these VALs, DM me if you don't.

FLcruising:
I see...they're the signal description. Yes I have them but have translated to English. Do you care?

jyoung8607:
I prefer to keep them in their original form, at least for now. There's a valid argument to be made for translating things, but we trade off easy reconciliation with other signal names and comments, reference docs, vehicle DTCs, etc etc.

FLcruising:
Got it. Let me work on that.

vw_golf_mk4.dbc
```diff
@@ -1169,27 +1169,38 @@ BO_ 644 Motor_Bremse: 6 XXX
...
- SG_ AWV_2_Fehler : 25|1@1+ (1,0) [0|1] ""  Gateway_Kbi_VW_A,Bremsbooster
- SG_ AWV_2_Status : 24|1@1+ (1,0) [0|1] ""  Gateway_Kbi_VW_A,Bremsbooster
- SG_ AWV_res_20 : 20|4@1+ (1,0) [0|0] ""  Bremsbooster
- SG_ AWV_1_Parameter : 18|2@1+ (1,0) [0|3] ""  Bremsbooster
- SG_ AWV_1_Prefill : 17|1@1+ (1,0) [0|1] ""  Bremsbooster
- SG_ AWV_1_Freigabe : 16|1@1+ (1,0) [0|1] ""  Bremsbooster
- SG_ AWV_Text : 12|4@1+ (1,0) [0|15] ""  Gateway_Kbi_VW_A,Bremsbooster
- SG_ AWV_Zaehler : 8|4@1+ (1,0) [0|15] ""  Bremsbooster
- SG_ AWV_Checksumme : 0|8@1+ (1,0) [0|255] ""  Bremsbooster
+BO_ 870 mAWV: 8 XXX
```
jyoung8607, line 1172: The one thing we don't retain, from this particular data source, is the `m` prefix on message names.

```suggestion
BO_ 870 AWV: 8 XXX
```

jyoung8607 (changes requested):
Since we're overhauling this whole message, can we go ahead and sort these signals by their bit offset order? `CHECKSUM` first, `ANB_CM_Anforderung` last, etc. I try to clean these up whenever I'm working on a given message.

jyoung8607 (approved):
LGTM. openpilot [doesn't currently look](https://github.com/commaai/openpilot/blob/ee15d6b1d1b2f90815d0cbf3c0f6b5adfdb0a59f/selfdrive/car/volkswagen/carstate.py#L222-L224) at the FCW/AEB signals for VW PQ, so this can be merged anytime. I'd like to do so in the future, but there will be nuances with older cars to worry about.

sshane:
Sorry for the delay, @jyoung8607 this still GTM?

jyoung8607 (approved):

PR #828, opened 2023-04-23
CANParser: parse all signals for given messages

adeebshihadeh:


sshane (approved):
Will get OP PR ready! Want to get this in before we do any more car ports (like the Palisades)

PR #835, opened 2023-05-07
EV6 dbc updates

dibble9012:
Adds more signals and comments for doors/seatbelts/drive mode/local time/HVAC.


adeebshihadeh:
Thanks! Next time, make sure to make the matching openpilot PR like this https://github.com/commaai/openpilot/pull/28322

PR #844, opened 2023-05-29
Hyundai SantaFe CM and Opel Omega B databases added

MonsterPerformance:
Hyundai SantaFe 2007 2.7V6 4AT:
Now the signals have understandable names and all ones have been checked on my car.
Useless signals are removed at all.

Opel Omega B 2001 3.2V6 5MT:
New database for GM's Vectra B/Omega B with V6 engines.


MonsterPerformance:
@adeebshihadeh could you please review and merge?

PR #847, opened 2023-06-03
Subaru: DBC Update for gen2 long

jnewb1:
see https://github.com/commaai/openpilot/pull/28391

Cruise_Active seems to just so happen to be correct, because of how the bits line up in the set speed. Since it doesn't appear to be used anywhere, I removed it.


jnewb1:
Cruise_Set_Speed:
image

PR #855, opened 2023-06-12
VW MQB Kombi Reverse Engineering

V8Joel:
Various improvements and bug fixes to VW MQB DBC file

1. DBC file repaired such that it can now be opened in Vector tools such as CANalyzer and CANdb++ - Previously unavailable due to artefacts that caused errors when trying to open the DBC in Vector tools
2. Messages proven to be emanating from the Kombi meter were re-assigned so that they had the correct transmitter node
3. New English variant of MQB DBC with various translations - Still lots more to do


rusefillc:
@sshane what's the process for this PR to move forward?

jyoung8607:
Closing this PR due to age, conflicts, and too many changes to effectively review.

@V8Joel We would be open to a standalone PR that makes the minimum set of repairs necessary to harmonize this DBC with the Vector standard format, and for CANdb++ to be able to open the file. I know that's a problem right now, and it should be fixed. I would also be open to fixing other known issues, like the transmit node for a given message, but they should be done in separate PRs.

The opendbc project would not merge the Vector tool `.ini` file, nor would it merge a hand-translated duplicate DBC. However, we would definitely review English `CM_` comments on both messages and signals, whether those are new or translated from the existing German. For a variety of reasons, we will not change the actual message and signal names themselves.

PR #856, opened 2023-06-14
Toyota: fix and add some PCM signals

sshane:


generator/toyota/_toyota_2017.dbc
```diff
@@ -75,8 +75,8 @@ BO_ 452 ENGINE_RPM: 8 CGW
 BO_ 466 PCM_CRUISE: 8 XXX
  SG_ GAS_RELEASED : 4|1@0+ (1,0) [0|1] "" XXX
  SG_ CRUISE_ACTIVE : 5|1@0+ (1,0) [0|1] "" XXX
- SG_ STANDSTILL_ON : 12|1@0+ (1,0) [0|1] "" XXX
- SG_ ACCEL_NET : 23|16@0- (0.001,0) [-20|20] "m/s^2" XXX
+ SG_ ACC_BRAKING : 12|1@0+ (1,0) [0|1] "" XXX
```
sshane, line 78: this is actually ` SG_ ACCSTBK : 12|3@0+ (1,0) [0|0] "" Vector__XXX` but I haven't seen the other bits change (except bit 10, that's the inverse of this bit), even with an ACC fault. can check data if we ever need to

generator/toyota/_toyota_2017.dbc
```diff
@@ -75,8 +75,9 @@ BO_ 452 ENGINE_RPM: 8 CGW
...
+ SG_ ACCEL_NET : 23|16@0- (0.0009765625,0) [-20|20] "m/s^2" XXX
+ SG_ NEUTRAL_FORCE : 39|16@0- (2,0) [-65536|65534] "N" XXX
```
sshane, line 80: this one is `SG_ ACCAVL : 39|16@0- (2,0) [0|0] "N" Vector__XXX`. seems to change proportionally with the residual force the engine or electric motors are applying. when going up a steep hill, the signal changes from its usual 1400 near a stop to over 4000 and you could feel the motors really trying to keep you from rolling back without any pedal input

image

PR #861, opened 2023-06-20
Subaru: Buttons

jnewb1:
see https://github.com/commaai/openpilot/pull/28588


adeebshihadeh:
I read the matching PR, but still not sure why this change is necessary.

jnewb1:
> I read the matching PR, but still not sure why this change is necessary.

shane said other car ports use a single field for the buttons and a dict mapping between the button value and button type
https://github.com/commaai/openpilot/pull/28588#discussion_r1233186732

PR #874, opened 2023-07-03
Honda: new EPS faulted value

vanillagorillaa:
Not sure if this value is also the same on other Honda's, left on the civic for now

### DIFF
--- a/can/tests/test_define.py
+++ b/can/tests/test_define.py
@@ -14,7 +14,7 @@ class TestCADNDefine(unittest.TestCase):
     self.assertDictEqual(defs.dv[399],
                          {'STEER_STATUS':
                           {7: 'PERMANENT_FAULT',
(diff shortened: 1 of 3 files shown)


can/tests/test_define.py
```diff
@@ -13,7 +13,8 @@ def test_civic(self):
     self.assertDictEqual(defs.dv[399], defs.dv['STEER_STATUS'])
     self.assertDictEqual(defs.dv[399],
                          {'STEER_STATUS':
-                          {6: 'TMP_FAULT',
+                          {7: 'PERMANENT_FAULT',
+			   6: 'TMP_FAULT',
```
incognitojam:
```suggestion
                           6: 'TMP_FAULT',
```

sshane:
For https://github.com/commaai/openpilot/issues/28347

PR #876, opened 2023-07-05
Toyota: add comment for LTA signal

sshane:
If TSS/LSS 2.5, this signal is always 1.0. If TSS 2.0, it's usually 3, but can be 0 with these exceptions:

- Highlander & Highlander Hybrid is 0 instead of 3 if TSS 2.0 (checked a few VINs for model years and cross referenced with TSS applicability chart)
- Lexus NX is also 0 if TSS 2.0 (LSS+ 3.0 is 2022 MY, have only seen 2021 MY)
- Alphard is also 0 (not US market car)

There's also one exception to the rule of all TSS 2.0 cars having this signal at either 0 or 3:
- RAV4 & RAV4 Hybrid 2022 (non-LTA, TSS 2.0) have the signal at 1. Interestingly, they also are the only cars we've seen use LKA_ACTIVE in the LTA signal, but still accept LKA.


sshane:
So this doesn't describe if the car uses the LTA message for LKA, or even purely TSS 2.5 since the 2022 RAV4s are 1. But without injection testing on more cars, my best guess is the max torque that the EPS allows using LTA.

Cars with 3 or 0 (TSS 2.0) allow 1500 torque units, like a few Corolla's I've seen, and cars with 1 (mostly TSS 2.5) allow the higher limit of 3500 units.

Issue #882, opened 2023-07-06
CAN Packer: undefined behavior when not setting a signal


sshane:
While looking at https://github.com/commaai/openpilot/pull/27785 I noticed that the CAN packer takes on different behavior depending on if you define a signal to 0, or leave it out of the dictionary. if you define it to 0, it will take the correct offset and scaling into account, but if you don't it will just set the bytes to `0x00`. This can lead to unintended behavior as you expect them to produce the same behavior

reproduction:

```python
from opendbc.can.packer import CANPacker
packer = CANPacker('hyundai_canfd')
packer.make_can_msg('SCC_CONTROL', 0, {'VSetDis': 0})[2][16]
Out[59]: 0
packer.make_can_msg('SCC_CONTROL', 0, {'VSetDis': 0, 'aReqValue': 0})[2][16]
Out[61]: 255
```

openpilot issue: https://github.com/commaai/openpilot/issues/28822

sshane:
@deanlee what do you think is better, in cython adding all signals with 0 value defaults, or handle in packer.cc?

deanlee:
It might be better to implement it inside packer.cc.

PR #884, opened 2023-07-06
CANPacker: refactor to avoid undefined signals and set default value for  the unspecified signals.

deanlee:
@sshane any thoughts on this implementation?

resolve https://github.com/commaai/opendbc/issues/500 https://github.com/commaai/opendbc/issues/882

### DIFF
--- a/can/packer.cc
+++ b/can/packer.cc
@@ -40,16 +40,24 @@ CANPacker::CANPacker(const std::string& dbc_name) {
   init_crc_lookup_tables();
 }
 
+// set all values for all given signal/value pairs
 std::vector<uint8_t> CANPacker::pack(uint32_t address, const std::vector<SignalPackValue> &signals) {
+  auto msg_it = message_lookup.find(address);
+  if (msg_it == message_lookup.end()) {
+    throw std::runtime_error("CanPacker::pack(): invalid address " + std::to_string(address));
+  }
+
+  for (const auto &sig: signals) {
+    if (signal_lookup.find({address, sig.name}) == signal_lookup.end()) {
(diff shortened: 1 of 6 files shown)


can/packer.cc
```diff
@@ -45,24 +45,23 @@ std::vector<uint8_t> CANPacker::pack(uint32_t address, const std::vector<SignalP
 
   // set all values for all given signal/value pairs
   bool counter_set = false;
-  for (const auto& sigval : signals) {
-    auto sig_it = signal_lookup.find(std::make_pair(address, sigval.name));
-    if (sig_it == signal_lookup.end()) {
-      // TODO: do something more here. invalid flag like CANParser?
-      WARN("undefined signal %s - %d\n", sigval.name.c_str(), address);
-      continue;
```
sshane, line 53: we should make sure to preserve the warning for specifying non-existent signals, or better yet, make it apparent you're doing something weird: https://github.com/commaai/opendbc/issues/500

sshane:
anything you can think of to split into separate changes/PRs?

sshane:
We can't merge this just yet, as is, since certain places in openpilot rely on this behavior. For example, openpilot zeros the bytes for the gas interceptor command, which when you set the values to 0, you get non-zero bytes

image

Previous parsed values (supposed to be 0):

{'COUNTER_PEDAL': 9.0,
 'CHECKSUM_PEDAL': 132.0,
 'GAS_COMMAND': -75.555,
 'GAS_COMMAND2': -151.111}

And now they're

{'COUNTER_PEDAL': 9.0,
 'CHECKSUM_PEDAL': 86.0,
 'GAS_COMMAND': -0.01125000000001819,
 'GAS_COMMAND2': -0.023500000000012733}

deanlee:
I see. closed it for now

PR #891, opened 2023-07-07
CANPacker: refactor to avoid undefined signals

deanlee:
split from https://github.com/commaai/opendbc/pull/884
resolve https://github.com/commaai/opendbc/issues/500


sshane:
thanks!

PR #893, opened 2023-07-13
Subaru: more steering messages

jnewb1:
Couple new steering messages


sshane:
so the same cars with the different LKAS message id also use these?

jnewb1:
> so the same cars with the different LKAS message id also use these?
> 
> 

Unknown, just found those while looking for a steer rate message

sshane:
oh these are unrelated to the other Subaru stuff, got it. if these are on a few cars, feel free to merge

sshane (approved):

PR #896, opened 2023-07-14
CANPacker: assert signals w/ non-zero offsets be explicitly set

sshane:
If the signal has a non-zero offset, require it to be explicitly set. Leaving the bytes at zero for any other signal doesn't change the parsed value, but it does for offset signals

### DIFF
--- a/can/common.h
+++ b/can/common.h
@@ -96,10 +96,11 @@ private:
   std::map<uint32_t, Msg> message_lookup;
   std::map<std::string, uint32_t> message_name_to_address;
   std::map<uint32_t, uint32_t> counters;
+  bool enforce_checks = false;
 
 public:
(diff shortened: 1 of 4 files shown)


sshane:
added `enforce_checks` flag to mitigate this. there's not many signals with offsets we need to handle, so this should be fine as a user of CANPacker

adeebshihadeh:
> @adeebshihadeh do you think we should change the CANPacker behavior to change all these signals to 0 (#884), or assert that all signals with offsets be set? I could see the latter make the CANPacker more annoying to use, but it seems better to me. Maybe some flag to only check in CI?

I think the current behavior makes the most sense. Requiring some signals seems arbitrary.

sshane:
Going to add test cases for each type of message openpilot sends to test_models instead

PR #898, opened 2023-07-18
CI: Fix panda CI by not upgrading Cython

jnewb1:
Cython 3.0.0 breaks opendbc, but now it wants to upgrade by default in the CI. Force to <3.0.0

### DIFF
--- a/can/packer_pyx.pyx
+++ b/can/packer_pyx.pyx
@@ -29,7 +29,7 @@ cdef class CANPacker:
 
     return self.packer.pack(addr, values_thing)
 
-  cpdef make_can_msg(self, name_or_addr, bus, values) except +RuntimeError:
+  cpdef make_can_msg(self, name_or_addr, bus, values):
     cdef uint32_t addr
     if type(name_or_addr) == int:
       addr = name_or_addr
--- a/can/packer_pyx.pyx
+++ b/can/packer_pyx.pyx
@@ -29,7 +29,7 @@ cdef class CANPacker:
 
     return self.packer.pack(addr, values_thing)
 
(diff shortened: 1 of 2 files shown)


jnewb1:
https://github.com/commaai/panda/pull/1520

sshane:
Found this PR which added it: https://github.com/cython/cython/pull/4013 Seems like a good warning to me, can we revert this? The real fix is removing the RuntimeError that has ambiguous behavior (and is being reverted)

jnewb1:
> Found this PR which added it: [cython/cython#4013](https://github.com/cython/cython/pull/4013) Seems like a good warning to me, can we revert this? The real fix is removing the RuntimeError conversion that has ambiguous behavior (and is being reverted)

yea we can revert it. it would be nice if we had a better way to sync dependencies between openpilot and opendbc for ci

jnewb1:
don't think we can revert this yet due to some weird paren choices in 3.0.0 __pyx_array_new (causes compile warning -> error)

visionipc/visionipc_pyx.cpp:9332:28: error: equality comparison with extraneous parentheses [-Werror,-Wparentheses-equality]
  if (((__pyx_v_c_mode[0]) == 'f')) {
       ~~~~~~~~~~~~~~~~~~~~^~~~~~
visionipc/visionipc_pyx.cpp:9332:28: note: remove extraneous parentheses around the comparison to silence this warning
  if (((__pyx_v_c_mode[0]) == 'f')) {
      ~                    ^     ~
visionipc/visionipc_pyx.cpp:9332:28: note: use '=' to turn this equality comparison into an assignment
  if (((__pyx_v_c_mode[0]) == 'f')) {
                           ^~
                           =
1 error generated.

PR #901, opened 2023-07-18
Add Subaru driver torque sensor Nm scaling factor and unit

martinl:
This PR adds Nm unit for Subaru driver torque sensor signal, matched with SSM4 signal output by to jnewb1 at
https://github.com/martinl/openpilot/issues/80

I will update openpilot carstate when this PR passes initial review


adeebshihadeh:
Seems fine, want to update the interface?

martinl:
I will open new PR due to complicated git history

martinl:
New PR https://github.com/commaai/opendbc/pull/1198

PR #912, opened 2023-08-08
parser: update_strings propagate cpp exceptions in py bindings

fredyshox:
Right now, when `update_strings` is called with some corrupted/arbitrary bytes, cpp exception is thrown which crashes the process. With this, exception will be propagated as `RuntimeError`

### DIFF
--- a/can/tests/test_dbc_exceptions.py
+++ b/can/tests/test_dbc_exceptions.py
@@ -28,6 +28,10 @@ class TestCanParserPackerExceptions(unittest.TestCase):
     with self.assertRaises(RuntimeError):
       CANParser(dbc_file, signals, [], 0)
 
+    with self.assertRaises(RuntimeError):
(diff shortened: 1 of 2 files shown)


sshane:
Can you add a failing test to the exceptions file?

can/tests/test_dbc_exceptions.py
```diff
@@ -28,6 +28,10 @@ def test_civic_exceptions(self):
     with self.assertRaises(RuntimeError):
       CANParser(dbc_file, signals, [], 0)
 
+    with self.assertRaises(RuntimeError):
+      parser = CANParser(dbc_file, signals, checks, 0)
+      parser.update_strings([b''])
```
sshane, line 33:
```suggestion
    parser = CANParser(dbc_file, signals, checks, 0)
    with self.assertRaises(RuntimeError):
      parser.update_strings([b''])
```

PR #927, opened 2023-08-26
Hyundai CAN-FD HDA2: specify signals to suppress LFA

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/openpilot/pull/29658

image

### DIFF
--- a/hyundai_canfd.dbc
+++ b/hyundai_canfd.dbc
@@ -470,7 +470,10 @@ BO_ 676 CAM_0x2a4: 24 XXX
  SG_ BYTE4 : 32|8@1+ (1,0) [0|255] "" XXX
  SG_ BYTE5 : 40|8@1+ (1,0) [0|255] "" XXX
  SG_ BYTE6 : 48|8@1+ (1,0) [0|255] "" XXX
- SG_ BYTE7 : 56|8@1+ (1,0) [0|255] "" XXX
+ SG_ LEFT_LANELINES : 56|2@1+ (1,0) [0|3] "" XXX
+ SG_ NEW_SIGNAL_1 : 58|2@1+ (1,0) [0|3] "" XXX
+ SG_ RIGHT_LANELINES : 60|2@1+ (1,0) [0|3] "" XXX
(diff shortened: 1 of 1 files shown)


hyundai_canfd.dbc
```diff
@@ -470,7 +470,10 @@ BO_ 676 CAM_0x2a4: 24 XXX
  SG_ BYTE4 : 32|8@1+ (1,0) [0|255] "" XXX
  SG_ BYTE5 : 40|8@1+ (1,0) [0|255] "" XXX
  SG_ BYTE6 : 48|8@1+ (1,0) [0|255] "" XXX
- SG_ BYTE7 : 56|8@1+ (1,0) [0|255] "" XXX
+ SG_ LEFT_LANELINES : 56|2@1+ (1,0) [0|3] "" XXX
+ SG_ NEW_SIGNAL_1 : 58|2@1+ (1,0) [0|3] "" XXX
+ SG_ RIGHT_LANELINES : 60|2@1+ (1,0) [0|3] "" XXX
+ SG_ NEW_SIGNAL_2 : 62|2@1+ (1,0) [0|3] "" XXX
```
sshane, line 476: Since the Lane Safety setting in vehicle settings disappears, there must be something other than "no lanelines" we're sending, unless no lane lines is not 0, and 0 means something else. Some comments of what you've discovered would go a long way!

sunnyhaibin: We noticed that on the Ioniq 6 HDA2, the Lane Safety setting still disappeared even when we didn't block the message to suppress LFA. We confirmed this with and without blocking this message. I suspect the EV6 HDA2 might have the same behavior, so the Lane Safety setting is probably affected by something other than this message.

The signals seem to change on the level of certainty with left/right lanelines. I haven't seen any noticeable pattern with it at the moment. As long as one of the signals is at least `1`, LFA would allow to engage. I will write some comments with my findings. See example below:
image

sshane:
Looks like it matches our EV6! Found some signals that seems to describe some aspect of the left and right lane lines.

image

Issue #914, opened 2023-08-09
CANParser: assert no duplicate checks are passed in


sshane:
Found we add Dashlights multiple times with different frequencies for Subaru:

https://github.com/commaai/openpilot/blob/cf64b27168614f851087adab2d7f766425922f64/selfdrive/car/subaru/carstate.py#L197

PR #930, opened 2023-08-29
Parser: disallow duplicate message checks

jnewb1:
runtime error if message is checked twice

resolves: https://github.com/commaai/opendbc/issues/914

### DIFF
--- a/can/parser.cc
+++ b/can/parser.cc
@@ -2,6 +2,8 @@
 #include <cassert>
 #include <cstring>
 #include <limits>
+#include <stdexcept>
+#include <sstream>
 
 #include <unistd.h>
 #include <fcntl.h>
@@ -98,6 +100,12 @@ CANParser::CANParser(int abus, const std::string& dbc_name, const std::vector<st
   bus_timeout_threshold = std::numeric_limits<uint64_t>::max();
 
   for (const auto& [address, frequency] : messages) {
+    if(message_states.find(address) != message_states.end()) { // disallow duplicate message checks
+      std::stringstream is;
+      is << "Duplicate Message Check: " << address;
+      throw std::runtime_error(is.str());
+    }
+
     MessageState &state = message_states[address];
     state.address = address;
     // state.check_frequency = op.check_frequency,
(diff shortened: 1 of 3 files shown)


can/parser.cc
```diff
@@ -98,6 +100,13 @@ CANParser::CANParser(int abus, const std::string& dbc_name, const std::vector<st
   bus_timeout_threshold = std::numeric_limits<uint64_t>::max();
 
   for (const auto& [address, frequency] : messages) {
+    // disallow duplicate message checks
+    if (message_states.find(address) != message_states.end()) { 
+      std::stringstream is;
+      is << "Duplicate Message Check: " << address;
+      throw std::runtime_error(is.str());
+    }
+
```
sshane, line 109: might make more sense in the cython wrapper?

jnewb1: wouldn't we want to enfore this in C too? I can move it if you think its better in Python

adeebshihadeh: this is the right place. the wrapper does too much

can/tests/test_packer_parser.py
```diff
@@ -309,6 +309,15 @@ def test_track_all_signals(self):
       "CHECKSUM": 0,
     })
 
+  def test_disallow_duplicate_messages(self):
+    CANParser("toyota_nodsu_pt_generated", [("ACC_CONTROL", 5)])
+
+    with self.assertRaises(RuntimeError):
+      CANParser("toyota_nodsu_pt_generated", [("ACC_CONTROL", 5), ("ACC_CONTROL", 10)])
```
adeebshihadeh, line 316: tests should ideally use the test dbc in can/tests/

can/parser.cc
```diff
@@ -98,6 +100,13 @@ CANParser::CANParser(int abus, const std::string& dbc_name, const std::vector<st
   bus_timeout_threshold = std::numeric_limits<uint64_t>::max();
 
   for (const auto& [address, frequency] : messages) {
+    // disallow duplicate message checks
+    if (message_states.find(address) != message_states.end()) { 
+      std::stringstream is;
+      is << "Duplicate Message Check: " << address;
+      throw std::runtime_error(is.str());
+    }
```
sshane, line 108: I think we can make `DBC_ASSERT` common, from dbc.cc

PR #933, opened 2023-09-04
Hyundai CAN: gear shifter and accelerator pedal signals for FCEV

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/openpilot/pull/29771


opendbc/car/hyundai/values.py
```diff
@@ -745,6 +752,8 @@ def match_fw_to_car_fuzzy(live_fw_versions, vin, offline_fw_versions) -> set[str
 
 EV_CAR = CAR.with_flags(HyundaiFlags.EV)
 
+FCEV_CAR = CAR.with_flags(HyundaiFlags.FCEV)
```
jyoung8607, line 755: This is redundant, it can be removed along with the other places it appears.

Think we can probably ditch `EV_CAR` too, but haven't fully checked, and that's a separate PR anyway.

jyoung8607 (changes requested):

sunnyhaibin:
Awaiting for test route from user `simon402_61863 (Discord)`.

Issue #825, opened 2023-04-21
Remove dependence on cereal


adeebshihadeh:
Replace the CANParser's inputs that use `cereal::CanData`; will need to be done carefully to avoid a large hit to performance.

adeebshihadeh:
@deanlee want to look into this? I'm thinking we make or use an existing struct from opendbc that represents a CAN message, then we write a fast cython function in openpilot to go from cereal's CAN struct to opendbc's struct

deanlee:
sure,working on this

PR #934, opened 2023-09-05
CANParser: remove dependence on cereal

deanlee:
resolve https://github.com/commaai/opendbc/issues/825


can/common.h
```diff
@@ -58,34 +51,34 @@ class MessageState {
...
   uint64_t first_sec = 0;
-  uint64_t last_sec = 0;
   uint64_t last_nonempty_sec = 0;
   uint64_t bus_timeout_threshold = 0;
   uint64_t can_invalid_cnt = CAN_INVALID_CNT;
 
   CANParser(int abus, const std::string& dbc_name,
             const std::vector<std::pair<uint32_t, int>> &messages);
   CANParser(int abus, const std::string& dbc_name, bool ignore_checksum, bool ignore_counter);
-  #ifndef DYNAMIC_CAPNP
```
adeebshihadeh, line 81: note that we'll also need to update PlotJuggler after this change

can/common.h
```diff
@@ -58,34 +51,34 @@ class MessageState {
...
-  void update_string(const std::string &data, bool sendcan);
-  void update_strings(const std::vector<std::string> &data, std::vector<SignalValue> &vals, bool sendcan);
-  void UpdateCans(uint64_t sec, const capnp::List<cereal::CanData>::Reader& cans);
-  #endif
-  void UpdateCans(uint64_t sec, const capnp::DynamicStruct::Reader& cans);
+  void update_frames(uint64_t frame_fist_sec, const std::vector<CanFrame> &frames, std::vector<SignalValue> &vals);
+  bool updateFrame(const CanFrame &data);
```
adeebshihadeh, line 79: make sure to keep the style consistent

deanlee:
So we still need to support parsing dynamic capnp in parser.cc if the  DYNAMIC_CAPNP is defined?

Also parser_pyx.pyx still depends on cereal, do we need to move it to openpiot? Or just leave it as it is?

adeebshihadeh:
The goal of this is to make opendbc a real independent submodule,, and that means getting rid of the cereal dependence entirely.

I'm  imagining we have a very fast  function that converts from cereal's CAN struct to opendbc's. Then, we can wrap it in Cython for openpilot and PlotJuggler can also use it.

deanlee:
got it

adeebshihadeh:
@deanlee want to finish this one up?

deanlee:
y, but it might take a little while.

deanlee:
@adeebshihadeh : Done.  there is no loss in performance.
PlotJuggler  needs to be modified after this to use the new api.

can/strings_to_candata.cc
```diff
@@ -0,0 +1,37 @@
+#include <capnp/serialize.h>
+
+#include <string>
+#include <vector>
+
+#include "cereal/gen/cpp/log.capnp.h"
+#include "opendbc/can/common.h"
+
```
adeebshihadeh, line 8: I think we want this to go in openpliot. opendbc should be completely independent of cereal, i.e. `grep cereal` should have no hits in opendbc/

deanlee:
@adeebshihadeh removed all dependencies on cereal.  the CAN frame uses the same format as `CanPacker`: [address, 0, dat, source]

sshane:
Is this ready to review?

sshane:
@deanlee let me know when this is ready for review

deanlee:
Sure, but it might take a few more days.

deanlee:
@sshane The extra conversion steps make it (https://github.com/commaai/openpilot/pull/32009) more than twice as slow as the master version.
### this branch

> 6000 CAN packets, 10 runs
> 263.47 mean ms, 367.93 max ms, 258.64 min ms, 37.38 std ms
> 0.0506 mean ms / CAN packet
> 

### master

> 
> 6000 CAN packets, 10 runs
> 125.61 mean ms, 130.85 max ms, 118.84 min ms, 2.78 std ms
> 0.0209 mean ms / CAN packet

sshane:
Seeing similar results here. I think this is acceptable, but would be nice to speed it up in the future. Before https://github.com/commaai/opendbc/pull/1039 I'm seeing

```
6000 CAN packets, 10 runs
295.08 mean ms, 303.76 max ms, 292.93 min ms, 2.99 std ms
0.0492 mean ms / CAN packet
```

on this PR I'm seeing

```
6000 CAN packets, 10 runs
214.52 mean ms, 216.95 max ms, 212.54 min ms, 1.13 std ms
0.0358 mean ms / CAN packet
```

and on current master I'm seeing 

```
6000 CAN packets, 10 runs
97.72 mean ms, 98.73 max ms, 96.72 min ms, 0.70 std ms
0.0163 mean ms / CAN packet
```

sshane:
@deanlee is this ready now? Once it's in we can really start on https://github.com/commaai/openpilot/pull/32630!

sshane:
BTW here's a list of things we need to remove from selfdrive/car, do you know if we have Python logging in opendbc yet? https://github.com/commaai/openpilot/pull/32726

deanlee:
Y, it's ready.

> BTW here's a list of things we need to remove from selfdrive/car, do you know if we have Python logging in opendbc yet? https://github.com/commaai/openpilot/pull/32726BTW here's a list of things we need to remove from selfdrive/car, do you know if we have Python logging in opendbc yet? https://github.com/commaai/openpilot/pull/32726BTW here's a list of things we need to remove from selfdrive/car, do you know if we have Python logging in opendbc yet? https://github.com/commaai/openpilot/pull/32726

There shouldn't be any, but I'm not very sure right now

can/parser.cc
```diff
@@ -157,64 +157,40 @@ CANParser::CANParser(int abus, const std::string& dbc_name, bool ignore_checksum
...
-    update_string(d, sendcan);
+  for (const auto &c : can_data) {
+    if (first_nanos == 0) {
+      first_nanos = c.nanos;
+    }
     if (current_nanos == 0) {
-      current_nanos = last_nanos;
+      current_nanos = c.nanos;
     }
+    last_nanos = c.nanos;
```
sshane, line 169: So usually we won't receive multiple items inside `can_data`, but if card/controlsd lags and gets two CAN packets from `drain_sock_raw`, it will. Then `current_nanos` will be the first message and not the second/latest. I like that this refactor is making it much easier to reason about this logic! It was much more complicated before

sshane:
LGTM, thanks for the PR!

sshane:
@deanlee want to update PlotJuggler?

deanlee:
I keep encountering the following error when trying to clone the PlotJuggler:

> Cloning into 'PlotJuggler'...
> remote: Enumerating objects: 18341, done.
> remote: Counting objects: 100% (4032/4032), done.
> remote: Compressing objects: 100% (428/428), done.
> error: RPC failed; curl 56 GnuTLS recv error (-9): Error decoding the received TLS packet.
> fatal: the remote end hung up unexpectedly
> fatal: early EOF
> fatal: index-pack failed
> 

I'll attempt it again later.

PR #935, opened 2023-09-05
Do not copy Msg in __init__

deanlee:
before:
`struct Msg __pyx_v_msg = ((__pyx_v_self->dbc[0]).msgs[__pyx_v_i]);`

after:

struct Msg const *__pyx_v_msg = (&(__pyx_v_self->dbc->msgs[__pyx_v_i]));


adeebshihadeh:
Does this make a noticeable difference? I think the real thing we want to do here is make the Cython do much less and only thinly wrap the C++.

deanlee:
Yes, you are right, such optimization does not make sense at the moment

Issue #902, opened 2023-07-18
CANPacker: some exceptions are silent


sshane:
For example, since we don't propagate exceptions from this function, if you pass certain inputs to make_can_msg, you'll get an exception that Python never knows about

https://github.com/commaai/opendbc/blob/3ef35ed2298a3a9d199f9145409547710065884c/can/packer_pyx.pyx#L29-L39

sshane:
@adeebshihadeh I think this is worth raising an exception for, rather than adding to the valid flag we want to add. What do you think?

adeebshihadeh:
No more Cython!

PR #942, opened 2023-09-06
build message lookup table in DBC

deanlee:
Both the Cython and C++ code currently build their own message lookup tables independently each time they are run. This redundancy leads to code duplication and decreased initialization performance. To improve efficiency, it would be better to create the message lookup table once and share it among all instances.

this pr builds the message lookup table once in DBC, it can simplify the cython code, and improve the runtime performace.

~For ease of review. This PR only refactored the packer related code. The parser related is put in the next PR. https://github.com/commaai/opendbc/pull/943~ already implemented in this PR.

this pr may also resolve: https://github.com/commaai/opendbc/issues/500, https://github.com/commaai/opendbc/issues/902

### DIFF
--- a/can/packer.cc
+++ b/can/packer.cc
@@ -92,6 +92,6 @@ std::vector<uint8_t> CANPacker::pack(uint32_t address, const std::vector<SignalP
 }
 
 // This function has a definition in common.h and is used in PlotJuggler
-Msg* CANPacker::lookup_message(uint32_t address) {
-  return (Msg*)dbc->addr_to_msg.at(address);
+const Msg* CANPacker::lookup_message(uint32_t address) {
+  return dbc->addr_to_msg.at(address);
 }
--- a/can/packer.cc
+++ b/can/packer.cc
@@ -33,7 +33,6 @@ CANPacker::CANPacker(const std::string& dbc_name) {
   assert(dbc);
 
   for (const auto& msg : dbc->msgs) {
-    message_lookup[msg.address] = msg;
     for (const auto& sig : msg.sigs) {
       signal_lookup[std::make_pair(msg.address, sig.name)] = sig;
     }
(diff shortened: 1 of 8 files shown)


sshane:
@deanlee can you rebase?

deanlee:
ok but I'm on vacation without laptop. It might be ten days before I go back.

deanlee:
rebased

sshane:
@deanlee can you fix so it passes the test?

deanlee:
sure

deanlee:
Should we modify the test case to catch exceptions, or change opendbc to avoid throwing exceptions and output empty values instead?

deanlee:
I think it would be better to throw exceptions. What do you think?

sshane:
I would ideally like to match the previous behavior in a refactor like this. For context, adeeb didn't want to throw an exception on unknown message *signals* as there is a non-zero chance we don't hit an obscure code path in the car controllers and we ship a regression that goes untested.

Now for messages, that is far more unlikely we won't be able to hit that with our fuzzy test, but still non-zero given we conditionally send button messages.

deanlee:
Completed. return empty values and log invalid addresses / signal names using LOGE_100

can/packer_pyx.pyx
```diff
@@ -37,11 +37,17 @@ cdef class CANPacker:
...
-    if isinstance(name_or_addr, int):
-      m = self.dbc.addr_to_msg.at(name_or_addr)
-    else:
-      m = self.dbc.name_to_msg.at(name_or_addr.encode("utf8"))
-
-    cdef vector[uint8_t] val = self.pack(m.address, values)
-    return [m.address, 0, (<char *>&val[0])[:val.size()], bus]
+    try:
+      if isinstance(name_or_addr, int):
+        m = self.dbc.addr_to_msg.at(name_or_addr)
```
sshane, line 44: you can save a lookup here, just use name_or_addr?

can/packer.cc
```diff
@@ -99,6 +98,6 @@ std::vector<uint8_t> CANPacker::pack(uint32_t address, const std::vector<SignalP
 }
 
 // This function has a definition in common.h and is used in PlotJuggler
-Msg* CANPacker::lookup_message(uint32_t address) {
-  return &message_lookup[address];
+const Msg* CANPacker::lookup_message(uint32_t address) {
+  return dbc->addr_to_msg.at(address);
 }
```
sshane, line 103: @deanlee can you open a new PR with this removed? PlotJuggler can just use a `dbc` now, right?

PR #945, opened 2023-09-08
Honda Civic 2022: Add Speed Limit message

landonh12:
Found the message and signal for speed limit detection in 2022+ Honda Civics. As soon as a speed limit sign leaves the frame on the right side of the camera feed, a CAN message with ID 0x1EF is sent with the speed limit at byte 5 in MPH. See below:

image

I am in the US, and I'm not sure if other parts of the world have a separate message for KPH and for signs dissapearing on the other side of the image in RHD countries.

Also, it looks like cabana rearranged the DBC file, please let me know if I need to do this again through a text editor to conserve the DBC order.


landonh12:
Need to use generator instead of saving from cabana. Will reopen PR properly!

landonh12:
Please delete this - it is a duplicate

PR #946, opened 2023-09-08
Honda Civic 2022: Add traffic sign detections

landonh12:
Found the message and signal for speed limit detection in 2022+ Honda Civics. As soon as a speed limit sign leaves the frame on the right side of the camera feed, a CAN message with ID 0x1EF is sent with the speed limit in MPH at byte 5. See below:

image

I am in the US, and I'm not sure if other parts of the world have a separate message for KPH and for signs dissapearing on the other side of the image in RHD countries.


kevinharbin:
Here are the fields that I recorded, with a couple notes.
I was able, once or twice to trigger the sign detection by walking a printed sign past my parked car.
Plus, here is a note from the manual for any testing you might do. I was able, once or twice to trigger the 
image

<html><body>
<div dir="ltr" style="margin-left:0pt;" align="left" id="docs-internal-guid-4e4b8dcf-7fff-1cae-5b48-f0843fa0ecbf">

Hex | Binary | Seen in Car
-- | -- | --
00 | 00000000 | OFF
  |   |  
  |   |  
61 | 01100001 | 5mph=1*5
62 | 01100010 |  
63 | 01100011 | 15
64 | 01100100 |  
65 | 01100101 | 25
66 | 01100110 | 30
67 | 01100111 | 35
68 | 01101000 | 40
69 | 01101001 | 45=9*5
6a | 01101010 |  
6b | 01101011 | 55=11*5
6c | 01101100 | 60
6d | 01101101 |  
6e | 01101110 | 70mph=14*5
6f | 01101111 |  
70 | 01110000 |  
7c | 01111100 |  
7d | 01111101 | No sign detected. 
7e | 01111110 | camera not available, crossed out sign [Malfunction/too cold/etc]

</div>
</body>
</html>

landonh12:
Opening this up for review - I've sort of stalled on the work. Main thing that needs to be reviewed is whether or not the messages I added are in the correct files. I don't have another Honda to test on so I'm not sure if these messages are specific to 2022+ Civics. Will post in the Discord to see if anyone else can find these on their Accord, Pilot, Ridgeline, etc..

vanillagorillaa:
LGTM - These signals are also present on the 2023 Pilot (port in progress). Would be nice to get more values for the road sign signal. Wonder what else it detects (and it's value) besides stop signs.

landonh12:
One detail to add - it seems that the speed limit does not follow units, only the number seen on the sign.

landonh12:
An update to this PR:

I drove a 2024 Honda Accord as a loaner from a dealership. The full screen dash display also shows additional information, such as stop signs, school speed limits, etc. So that's likely where this data from Bosch ADAS gets used.

adeebshihadeh:
Can you rebase? Otherwise, LGTM.

jyoung8607:
LGTM as well, and it rebased pretty cleanly.

Didn't appear to have any impact to working car ports, but just in case, validated offline:

* `pytest opendbc` passes
* `pytest selfdrive/car/tests` passes
* `selfdrive/test/process_replay/test_processes.py`, no changes

PR #947, opened 2023-09-10
VAG focusing on TCU

rusefillc:
Some TCU fields translation


rusefillc:
@sshane @adeebshihadeh poke :)

PR #961, opened 2023-10-10
fix(README.md): start at 2nd bit and reduce length by one

noel-schenk:


jyoung8607:
Closing. I looked at the existing README and explanation, and I think it's correct as it stands. If you think it's a problem, please open a new PR explaining what you think is wrong or unclear, against the DBC README in its new location.

PR #976, opened 2023-11-18
CANParser: clip `counter_fail` to the max

sshane:
Matches panda's behavior exactly, and solves a mismatch in https://github.com/commaai/openpilot/pull/30443

### DIFF
--- a/can/parser.cc
+++ b/can/parser.cc
@@ -77,18 +77,12 @@ bool MessageState::update_counter_generic(int64_t v, int cnt_size) {
   uint8_t old_counter = counter;
   counter = v;
   if (((old_counter+1) & ((1 << cnt_size) -1)) != v) {
-//    counter_fail += 1;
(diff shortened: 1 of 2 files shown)


can/parser.cc
```diff
@@ -74,20 +74,14 @@ bool MessageState::parse(uint64_t nanos, const std::vector<uint8_t> &dat) {
...
-    }
+  if (((counter + 1) & ((1 << cnt_size) -1)) != v) {
+    counter_fail = std::min(counter_fail + 1, MAX_BAD_COUNTER);
+    LOGE("0x%X COUNTER FAIL #%d -- %d -> %d\n", address, counter_fail, counter, (int)v);
   } else if (counter_fail > 0) {
     counter_fail--;
   }
-  return true;
+  counter = v;
+  return counter_fail < MAX_BAD_COUNTER;
```
sshane, line 84: This change has no effect on this PR's behavior, but before you could wind up `counter_fail` to a high value, but one good counter would be valid again

PR #981, opened 2023-12-01
Hyundai CAN: Fix `Elect_Gear_Shifter` Sport Gear value

sunnyhaibin:
The value for Sport gear is wrong in the DBC for EV/HEV/PHEV platforms.

Route ID: `26a38e4bf36bb74a|2023-11-15--10-11-14`

**Prerequisite of**
- https://github.com/commaai/openpilot/pull/30648

image


devtekve:
@adeebshihadeh sorry for the ping but this would make our life nicer on hev platforms <3

adeebshihadeh:
Where did the old value come from? How many platforms have been checked? Have all three of EV, PHEV, and HEV been checked?

ghostnetworksolutions:
I have same car as DevTekVe and would like this change also.

This was my video from above sunny has inputted here to commit to master

sshane:
No real source for this, it was copied from CF_Lvr_Gear in the OG port: https://github.com/commaai/openpilot/pull/1362, commit: https://github.com/commaai/openpilot/commit/e43fa32077e7d1dfd75dfdd9f18ab21525dde162

sshane:
Here's a random Ioniq PHEV, it can be both 4 and 8, so we can't remove the 8 signal def:

image

---

Can you guys try to replicate this so we can understand the meaning? Are there different sport modes?

sunnyhaibin:
@sshane Brought back 8.

ghostnetworksolutions:
Any update on this ? We have confirmed this working on PHEV platforms

ghostnetworksolutions:
@sshane 

Will there be anything else that is needed to merge this?

devtekve:
> Here's a random Ioniq PHEV, it can be both 4 and 8, so we can't remove the 8 signal def:
> 
> image
> 
> Can you guys try to replicate this so we can understand the meaning? Are there different sport modes? Try to toggle _all_ the gears/sport options and post a route.
> 
> If we can verify that 8 is also sport, then we can merge 4 and 8

Been a while but this came up again recently :D, I will need to test, but you asked 
> Are there different sport modes?
Technically, you can either have it on "S" or you can also have it on a specific gear selected (1 thru 6) and that's still sports mode

PR #985, opened 2023-12-06
Chrysler: DBCs for Compact US Wide

jyoung8607:
Work in progress, based on a port originally developed by @Tundergit. Holding in draft pending various cleanup and better alignment with the other Chrysler ports.


jyoung8607:
I'd like to go ahead and merge this. It's not perfect but it's good enough to move forward. I don't think there's enough true sharing with other Chrysler to bother with the DBC generator at this time. A couple messages vary in length between cars of different ages, which will be annoying to deal with, and THAT may force us toward the generator, but we'll cross that bridge when we come to it.

PR #988, opened 2023-12-11
Ford Edge 2022 DBC: Different PSCM SteeringPinion_Data ID

MikeDoyal:
Added new DBC for 2022 Ford Edge. Edge PSCM reports SteeringPinion_Data at ID 85 which is different from the currently supported models using ID 7e. ford_edge_pt.dbc was started from ford_lincoln_base_pt.dbc with an updated ID for SteeringPinion_Data

### DIFF
--- a/ford_edge_pt.dbc
+++ b/ford_edge_pt.dbc
@@ -1,0 +1,12695 @@
+VERSION ""
+
+
+NS_ :
+    NS_DESC_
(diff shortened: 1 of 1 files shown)


MikeDoyal:
New DBC is not required for Ford Edge MK2 car port commaai/openpilot#30762

Issue #496, opened 2022-01-18
Honda DBC cleanup


adeebshihadeh:
Ideally, we can use a single DBC for all Hondas, like we do with many of the more recent car ports. This can be done incrementally by doing the following:
* merging DBCs
* using common, universal signals when possible
* abstracting out repeated definitions into common DBCs

sshane:
- [ ] See if `LAT_ACCEL` under `KINEMATICS` on 145 is also available on 148 for cars using 145. The LAT_ACCEL signal is defined in the same location in the message
- [ ] Fix duplicate gear signals defined
- [ ] A lot of these messages are common in bosch vehicles, possible unify:
  - [ ] DOORS_STATUS
  - [ ] STANDSTILL
  - [ ] GEARBOX
  - [ ] GAS_PEDAL_2

sshane:
Notes about common signals:

-  [ ] `DOORS_STATUS`:
  - Identical: `honda_civic_hatchback_ex_2017_can`, `honda_civic_sedan_16_diesel_2019_can`, `honda_crv_ex_2017_can`, `honda_insight_ex_2019_can`
  - Missing (using `SCM_FEEDBACK`): `honda_crv_hybrid_2019_can`, `honda_accord_2018_can`, `acura_rdx_2020_can`
- [ ] `GAS_PEDAL_2`: Not used in openpilot
- [ ] `STEERING_CONTROL`:

Common messages:

- [ ] Common for _bosch_2020 and _honda_2017: `VSA_STATUS`, `STEER_MOTOR_TORQUE`, `STANDSTILL`, `ENGINE_DATA`, `POWERTRAIN_DATA`, `WHEEL_SPEEDS`, `VEHICLE_DYNAMICS`

[...]

PR #998, opened 2024-01-16
Honda: Remove BOH signal overlap. Add 'ENABLED' in LKAS_HUD. Full coverage of SCM_BUTTONS. Add VALs to Civic 2022

csouers:
Related to: https://github.com/commaai/openpilot/pull/31022

**Adjustments**:
- Remove overlapping signal in LKAS_HUD
- Define the ENABLED bit for the stock LKAS systems; used by Honda Bosch Radarless and stock ACC.
- Honda Bosch: Define catch-all signals in SCM_BUTTONS allowing full pass-through/recreation of the frame without reverse engineering all undiscovered signals.
- Civic 22: Add missing value (VAL) defs for both button signals (cruise and setting)


jyoung8607:
This does contain some valid changes, but would need a lot of work to catch up with modern opendbc as well as some additional car port code changes. We'll capture most or all of this as part of #496.

PR #1002, opened 2024-01-23
Toyota: LVSTP definition

cydia2020:
Seems like LVSTP means Lead Vehicle SToPped, it's set to 1 on stock when lead is stopped. Thanks @Smartype for the initial discovery. For commaai/openpilot#30697


generator/toyota/_toyota_2017.dbc
```diff
@@ -405,6 +406,7 @@ CM_ SG_ 835 ACC_MALFUNCTION "display ACC fault on dash if set to 1";
 CM_ SG_ 835 ACC_CUT_IN "Display blinking yellow lead if set to 1";
 CM_ SG_ 835 DISTANCE "Cycle through ACC following distance from long, mid, short when set to 1";
 CM_ SG_ 835 ITS_CONNECT_LEAD "Displayed when lead car is capable of ITS Connect";
+CM_ SG_ 835 LEAD_VEHICLE_STOPPED "LVSTP in leaked DBC, set to 1 when lead is stopped";
```
sshane, line 409:
```suggestion
CM_ SG_ 835 LEAD_VEHICLE_STOPPED "Set to 1 when lead is stopped, likely only used in older TSS-P vehicles";
```

PR #1009, opened 2024-02-13
GM: Add sportmode

AreYouRlyScared:
Allows us to track when Sport mode is enabled. 

image

D2E/E2D has been removed as we have no reason to know when the user clicked the button, to enable/disable.

Sport mode in Bolts increases the stiffness in the steering wheel, resulting in less of a turning radius and more torque required to steer. As seen below, disabling sport mode jerks the steering wheel. 

https://github.com/commaai/opendbc/assets/1757840/34213edc-18c2-48a8-88c4-65c455cfe5da


sshane:
Called `DrvSelMd1Stat`, `VAL_TABLE_ DrvSelMd1Stat 1 "Active" 0 "Inactive" ;`

PR #951, opened 2023-09-12
GM: fix ASCMGasRegenCmd scale and units

sshane:
Correct scaling and offset now. Plus checksum is 1 bit larger, which captures that fake inverse bit, removed some set mes.

Opened to preserve, will verify later.

This does not change behavior.

PR #1025, opened 2024-03-20
GM: Set new longitudinal scaling

nworb-cire:
Superseding #951 to fix overlapping signal definitions


sshane:
I see in the original DBC the signals are often sorted arbitrarily, can we leave that out from this PR? It inflates the diff.

1454:
Is there any progress on this to help us GM1500 folks with painfully slow acceleration? It's borderline unusable.

sshane:
fyi: https://github.com/commaai/opendbc/pull/951#issuecomment-2387238165

PR #1027, opened 2024-03-21
Ford: CAN-FD radar messages

blue-genie:
CANFD MRR_Detection messages:
image

Current mapping:
image

New Mapping:
image
image
image

Might be useful with RT and later models.


blue-genie:
Original debug UI view:
image

Updated view using all the tracks available in the 64 byte messages:
image

adeebshihadeh:
Can you rebase? Otherwise, LGTM!

blue-genie:
@adeebshihadeh rebased on the latest master, not sure why the PR was automatically closed when I synced my branch.

PR #1028, opened 2024-03-21
VW MQB: Additional TSK support

jyoung8607:
Updates to `TSK_06` and `TSK_07`, new message `TSK_08`, with VAL enums.

Holding in Draft while we figure out how to detect switchable CC/ACC modes on the cars that support it. It's unclear how we'll detect that yet, and if it's in `TSK_07` or `TSK_08` I'll need to add checksum support code.


jyoung8607:
This can be merged as-is, now that the DBC-relevant part of the speed-limiter/CC/ACC question has been resolved,

PR #1029, opened 2024-03-25
VW MQB: add adaptive cruise signal

sshane:
Needs verification


sshane:
Not using this in openpilot, but it seems to match nearly the same as the UI distance signal we chose. This is comparing it with the blinker as the user toggled CC mode on and off cc @jyoung8607 
image

sshane:
Here's a few more:

image
image

PR #1039, opened 2024-05-06
CANParser: 2x faster parsing

deanlee:
Issue: The following two code blocks consume nearly two-thirds of cython's update_string()'s runtime:

```
for v in self.vl_all.values():
  for l in v.values():  # no-cython-lint
    l.clear()
```

```
while it != new_vals.end():
  cv = &deref(it)
  # Cast char * directly to unicode
  cv_name = <unicode>cv.name
  self.vl[cv.address][cv_name] = cv.value
  self.vl_all[cv.address][cv_name] = cv.all_values
  self.ts_nanos[cv.address][cv_name] = cv.ts_nanos
  updated_addrs.insert(cv.address)
  preinc(it)

```
Resolve:
1. Clear only the address key's value instead of all key-value pairs, as the name key's value refers to the address. This change halves the execution time of the first block of code.
2. Since the vector return from CanParser::update_strings() groups data by address, caching dictionary references for each address key reduces redundant lookups. This simple adjustment significantly cuts execution time by at least half, often more.
3. disable wraparound and boundcheck, update_string does not require these two options. disabling them can slightly improve performance

the results of running `selfdrive/debug/check_can_parser_performance.py` :

Before:

> 6000 CAN packets, 10 runs
> 283.51 mean ms, 297.98 max ms, 275.58 min ms, 5.45 std ms
> 0.0473 mean ms / CAN packet

After:

> 6000 CAN packets, 10 runs
> 128.54 mean ms, 148.90 max ms, 119.80 min ms, 9.34 std ms
> 0.0214 mean ms / CAN packet
>

### DIFF
--- a/can/parser_pyx.pyx
+++ b/can/parser_pyx.pyx
@@ -1,7 +1,6 @@
 # distutils: language = c++
 # cython: c_string_encoding=ascii, language_level=3
 
-cimport cython
 from cython.operator cimport dereference as deref, preincrement as preinc
 from libcpp.pair cimport pair
 from libcpp.string cimport string
@@ -72,8 +71,6 @@ cdef class CANParser:
     if self.can:
       del self.can
 
-  @cython.wraparound(False)
-  @cython.boundscheck(False)
   def update_strings(self, strings, sendcan=False):
     for address in self.addresses:
       self.vl_all[address].clear()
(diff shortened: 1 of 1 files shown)


sshane:
> Since the vector return from CanParser::update_strings() groups data by address, caching dictionary references for each address key reduces redundant lookups. This simple adjustment significantly cuts execution time by at least half, often more.

I'm seeing a very small reduction of about 7% for this optimization on device: `master`: 747 mean ms, address group caching: 693 mean ms, I don't think it's worth it right now. Can we refactor the C++ CAN parser so it groups the data?

The clearing is what more than halved the time for me on device and on PC. I checked and we don't rely on every signal being inside `vl_all` if there aren't any new messages, for example we never do `vl_all.keys()` or `len(vl_all)`, so missing signals that return `[]` on get is fine.

PR #1040, opened 2024-05-07
CANParser: Update CAN data handling functions to accept  raw data and size parameters

deanlee:
When parsing CAN messages, `CanParser` creates a memory copy for each message solely to satisfy the checksum calculation function's requirement, which expects a `std::vector<uint8_t>` as parameter. This approach is inefficient and slows down the parsing process.  

> std::vector<uint8_t> data(dat.size(), 0);
> memcpy(data.data(), dat.begin(), dat.size());
> 

This PR updates all functions responsible for handling CAN data to accept raw data pointers `uint8_t* data` and size parameters `size_t size` instead of using `std::vector<uint8_t>`. This adjustment aims to enhance performance and reduce memory overhead by avoiding unnecessary memory allocations and copies associated with vector usage.

### DIFF
--- a/can/parser.cc
+++ b/can/parser.cc
@@ -13,12 +13,12 @@
 #include "cereal/logger/logger.h"
 #include "opendbc/can/common.h"
 
-int64_t get_raw_value(const std::vector<uint8_t> &msg, const Signal &sig) {
+int64_t get_raw_value(const uint8_t *msg, size_t msg_size, const Signal &sig) {
   int64_t ret = 0;
 
   int i = sig.msb / 8;
   int bits = sig.size;
-  while (i >= 0 && i < msg.size() && bits > 0) {
+  while (i >= 0 && i < msg_size && bits > 0) {
     int lsb = (int)(sig.lsb / 8) == i ? sig.lsb : i*8;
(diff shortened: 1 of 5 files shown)


can/parser.cc
```diff
@@ -238,9 +238,7 @@ void CANParser::UpdateCans(uint64_t nanos, const capnp::List<cereal::CanData>::R
     //  continue;
     //}
 
-    std::vector<uint8_t> data(dat.size(), 0);
-    memcpy(data.data(), dat.begin(), dat.size());
-    state_it->second.parse(nanos, data);
+    state_it->second.parse(nanos, dat.begin(), dat.size());
```
sshane, line 241: Can we just pass the `capnp::Data::Reader`, or do we want to remove capnp as a dependency too?

sshane:
How much does this improve performance? I don't see much of a difference on a comma 3X

deanlee:
This change won't significantly impact the performance of the master branch, as the master spends lots of time on Cython code. However, it will enhance the performance of the C++ code, particularly after https://github.com/commaai/opendbc/pull/1046. Currently, there are two performance bottlenecks in the C++ code that can be addressed: memory-aligned copying and the use of vectors. This adjustment is expected to boost the performance of the C++ code, with even greater improvements expected on devices.

PR #1042, opened 2024-05-14
CANParser: faster parsing

sshane:
`cv_name = <unicode>cv.name` was very slow for every signal. We already know the address to name mapping.

PC master: 258ms
PC https://github.com/commaai/opendbc/pull/1039: 90-100ms
PC above+this branch: 80ms


sshane:
oh duh, this is the signal name!

Issue #1003, opened 2024-01-24
Don't spam CAN error log messages


sshane:
Can lead to huge log sizes

PR #1048, opened 2024-06-04
improve logging

deanlee:
1. Eliminate the dependency on cereal/logger.h and use own logger.h instead.
2. Support cloudlog_rl in logger.h, resolve https://github.com/commaai/opendbc/issues/1003


adeebshihadeh:
Nice! Btw Shane is still working on getting your other PRs merged related to the rest of the cereal removal.

PR #1057, opened 2024-06-21
Honda: Non-hybrid CRV gearbox

vanillagorillaa:
This gearbox message is actually used on the non-Hybrid 6th gen CRV. Similar to the Accord 1.5/2.0L gearbox differences. 

Verification - 2dc4489d7e1410ca/00000004--eb4c907d01


beckej13820:
@sshane This is another example of a pull request that would benefit Honda/Acura users, but was closed by vanillagorillaa because he is no longer contributing to OpenPilot.

If this could be left open and reviewed, members of the Honda/Acura channel can provide routes.

PR #1059, opened 2024-07-05
Sconscript: enable openpilot compilation for macOS

bongbui321:
prereq for: https://github.com/commaai/openpilot/pull/32909

Good reference: http://clarkkromenaker.com/post/library-dynamic-loading-mac/#:~:text=On%20Mac%20and%20Linux%2C%20the,path%20embedded%20inside%20the%20executable.

MacOS linking to dynamic library on scons is very weird, and RPATH is not well supported on macOS.

reference:
Similar trouble I encounter: https://github.com/SCons/scons/issues/3128
Weird behaviour on macOS for rpath: https://github.com/SCons/scons/issues/2127

Edit:
Uses `@loader_path` instead, for relative path to the file referencing the dynamic lib

### DIFF
--- a/can/SConscript
+++ b/can/SConscript
@@ -9,11 +9,10 @@ src = ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]
 libs = [common, "capnp", "kj", "zmq"]
 
 # shared library for openpilot
-libdbc = envDBC.SharedLibrary('libdbc', src, LIBS=libs)
+LINKFLAGS = envDBC["LINKFLAGS"] + ["-Wl,-install_name,@rpath/libdbc.dylib"] if arch == "Darwin" else envDBC["LINKFLAGS"]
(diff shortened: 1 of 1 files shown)


can/SConscript
```diff
@@ -9,7 +9,8 @@ src = ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]
 libs = [common, "capnp", "kj", "zmq"]
 
 # shared library for openpilot
-libdbc = envDBC.SharedLibrary('libdbc', src, LIBS=libs)
+LINKFLAGS = envDBC["LINKFLAGS"] + ["-Wl,-install_name,@loader_path/libdbc.dylib"] if arch == "Darwin" else envDBC["LINKFLAGS"]
```
adeebshihadeh, line 12: can you split this into multiple lines?

adeebshihadeh (changes requested):

PR #1060, opened 2024-07-09
Convert tests to pytests

UkuLoskit:
Part of https://github.com/commaai/openpilot/issues/32941

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Build opendbc
       run: ${{ env.RUN }} "cd ../ && scons -j$(nproc) --minimal"
     - name: Unit tests
-      run: ${{ env.RUN }} "pytest"
+      run: ${{ env.RUN }} "pytest -n auto"
 
   static-analysis:
     name: static analysis
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Build opendbc
       run: ${{ env.RUN }} "cd ../ && scons -j$(nproc) --minimal"
     - name: Unit tests
-      run: ${{ env.RUN }} "pytest -n auto"
+      run: ${{ env.RUN }} "pytest -n logical"
 
   static-analysis:
     name: static analysis
--- a/can/tests/test_define.py
+++ b/can/tests/test_define.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-import unittest
-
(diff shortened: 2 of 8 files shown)


.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     - name: Build opendbc
       run: ${{ env.RUN }} "cd ../ && scons -j$(nproc) --minimal"
     - name: Unit tests
-      run: ${{ env.RUN }} "pytest"
+      run: ${{ env.RUN }} "pytest -n auto"
```
sshane, line 25: openpilot uses -n logical for hyperthreading I believe

can/tests/test_define.py
```diff
@@ -1,35 +1,27 @@
...
   def test_civic(self):
 
     dbc_file = "honda_civic_touring_2016_can_generated"
     defs = CANDefine(dbc_file)
 
-    self.assertDictEqual(defs.dv[399], defs.dv['STEER_STATUS'])
-    self.assertDictEqual(defs.dv[399],
-                         {'STEER_STATUS':
+    assert defs.dv[399] == defs.dv['STEER_STATUS']
+    assert defs.dv[399] == \
```
sshane, line 12: are parenthesis more pretty here?

can/tests/test_define.py
```diff
@@ -1,35 +1,27 @@
...
+                         {'STEER_STATUS': \
```
sshane, line 13:
```suggestion
                         {'STEER_STATUS':
```

I don't think you need backslash here?

can/tests/test_define.py
```diff
@@ -1,35 +1,27 @@
...
                           {7: 'PERMANENT_FAULT',
                            6: 'TMP_FAULT',
                            5: 'FAULT_1',
                            4: 'NO_TORQUE_ALERT_2',
                            3: 'LOW_SPEED_LOCKOUT',
                            2: 'NO_TORQUE_ALERT_1',
-                           0: 'NORMAL'}
+                           0: 'NORMAL'} \
```
sshane, line 20:
```suggestion
                           0: 'NORMAL'}
```

.github/workflows/tests.yml
maxime-desroches: Normally it uses auto: https://github.com/commaai/openpilot/blob/master/pyproject.toml#L139
but in CI we use logical: https://github.com/commaai/openpilot/blob/master/.github/workflows/selfdrive_tests.yaml#L30

PR #1063, opened 2024-07-18
clean/remove community (SDSU/Pedal/IPAS) dbc

eFiniLan:


adeebshihadeh:
I don't think it's necessary to remove the base community.dbc, though I'd merge the removal of the inclusion of it. 

It's helpful for us when debugging routes and we need to check if a user has this HW installed.

PR #1064, opened 2024-07-21
Tesla: Move DI_speed to correct bus and update signals within

carleeno:
This PR moves the DI_speed message definition to the correct bus and updates the signals within to match the actual canbus data from model 3/Y

DI_speed doesn't exist on the party bus (at least not at 599).

At the moment nothing uses DI_speed so it wasn't discovered that it wasn't present on party bus. As such this change is a no-op to OP behavior on Tesla, but enables a future improvement to use uiSpeed for vEgoCluster

verification: bbbf82d987d681bc/000001cc--b24bcec625/0


lukasloetkolben:
The message is on both buses. i wouldn't change anything here.

carleeno:
not needed after all

PR #1069, opened 2024-07-31
move to nested opendbc/

sshane:
prepare to package-ization


adeebshihadeh:
Let's nest all the dbc-related stuff in a `dbc` directory too.

adeebshihadeh:
generator should go in dbc/

Issue #1068, opened 2024-07-31
parser: clean up timestamps


sshane:
related: https://github.com/commaai/opendbc/pull/1067

Issue #1066, opened 2024-07-31
`current_nanos` can be previous message timestamp


sshane:
https://github.com/commaai/opendbc/pull/934/files#r1697941074

sshane:
Addressed by https://github.com/commaai/opendbc/pull/1382

PR #1046, opened 2024-05-26
Refactor CAN Parser

deanlee:
This PR refactors the existing code by shifting most of the logic to C++, significantly simplifying the update process, improving performance, and resolving issues related to the mixed update logic that was previously split between Cython and C++. The parser_pyx.pyx now functions as a streamlined interface wrapper around the C++ functions.

**Key Changes:**

1. **Removed query_latest() Function:**  The query_latest() function, which previously copied updated values to Cython and was invoked with empty CAN strings in Cython's `__init__` method to initialize value dictionaries, has been removed. This led to unnecessary complexity and has now been streamlined.
2. **Exposed MessageState to Cython:** MessageState is now directly accessible from Cython, allowing values to be read directly from the C++ MessageState object without the need for intermediate copying. This change streamlines data access, reduces overhead, and minimizes complexity and potential bugs.
3. **Enhanced Message Tracking:** Updates are now tracked using and returning a set of updated addresses in MessageState::update, ensuring comprehensive capture of every update, independent of can frame's timestamp.

By centralizing most of the logic in C++, this PR simplifies the Update Logic, Reduces the complexity of the code by removing redundant operations and unnecessary data copying.The new implementation is more efficient, as demonstrated by performance comparisons.

**Issues Resolved:**

This PR resolves multiple issues, and the test cases have been updated to catch them:

   - [Issue 1068](https://github.com/commaai/opendbc/issues/1068)
   - [Issue 913](https://github.com/commaai/opendbc/issues/913)
   - [Issue 1066](https://github.com/commaai/opendbc/issues/1066) (partially resolved)

**Performance Comparison:**

master:

> 6000 CAN packets, 10 runs
327.69 mean ms, 334.36 max ms, 321.73 min ms, 3.74 std ms
0.0546 mean ms / CAN packet

this branch:
> 6000 CAN packets, 10 runs
113.71 mean ms, 125.21 max ms, 109.85 min ms, 4.42 std ms
0.0189 mean ms / CAN packet

**Additional Notes:**

`PlotJuggler` will be broken after this PR because data is now read directly from `MessageState` instead of being copied using `query_latest`. However, it is quite easy to modify `PlotJuggler` to use the new API.

PR #1072, opened 2024-08-01
parser:  refactor value dictionary update logic

deanlee:
Main Changes:

1. Removed the `SignalValue` struct and the `query_latest()` function. These were previously used to copy updated values to Cython code and initializing the value dictionary in Cython's `__init__` method when query_latest() was called with empty CAN strings.
2. Exposed `MessageState` to Cython code, enabling direct access to values from `MessageState` after an update.
3. Tracked updated messages using and returning a set of updated_address in MessageState::update. This ensures that every update is captured and is not related to the `current_nanos`.
4. Simplified and robustified the initialization of the value dictionary maps.

These changes significantly simplify the update logic, improve performance, and eliminate issues related to the problematic `last_nanos`  and `current_nanos` in query_latest(). This refactor fixes multiple issues, including:

   - [Issue 1068](https://github.com/commaai/opendbc/issues/1068)
   - [Issue 913](https://github.com/commaai/opendbc/issues/913)
   - [Issue 1066](https://github.com/commaai/opendbc/issues/1066) (partially resolved)

Remaining Tasks for Issue 1066 (will be resolved in another PR):
    - Ensure the `vals` in MessageState are always the newest.
    - Ensure `all_vals` are sorted by received time.

### DIFF
--- a/opendbc/can/common.h
+++ b/opendbc/can/common.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <map>
+#include <set>
 #include <string>
 #include <utility>
 #include <unordered_map>
@@ -65,7 +66,6 @@ class CANParser {
 private:
   const int bus;
   const DBC *dbc = NULL;
-  std::unordered_map<uint32_t, MessageState> message_states;
 
 public:
   bool can_valid = false;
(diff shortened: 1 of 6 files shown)


adeebshihadeh:
Can you add some unit tests to enforce this behavior?

deanlee:
made a few modifications to the test case. It should be able to catch these issues now.

deanlee:
merged to https://github.com/commaai/opendbc/pull/1046

PR #1076, opened 2024-08-04
Toyota: add misc adas/toggle signal definitions

cydia2020:
Not sure if we can send these to control the vehicle in any ways


sshane:
Have you tried any of these signals yet?

cydia2020:
> Have you tried any of these signals yet? Did you decode them from your Prius 2017?

Yea they're from my Prius.

PR #1081, opened 2024-08-17
no docker

adeebshihadeh:


adeebshihadeh:
This is pretty minimal now, but build times are still much longer than I'd expect. @deanlee want to look into why the build time is as long as it is? Is it SCons, is it Cython, are we not parallelizing properly, etc?

deanlee:
here is a cleanup: https://github.com/commaai/opendbc/pull/1084 . Cabana no longer requires the static library from opendbc.

opendbc/can/SConscript
```diff
@@ -1,30 +1,24 @@
 Import('env', 'envCython', 'common', 'arch')
 
-import os
-
 envDBC = env.Clone()
 dbc_file_path = '-DDBC_FILE_PATH=\'"%s"\'' % (envDBC.Dir("../dbc").abspath)
 envDBC['CXXFLAGS'] += [dbc_file_path]
 src = ["dbc.cc", "parser.cc", "packer.cc", "common.cc"]
-libs = [common, "zmq"]
```
deanlee, line 9: Swaglog still requires the zmq library for compilation.

PR #1085, opened 2024-08-18
fix: relax ci timout 1min->2min

miettal:
some time ci execution reached to over 1 minutes. so this PR change ci timeout 1min to 2min.

Screenshot 2024-08-18 at 20 00 35

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
   unit-tests:
     name: unit tests
     runs-on: ubuntu-latest
-    timeout-minutes: 1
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Got some examples? We want to keep this as one minute.

miettal:
FYI
https://github.com/commaai/opendbc/actions/runs/10436143598/job/28900834202
https://github.com/commaai/opendbc/actions/runs/10439518460/job/28908227154

sshane:
Ci should be significantly faster, we have ~30 seconds of tolerance now

miettal:
👍

PR #1089, opened 2024-08-19
fp

sshane:
to test this offline:

with event.json:
```
{
  "pull_request": {
    "head": {
      "ref": "test-labeler"
    },
    "base": {
      "ref": "master"
    },
    "number": 1089
  }
}

```

```
gh act pull_request_target -j labeler -e event.json -s GITHUB_TOKEN=...
```


.github/workflows/auto_pr_review.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
     - uses: actions/labeler@v5.0.0
       with:
         dot: true
+        sync-labels: true
```
adeebshihadeh, line 20: I believe this will remove manual labels, but maybe that's fine in this repo. would be cool if it's all auto

sshane:
Thanks for contributing to opendbc! In order for us to review your PR as quickly as possible, check the following:
* Convert your PR to a draft unless it's ready to review
* Read the [contributing docs](https://github.com/commaai/openpilot/blob/master/docs/CONTRIBUTING.md)
* Before marking as "ready for review", ensure:
  * the goal is clearly stated in the description
  * all the tests are passing
  * include a route or your device' dongle ID if relevant

PR #1094, opened 2024-08-20
Add support for 2023-2024 Ford Escape (CAN-FD)

northportio:
Adds support for 2023-2024 Ford Escape (CAN-FD) -  Previously https://github.com/commaai/openpilot/pull/32319

Dependent on https://github.com/commaai/panda/pull/1524 getting merged (Brings CAN-FD support).

I've been running another fork with similar config for almost 5 months (4462 miles, 476 drives, 197 hours as of opening this PR). 

Findings:
- Outside of the turn angle warnings over 35º requiring manual intervention for ~10 seconds after turn is completed (which affects all fords), It truly does make driving chill.
- Rare Red CAN-Bus mismatch/slow/error causing a forced disengagement - possibly due to length of USB-C from wiring harness in rear loadspace. Happens 2x/mo. Requires vehicle restart or reboot of comma 3x.


opendbc/car/ford/values.py
```diff
@@ -108,6 +108,13 @@ class CAR(Platforms):
     ],
     CarSpecs(mass=1750, wheelbase=2.71, steerRatio=16.7),
   )
+  FORD_ESCAPE_MK4_23REFRESH = FordCANFDPlatformConfig(
```
adeebshihadeh, line 111: any reason this is a separate openpilot platform than `FORD_ESCAPE_MK4`?

adeebshihadeh:
Nope, no timeline or updates. Car support happens when it happens. The better state a port is in and the more popular the car is, the more likely we'll pick it up next. Ford Q4 has been at the top of the list for a while, but we just haven't had time for it yet.

adeebshihadeh:
@northportio we can get this merged in once CI is passing though

jyoung8607:
Superseded by #1922, which was merged earlier this year.

PR #1177, opened 2024-08-20
GM: Car Port for XT4 2023

garrettpall:
**Car**
Cadillac XT4 2023

**Route**
162796f1469f2f1b/00000005--6f334eda14

**Checklist**
- [x] added entry to CarInfo in opendbc/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: 162796f1469f2f1b/00000005--6f334eda14
- [x] route with stock system: 162796f1469f2f1b/00000003--79f5d5854f (old harness design)
- [x] car harness used (if comma doesn't sell it, put N/A): N/A https://github.com/commaai/openpilot/issues/32471

Supersedes https://github.com/commaai/openpilot/pull/30420


adeebshihadeh:
This ports looks smaller than I remember? If this is it, I think we're good to merge it.

garrettpall:
> This ports looks smaller than I remember? If this is it, I think we're good to merge it.

Let me test it on Friday. I haven’t tested since the switch of cars to opendbc. But the smaller port is due to an updated harness design (intercept different pins).

adeebshihadeh:
Converting to draft until then. We should also update the test route.

garrettpall:
Tested with `car` moved to opendbc. Everything works as expected.

Route: `162796f1469f2f1b/0000009d--717c2e1c0a`

(note: I have a SASCM (sdsu variant) installed that broadcasts `767, 4`, this was added to the fingerprint for testing locally but should not be included upstream. This route demonstrates the port works, the route in the PR should be used as good, clean data)

opendbc/car/gm/values.py
```diff
@@ -86,6 +89,12 @@ def init(self):
     # ASCM is supported, but due to a janky install and hardware configuration, we are not showing in the car docs
     self.car_docs = []
 
+@dataclass
+class GMSDGMPlatformConfig(GMPlatformConfig):
+  def init(self):
+    # SDGM is supported, but due to harness not being sold by comma, we are not showing in the car docs
```
adeebshihadeh, line 95:
```suggestion
    # Don't show in docs until the harness is sold. See https://github.com/commaai/openpilot/issues/32471
```

jyoung8607:
CI test route still needs to be uploaded:

image

PR #1179, opened 2024-08-21
speedup CI test setup

sshane:


sshane:
@maxime-desroches it would be nice to `curl tools/lint.sh` and use your own config file, pre-commit setup time takes forever! We should remove it here eventually

PR #1186, opened 2024-08-27
Toyota: raise max longitudinal accel to 2 m/s^2

sshane:
In line with the rest of the cars.

Toyota was first added here at 1.5 for an unknown reason (experimental port?): https://github.com/commaai/one/pull/33

It was the second port in the car folder, after Honda (1 m/s^2??): https://github.com/commaai/one/blob/4db5f6f8970b3bbdbe54877d64ff0606c3f86cda/selfdrive/car/honda/interface.py#L393C18-L393C69

openpilot PR: https://github.com/commaai/openpilot/pull/33386


sshane:
Superseded by recent Lexus tune work, to expand for all Toyota soon. https://github.com/commaai/opendbc/pull/1248

PR #1188, opened 2024-08-27
Reapply "GM: implement allow throttle/brakes"

sshane:
Reverts commaai/opendbc#1187


sshane:
The chosen solution by autonomy is to request coasting accel

PR #1189, opened 2024-08-27
Toyota: API is now gas/brake

sshane:
Needs testing, required for https://github.com/commaai/openpilot/pull/33384

Past PRs for history: https://github.com/commaai/openpilot/pull/29481 // https://github.com/commaai/openpilot/pull/32753


sshane (commented):
longcontrol.py resets the integral every time you disengage, so it may take a few seconds before it learns the offset on each engagement, issue should be on GM (gas/brake) as well

sshane:
There is an undefined signal in `CLUTCH` that describes a very similar attribute, but may not be affected by the ACC_BRAKING bit?

```dbc
BO_ 865 CLUTCH: 8 XXX
 SG_ TRUE_ACCEL_REQ : 48|16@1+ (0.0002,-6.5536) [0|4095] "" XXX
```

image

sshane:
GEAR_PACKET_HYBRID->CAR_MOVEMENT might be some force exerted by the engine/hybrid motors, need to find the ICE equivelent. And BRAKE_PEDAL->BRAKE might be the actual brake percentage!

While reviewing some recent routes using the new signal in the CLUTCH message, I discovered why there might still be some sub optimal performance. It appears that `PCM_CRUISE->ACCEL_NET` is only accurate for braking (it's a simple scaling factor compared to a braking signal) and the new signal (`CLUTCH->ACCEL_NET`) is only accurate for acceleration with another scaling factor, but doesn't capture some small residual force the engine applies at low speed for creep, which is why the car kept releasing brakes too much coming to a stop (seen in the plot):

image

Here is switching the acceleration read back signals to show how each is only accurate for brake/gas:

image

adeebshihadeh:
Good to close?

sshane:
Superseded by recent Lexus tune work, to expand for all Toyota soon. https://github.com/commaai/opendbc/pull/1248

PR #1195, opened 2024-08-28
Subaru: remove LKAS_ANGLE cars from dashcam only

martinl:
This PR removes Subaru LKAS_ANGLE cars from dashcam only. Please let me know if anything is still missing before this can be done. I did not find anything missing during my initial review.

LKAS_ANGLE cars are currently:
- Subaru Forester 2022-24
- Subaru Outback 2023
- Subaru Ascent 2023


adeebshihadeh:
Needs safety verification. @jyoung8607 this may be a good candidate when you're around

PR #1198, opened 2024-08-29
Subaru: driver torque sensor Nm scaling factor and unit

martinl:
This PR adds Nm unit for Subaru driver torque sensor signal for gen1 and gen2 and updates carstate to use new signal. Matched with SSM4 signal output by to jnewb1 at https://github.com/martinl/openpilot/issues/80


jyoung8607 (changes requested):
It's good to have known correct units for the driver input torque signal. However:

1) You switched to a new signal; it's very important to reference the same signal in Panda safety
2) Having re-scaled the signal, you also need to re-scale the driver input threshold in openpilot
3) You will need corresponding scaling and driver input threshold updates to Panda

In short, this is the beginning of a valid change but needs additional work, including a companion Panda PR.

martinl:
Thanks for the comments, I will set up panda safety PR

martinl:
I added the scaling factor and panda PR is at https://github.com/commaai/panda/pull/2016

adeebshihadeh:
Can you open a new PR with the panda stuff all in one?

PR #1203, opened 2024-08-29
Honda long experiment

sshane:
0 accel is now 0 gas/brake


sshane:
Closing since we aren't planning on making this change just yet, and the gas response still isn't fully understood

PR #1221, opened 2024-09-04
add setuptools for python3.12

maxime-desroches:


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
   "crcmod",
   "tqdm",
   "pandacan@git+https://github.com/commaai/panda.git@master",
+  "setuptools",
```
adeebshihadeh, line 19: make sure to leave a comment why this is is necessary. afaik this is to fix an issue with one of our dependencies

PR #1223, opened 2024-09-05
GM: Car Port for 2019-20 GMC Yukon

garrettpall:
**Car**
GMC Yukon 2019-20 (Short and Long Wheel Base)

**Route**
TBD

**Checklist**
- [ ] added entry to CarInfo in opendbc/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [ ] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ] route with openpilot:
- [ ] route with stock system:
- [x] car harness used (if comma doesn't sell it, put N/A): GM / Camera

Note: Supersedes https://github.com/commaai/openpilot/pull/32107


opendbc/car/gm/values.py
```diff
@@ -245,7 +249,10 @@ class CanBus:
 EV_CAR = {CAR.CHEVROLET_VOLT, CAR.CHEVROLET_VOLT_2019, CAR.CHEVROLET_BOLT_EUV}
 
 # We're integrated at the camera with VOACC on these cars (instead of ASCM w/ OBD-II harness)
-CAMERA_ACC_CAR = {CAR.CHEVROLET_BOLT_EUV, CAR.CHEVROLET_SILVERADO, CAR.CHEVROLET_EQUINOX, CAR.CHEVROLET_TRAILBLAZER}
+CAMERA_ACC_CAR = {CAR.CHEVROLET_BOLT_EUV, CAR.CHEVROLET_SILVERADO, CAR.CHEVROLET_EQUINOX, CAR.CHEVROLET_TRAILBLAZER, CAR.GMC_YUKON}
+
+# Alt ASCMActiveCruiseControlStatus
+ALT_ACCS = {CAR.GMC_YUKON}
```
sshane, line 255: Can we detect this?

garrettpall: I think this would do it in interface.py:
`ret.networkLocation == NetworkLocation.fwdCamera and 0x370 not in fingerprint[CanBus.POWERTRAIN]`

What would be the best way to pass that to carstate?

sshane: You can utilize `CP.flags`

garrettpall: Longitudinal control is not the same, `ALT_ACCS` does not have OP long. I think these cars have a self contained radar module that send different gas/brake/status messages.

garrettpall:
This should work for Suburbans (looking at older OPGM code), and possibly Tahoes. Data also says should support 2016-2020 too, but looks to be un-tested.

garrettpall:
Oops, accidentally closed. Should have a route within a day or two

jyoung8607:
I'm reasonably happy with the state of the code, except I wish I knew a better label than ALT_ACCS, I don't know enough about the underlying GM architecture to name it something better.

The logs tell me the drive quality isn't very good, and our starting parameters aren't well matched to the car. We're not conforming very well to the desired lateral accelerations, the torque factor learner thinks our configured factor is way off, the params learner thinks your steering ratio is around 25 as opposed to the configured 17, etc.

Would you be able to help get some more test routes, and try some adjusted parameters? It might be good to consult with others experienced in the unique GM control issues to see if its behavior matches other cars we can emulate. 

If you're not in a position to iterate on the tune right away, we can just tag it as dashcam and merge anyway. I don't think this drives well enough (yet) to advertise as fully supported, but we can definitely get the upstream merge done, and we can just remove the flag whenever the tune is sorted.

PR #1228, opened 2024-09-08
GM: Add Stock FCW to CarState

garrettpall:
Route: 162796f1469f2f1b/00000001--df8f9ff62b

Screenshot 2024-09-07 at 9 29 32 PM


sshane:
Note that this may not capture all FCW events on all cars, it doesn't seem to work on our Bolt according to what I found: https://github.com/commaai/openpilot/pull/26114

garrettpall:
Added a comment. Let me know if that explains it well enough.

PR #1235, opened 2024-09-11
VW MQB: Added unittests for CRC calculation

JaCzekanski:
As discussed in #1016, this PR adds coverage or CRC calculation for currently supported messages.
6 tests are disabled as field definitions for these messages are not complete.
Those are: Getriebe_11, ESP_21, Motor_20, EV_Gearshift, SWA_01, ESP_20

Adding missing fields and changing _CRC, _BZ to CHECKSUM, COUNTER might be done as a separate PR or as a commit on top of this.

PR #1236, opened 2024-09-11
VW MQB: Refactored CRC calculation

JaCzekanski:
As discussed in #1016, this change get's rid of duplicate constants for messages that use single hex for all counter values.

Do not merge before #1235


jyoung8607:
@JaCzekanski In the abstract, I'm pretty happy with this. Your tests validated it worked within opendbc, and I did some crude profiling with 06d8f46710612c10600fbda2583ef0e963539b53 to verify there was no performance impact.

I was quite surprised to see it breaks when openpilot uses it, just for the CAN messages that didn't get full value initialization. I'm puzzled why this works in opendbc but not in openpilot. All is fine with 98b3385a77c8963a06e642c8f862a376b2ec8795.

Do you have any insight here? I'm guessing there's been evolution in std::array over the years and some part of the openpilot/panda toolchain doesn't like it. I'm actually happy with this PR as it stands, we can keep the explicit array initialization and still get 99% of the value I hoped for, but would like to solve the mystery if we can.

JaCzekanski:
@jyoung8607 Yeah, it looks like it doesn't work the way intended it to: https://godbolt.org/z/9xeEd44Gc
I'm fine with leaving all values filled - slightly less neat, but still more condense compared to the previous implementation.

jyoung8607:
Validation:

* Simple profiling with 06d8f46710612c10600fbda2583ef0e963539b53, performance is equivalent or very slightly faster
* New CI coverage from #1235, all checks pass
* Local testing with `pytest selfdrive/car/tests`, all checks pass
* Local testing with `selfdrive/tests/process_replay/test_processes.py`, no changes

PR #1237, opened 2024-09-12
VW MQB: EV_Gearshift -> Motor_EV_01

jyoung8607:
Updated message and signal data for 0x187 Motor_EV_01. This updates the signal name and corrects the signal width for the special gearshift signal we use on the e-Golf. The additional signals are a prerequisite for #1235.


jyoung8607:
Did a local manual test with `run_process_on_route.py` using an e-Golf route privately shared with me, LGTM.

We should really have a CI test route for e-Golf due to this unique code path. I'll ask some owners.

PR #1238, opened 2024-09-12
VW MQB: Various message and signal updates

jyoung8607:
Updates, fixes, and additions in support of #1235. The additional signals aren't used by openpilot, but are used by the car, and must be replicated by the CAN packer to recreate the car's messages for CI testing.


jyoung8607:
Validated locally:

* No breakage in `pytest selfdrive/car/tests`
* Null effect confirmed, no changes seen by `selfdrive/test/process_replay/test_processes.py`

PR #1242, opened 2024-09-13
VW MQB: Add missing ESP_33 messages and tests

JaCzekanski:
Added new CRC without coverage - oversight from rebasing #1016


jyoung8607:
ESP_33 will actually be useful for #1229, it contains FCW/AEB relevant feedback from stability control. Thanks!

PR #1243, opened 2024-09-13
VW MQB: Keep CAN IDs in order

JaCzekanski:
ACC_02 had address in decimal format and was out of order.


JaCzekanski:
Resolved conflicts

PR #1244, opened 2024-09-16
Toyota: add ENG2F41

sshane:
Contains signals that describe how much force is being applied for positive acceleration.

I think `FDRVREAL` is the more accurate version, as that includes these RPM drops from down shifting, and it also exhibits delayed response in some cases. Perhaps `FDRV` is more of a request or ideal value.

Lexus ES route: 57048cfce01d9625/000000f5--a32b2b7add

image

Here is each divided by CP.mass to get an acceleration (+ pitch compensated):

image

### DIFF
--- a/opendbc/dbc/generator/toyota/_toyota_2017.dbc
+++ b/opendbc/dbc/generator/toyota/_toyota_2017.dbc
@@ -45,6 +45,14 @@ BO_ 37 STEER_ANGLE_SENSOR: 8 XXX
  SG_ STEER_FRACTION : 39|4@0- (0.1,0) [-0.7|0.7] "deg" XXX
  SG_ STEER_RATE : 35|12@0- (1,0) [-2000|2000] "deg/s" XXX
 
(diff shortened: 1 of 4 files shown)


opendbc/dbc/generator/toyota/_toyota_2017.dbc
```diff
@@ -45,6 +45,14 @@ BO_ 37 STEER_ANGLE_SENSOR: 8 XXX
  SG_ STEER_FRACTION : 39|4@0- (0.1,0) [-0.7|0.7] "deg" XXX
  SG_ STEER_RATE : 35|12@0- (1,0) [-2000|2000] "deg/s" XXX
 
+BO_ 119 ENG2F41: 6 CGW
+ SG_ FDRV : 7|16@0- (2,0) [0|0] "N" Vector__XXX
+ SG_ FDRVREAL : 23|13@0- (10,0) [0|0] "N" Vector__XXX
```
sshane, line 50: probably driving force request and actual measured driving force

PR #1246, opened 2024-09-16
Toyota: add drive force signal for hybrids

sshane:
COUNTER is not a counter, unsure what it might be still.

Found factor by comparing with PCM_CRUISE->NEUTRAL_FORCE which is from the reference DBC.

The divergences here are when the PCM is applying some gas, when they are the same no extra gas is being commanded, but there exists engine braking/creeping forces still.

image


sshane:
Accurate even for user gas! `08e4c2a99df165b1/0000027e--b51d0c81a8/31:36`

image

PR #1248, opened 2024-09-17
Lexus ES TSS2: improve start from stop for ICE variants

sshane:


adeebshihadeh:
How's the report look before/after?

sshane:
It's actually noticeable in the low speed creep test and start from stop! You can see it no longer overshoots and even reaches the acceleration earlier.

[LEXUS_ES_TSS2_57048cfce01d9625_0000010.zip](https://github.com/user-attachments/files/17023452/LEXUS_ES_TSS2_57048cfce01d9625_0000010.zip)

sshane:
master:
image

pcm comp:
image

PR #1251, opened 2024-09-18
FCA Giorgio: Alfa Romeo Stelvio

jyoung8607:
Resuming work on commaai/openpilot#32898.

Users have driven an earlier version of this code on both Stelvio and Giulia and seem to be loving it. The lateral acceleration torque controller feed-forward is an excellent match for these cars. There are some obstacles we need to solve for official support; it's possible we'll have to ship it in dashcam mode.

**Checklist**

- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ] route with openpilot: `TBD`
- [ ] route with stock system: `TBD`
- [x] car harness used (if comma doesn't sell it, put N/A): VW_A

**To-Do**

- [x] commaai/opendbc#1250
- [x] ~~commaai/panda#1980~~
- [ ] UDS/VIN fingerprinting
- [ ] Locate gearshift position signal
- [ ] Locate driver gas signal
- [ ] Locate cruise button signals
- [ ] Locate driver seatbelt signal
- [ ] Locate BSM signals
- [ ] Verify turn signal correctly processed (comfort vs latched, blink vs continuous)
- [ ] Plot out and test known HUD messaging (lane lines, etc)
- [ ] Fix reported issue with lane changes to the right


opendbc/car/structs.py
```diff
@@ -433,6 +433,7 @@ class SafetyModel(StrEnum):
     volkswagenMqbEvo = auto()
     chryslerCusw = auto()
     psa = auto()
+    fcaGiorgio = auto()
```
sshane, line 436: You can merge this now to reserve it, capnp and structs should remain in sync

jyoung8607:
Jeep Renegade *almost* works on this port, we'll see what happens after some tweaks and updates. Renegade is derived from the [Fiat SCCS Wide](https://en.wikipedia.org/wiki/SCCS_platform#Small_Wide) platform which has quite a lot of heritage predating the FCA merger. May end up renaming the port/safety to Fiat if this works out, we'll have to see how UDS fingerprinting looks.

ckwebmaster-cell:
Would this port work on the 2024 Dodge Hornet? which is a tonale?

jyoung8607:
- [ ] FIXME: how am I passing all CI tests when I haven't even moved the safety code over yet??

ckwebmaster-cell:
Would this or the jeep renegade work for the Dodge hornet?  Which is basically a alfa Tonale?

jyoung8607:
> Would this or the jeep renegade work for the Dodge hornet? Which is basically a alfa Tonale?

It's possible, we really can't say without someone doing the work to try it.

opendbc/car/fca_giorgio/carcontroller.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if self.frame % self.CCP.STEER_STEP == 0:
+      if CC.latActive:
+        new_torque = int(round(actuators.torque * self.CCP.STEER_MAX))
+        apply_torque = apply_driver_steer_torque_limits(new_torque, self.apply_torque_last, CS.out.steeringTorque, self.CCP)
+      else:
+        apply_torque = 0
+
+      self.apply_torque_last = apply_torque
+      can_sends.append(fca_giorgiocan.create_steering_control(self.packer_pt, CanBus.pt, apply_torque, CC.latActive))
```
belm0, line 32: these refs to `CanBus.pt` etc. are bugs.  `pt` is an instance property, but `CanBus` is the class itself, so the result is a Python `<property>` object, and the msg sends will fail.

barchielli:
Hi @jyoung8607 — thanks for keeping this port alive.

I have a 2020 Alfa Romeo Giulia Veloce (2.0 petrol, RWD, European market, factory ACC and active lane keeping), and I would rather contribute than just ask. Three questions before I buy hardware.

**Giulia vs Stelvio.** The branch defines a single platform (ALFA_ROMEO_STELVIO_1ST_GEN), so a Giulia would run with Stelvio mass and wheelbase. Would you consider the branch usable on a Giulia today, or is the Stelvio the only car you would vouch for? You mention users having driven earlier code on a Giulia — did anything come back from them that is not already in the TODO list?

**Expected limitations.** From reading the code: lateral only with stock ACC keeping longitudinal; FW fingerprint still PLACEHOLDER; checksums and counters ignored pending the CRC final XOR; no driver gas signal; no cancel spam; open issue with lane changes to the right. Is that a fair summary, and is there anything EU-market specific I should expect?

**Harness.** CarHarness.vw_a is no longer orderable as a car harness in the shop (VW is J533 now), but "VW A" still exists as a separate harness connector. Is "car harness, Developer variant, plus VW A harness connector" the right order today? And for this branch, comma four or a used 3X?

**What I can contribute.** I can produce CAN logs and UDS FW versions from a European Giulia MY2020 with annotated events (cruise buttons, accelerator, brake, indicators, ACC engage and disengage, DNA modes), recorded to whatever spec is useful to you. That should at least let a real Giulia fingerprint replace the placeholder, and might help with the remaining NEW_MSG_ addresses and the CRC. I have a development team and can do the work — I just need to know what would actually help.

Happy to discuss any arrangement that makes it worth your time. Anything produced goes upstream.

PR #1253, opened 2024-09-18
Lexus ES TSS2: improve braking performance ICE variants

sshane:


sshane:
There is not a clear enough improvement (not as clear as https://github.com/commaai/opendbc/pull/1248) here to want to merge this.

Plus I believe that PCM_CRUISE->ACCEL_NET is a request by the PCM to the brakes, and ***not*** what kind of brake pressure is being applied. The CLUTCH->ACCEL_NET signal was useful for improving acceleration because it closely matched aEgo, effectively being what force the engine was applying, with its lag and windup in all.

PR #1260, opened 2024-09-20
Toyota: pitch compensate new long tune

sshane:
Only for positive accel (given pitch). Now doesn't apply too much gas going down hills, or too little going up them.

Lexus ES TSS2 routes:

This branch:
Downhill with PCM + pitch compensation: https://connect.comma.ai/57048cfce01d9625/00000123--c1523da2eb
Uphill with PCM + pitch compensation: https://connect.comma.ai/57048cfce01d9625/00000124--db7c9676fb

Current master:
Downhill with PCM compensation: https://connect.comma.ai/57048cfce01d9625/00000125--1231d30430
Uphill with PCM compensation: https://connect.comma.ai/57048cfce01d9625/00000126--47613d0ae6

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -107,7 +107,7 @@ class CarController(CarControllerBase):
     # For cars where we allow a higher max acceleration of 2.0 m/s^2, compensate for PCM request overshoot
     # TODO: validate PCM_CRUISE->ACCEL_NET for braking requests and compensate for imprecise braking as well
     if self.CP.flags & ToyotaFlags.RAISED_ACCEL_LIMIT and CC.longActive:
-      # calculate amount of acceleration PCM should apply to reach target given pitch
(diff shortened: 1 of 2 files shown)


sshane:
Report looks like a success! On uphills, integral (seen in `actuators.accel`) does not need to wind up as high to achieve the same longitudinal acceleration:

image

Same with start from stop, the compensation knows to go higher immediately, resulting in less integral windup:

image

As well as the creep test:

image

---

Now for the downhill plots. There's not much difference in the "start from stop" test, if any, but the creep test shows noticeably less overshoot:

image

And the gas step response shows the compensation requesting less acceleration and slightly less overshoot:

image

---

Reports:
[LEXUS_ES_TSS2_57048cfce01d9625_0000012.zip](https://github.com/user-attachments/files/17070159/LEXUS_ES_TSS2_57048cfce01d9625_0000012.zip)

PR #1261, opened 2024-09-20
Lexus ES TSS2: reduce braking overshoot

sshane:
Re-open of https://github.com/commaai/opendbc/pull/1253. Compensates using PCM negative acceleration (brake) request (not actual applied value). This means while we can't attempt to undo any lag/jerk limits in the PCM (which doesn't seem possible anyway), we can still undo the internal PCM windup when requesting brake.

---

PCM accel net signals lexus es

---

Lexus ES master: 57048cfce01d9625/00000130--7556579cf1
Lexus ES w/ this branch: 57048cfce01d9625/00000136--215ea510b2

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -104,8 +104,8 @@ class CarController(CarControllerBase):
                                                           lta_active, self.frame // 2, torque_wind_down))
 
     # *** gas and brake ***
-    # For cars where we allow a higher max acceleration of 2.0 m/s^2, compensate for PCM request overshoot
-    # and imprecise braking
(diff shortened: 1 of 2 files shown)


sshane:
Positive acceleration actions should be the same here. The creep test definitely shows an improvement on not overshooting the request.

Left is before this PR, right is this PR:

image

One odd thing is that I had the CLUTCH signal show that the gas was winding up from the negative accel range rather than jumping immediately to 0. This meant there was more delay switching from braking to gas. **This did happen on previous branches, so it's not a new issue with this branch however.**

Here's a graphical explanation of the two behaviors:

image

---

The -1 m/s^2 brake step response also looked better with less request overshoot:

image

Same for -4 m/s^2:

image

PR #1265, opened 2024-09-21
Toyota: lower PCM compensation rate limit

sshane:
Heavily noticeable on hybrids, and effects are unknown on ICE


sshane:
Potentially causing these spikes, looks like compensation can definitely be slowed a bit. Will verify tomorrow

image

PR #1266, opened 2024-09-24
Lexus: prevent lagged gas after heavy braking

sshane:
If you switch from a high negative aceleration request to a positive one, sometimes the gas has to "unwind" from the negative acceleration value. Which doesn't make sense, but is what happens.

If you rate limit your request or send zero for one frame, then it is fine and the gas starts from 0 immediately.

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -43,7 +43,6 @@ class CarController(CarControllerBase):
     self.distance_button = 0
 
     self.pcm_accel_compensation = 0.0
-    self.last_accel = 0.0
 
     self.packer = CANPacker(dbc_name)
     self.accel = 0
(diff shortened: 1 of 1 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -158,6 +158,10 @@ def update(self, CC, CS, now_nanos):
       if pcm_cancel_cmd and self.CP.carFingerprint in UNSUPPORTED_DSU_CAR:
         can_sends.append(toyotacan.create_acc_cancel_command(self.packer))
       elif self.CP.openpilotLongitudinalControl:
+        # internal PCM gas command can get stuck at high negative values. send one frame of zero when transitioning from braking to gas
```
sshane, line 161:
```suggestion
        # internal PCM gas command can get stuck unwinding from negative accel. send one frame of zero when transitioning from braking to gas
```

PR #1267, opened 2024-09-24
body: remove orientation/angular velocity checks

fredyshox:


adeebshihadeh:
If these are still lists, this is still necessary. (This is why I wanted to remove lists in `livePose`)

fredyshox:
why does this check exist? those values are not used anywhere and this is the only carController referencing them

fredyshox:
@adeebshihadeh from what i see, this is a leftover from self-balacing code: https://github.com/commaai/openpilot/blob/16ec56398c3b23db917eb0f9474aa9f6ed2a459c/selfdrive/car/body/carcontroller.py#L48 so can be deleted

adeebshihadeh:
@sshane let's deprecate from carControl then?

PR #1271, opened 2024-09-24
Lexus ES TSS2: compensate for creeping force

sshane:
Seems to have a big effect at low speeds, we currently don't brake enough. Verify with https://github.com/commaai/openpilot/pull/33637

57048cfce01d9625/0000017b--93de962979/1

Before:

image

After:

image


sshane:
This doesn't seem to fix the undershooting on braking from 20 to 5 mph, but that is possibly another issue.

sshane:
Come to stop maneuvers overshoot much less on the braking:

image

image

image

Full reports: [LEXUS_ES_TSS2_57048cfce01d9625_000001d.zip](https://github.com/user-attachments/files/17158195/LEXUS_ES_TSS2_57048cfce01d9625_000001d.zip)

PR #1278, opened 2024-09-24
Car Interfaces: Refactor button state tracking with `ButtonTracker`

sunnyhaibin:
**Description**

- Introduce `ButtonTracker` class to encapsulate button state tracking and button event creation
- Move `create_button_events` function into `ButtonTracker` class
- Simplify car state updates by reducing button event handling to a single line

### DIFF
--- a/opendbc/car/__init__.py
+++ b/opendbc/car/__init__.py
@@ -35,7 +35,7 @@ def apply_hysteresis(val: float, val_steady: float, hyst_gap: float) -> float:
 
 class ButtonTracker:
   def __init__(self):
-    self.prev_btn = None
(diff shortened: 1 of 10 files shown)


adeebshihadeh:
Nice, I think we want to go a bit further here and change the fields to bools or enums in `carState`, e.g. `carState.followDistancePressed`. The list is pretty annoying.

sshane:
It should be a struct inside carState, CS.buttons.lkasButton, etc.

adeebshihadeh:
Reverted for now. There's a process replay difference after this: https://github.com/commaai/openpilot/actions/runs/11134269190/job/30942141454?pr=33690.

PR #1131, opened 2024-08-20
HKG: Car Port for Genesis GV70 Electrified 2023 (HDA2)

sunnyhaibin:
**Checklist**

- [x] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: `afe09b9f5d3f3548/00000011--15fefe1c50`
- [x] route with openpilot longitudinal: `afe09b9f5d3f3548/0000001b--a1129a4a15` 
- [x] route with stock system: `afe09b9f5d3f3548/00000005--fe8cd35e57`
- [x] car harness used (if comma doesn't sell it, put N/A): `Harness Q`

Thanks to community 2023 Genesis GV70 Electrified (HDA2) owner gv70e_81046 (Discord).

**Requires**
- https://github.com/commaai/opendbc/pull/1280

Superseded https://github.com/commaai/openpilot/pull/32765

PR #1280, opened 2024-09-25
Hyundai CAN-FD: Use `ACCELERATOR` for EV gear shifter positions

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/opendbc/pull/1131

**Related**
- https://github.com/commaai/openpilot/pull/33647


jyoung8607:
Offline validation:

* LGTM against commaCarSegments, commaai/openpilot#33647
* `pytest opendbc` passes
* `pytest selfdrive/car/tests` passes
* `selfdrive/test/process_replay/test_processes.py` which has an EV6 route, no changes

PR #1283, opened 2024-09-25
Hyundai CAN-FD Longitudinal: Enable for radar-based SCC platforms (non-HDA2)

sunnyhaibin:
**Supersedes**
- https://github.com/commaai/openpilot/pull/30033

**Requires**
- https://github.com/commaai/panda/pull/1668


jyoung8607:
@sunnyhaibin this one causes five HKG cars to fail openpilot test_models, and it's not immediately obvious why, so setting this back in Draft for you to look at.

sunnyhaibin:
> @sunnyhaibin this one causes five HKG cars to fail openpilot test_models, and it's not immediately obvious why, so setting this back in Draft for you to look at.

@jyoung8607 Found the root cause:
- Fix: https://github.com/sunnypilot/opendbc/commit/38e81d052f5409e56dcdc8e5168f53b2fa2ac134
- `test_models` validation: https://github.com/commaai/openpilot/actions/runs/13493760451

sunnyhaibin:
Do we want this merged first?
- https://github.com/commaai/opendbc/pull/1828

PR #1133, opened 2024-08-20
HKG: Car Port for Genesis GV70 2022-23 (HDA2)

sunnyhaibin:
**Checklist**

- [x] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot:
  - `f1b6af6fdf6c0a78`: `f1b6af6fdf6c0a78/00000004--5a6db3a9bd`
  - `1159e79fa5ec2cb0`: `1159e79fa5ec2cb0/0000000a--0dd85e8bd2`
- [x] route with stock system: `afe09b9f5d3f3548/00000005--fe8cd35e57`
- [x] car harness used (if comma doesn't sell it, put N/A): `N/A` (See below)
  - 2.5T, HDA2
    - Harness P
  - 3.5T, HDA2
    - Harness S: https://github.com/commaai/opendbc/issues/1297

Thanks to the community 2022 Genesis GV70 3.5T owners taylababygirl (Discord) and animatronichobbes.

**Offline Validation**
- [x] `pytest opendbc`
- [ ] `pytest selfdrive/car/tests`
- [ ] `selfdrive/test/process_replay/test_processes.py`

Superseded https://github.com/commaai/openpilot/pull/31704

PR #1132, opened 2024-08-20
HKG: Car Port for Genesis G80 2024 (2.5T, HDA2)

sunnyhaibin:
**Checklist**

- [x] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: `66eaa6c3b6b2afc6/00000009--3a5199aabe`
- [x] route with stock system: `66eaa6c3b6b2afc6/00000006--1fb9d3a187`
- [x] car harness used (if comma doesn't sell it, put N/A): Harness P
- [x] car architecture: CAN-FD

Superseded https://github.com/commaai/openpilot/pull/32633

PR #1284, opened 2024-09-25
Hyundai CAN-FD: Universal gear checks

sunnyhaibin:
**Description**

The refactor's goal is to improve the gear checking logic to be applicable to both HDA2 and non-HDA2 Hyundai vehicles, which was previously only applied to non-HDA2 cars.

**Verification**

This is tested in the following active PRs:
- https://github.com/commaai/opendbc/pull/1132
- https://github.com/commaai/opendbc/pull/1133

**Offline Validation**
- [x] `pytest opendbc`
- [x] `pytest selfdrive/car/tests`
- [x] `selfdrive/test/process_replay/test_processes.py`

**Supersedes**
- https://github.com/commaai/openpilot/pull/32732


sunnyhaibin:
**Offline Validation**
- [x] `pytest opendbc`
- [x] `pytest selfdrive/car/tests`
- [x] `selfdrive/test/process_replay/test_processes.py`

jyoung8607:
This does check out enough to merge, but as a future TODO, there's further simplification possible. According to commaCarSegments, there are no CAN-FD cars that truly depend on 0x130 GEAR_SHIFTER. Every car that has it, has one of the other messages too.

PR #1285, opened 2024-09-25
Hyundai CAN-FD: Add CAN fingerprint fallback checks for HDA2 detection

sunnyhaibin:
**Description**

Add fallback detection for HDA2 cars using CAM bus fingerprinting when ADAS Driving ECU response is absent.

**Prerequisite of**
- https://github.com/commaai/opendbc/pull/1132
- https://github.com/commaai/opendbc/pull/1133


opendbc/car/hyundai/interface.py
```diff
@@ -25,7 +25,17 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experime
     # FIXME: the Optima Hybrid 2017 uses a different SCC12 checksum
     ret.dashcamOnly = candidate in {CAR.KIA_OPTIMA_H, }
 
-    hda2 = Ecu.adas in [fw.ecu for fw in car_fw]
+    # Check if the ADAS Driving ECU is present in the firmware response.
+    # If not found, check the CAM CAN bus fingerprint for known HDA2 steering messages (0x50 or 0x110),
+    # that indicate an HDA2 car even without a direct ADAS Driving ECU response.
```
jyoung8607, line 30: Is the CAM CAN bus check reliable in all cases? If so, we should dispense with the FW check entirely.

sunnyhaibin: Yes, these two steering messages only exist on HDA2 cars, and we only care about if they are on bus 2 (tx).

opendbc/car/hyundai/interface.py
```diff
@@ -25,7 +25,17 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experime
...
+    if Ecu.adas in [fw.ecu for fw in car_fw]:
+      hda2 = True
+    else:
+      # Some HDA2 cars do not respond ADAS Driving ECU if the known HDA2 steering messages are present.
+      # The hda2 flag to initialize CanBus is not critical here since we only care about CAM bus steering messages.
```
sshane, line 35: I don't think we've ever verified what is sending LFA here btw. The ADAS could actually exist but behind a new gateway (a FW scan on OBD port should show it), or it could actually not exist and radar might take over this function.

opendbc/car/hyundai/interface.py
sshane: Yes, we should only have one source

jyoung8607: @sunnyhaibin setting to draft while you make that update, I'll validate when you're finished.

sunnyhaibin: @jyoung8607 Updated with 1 source only and unit test

opendbc/car/hyundai/tests/test_hyundai.py
```diff
@@ -44,17 +45,20 @@
...
-      car_fw = [CarParams.CarFw(ecu=Ecu.adas if has_adas else Ecu.fwdCamera)]
-      CP = CarInterface.get_params(CAR.KIA_EV6, gen_empty_fingerprint(), car_fw, False, False)
-      assert bool(CP.flags & HyundaiFlags.CANFD_HDA2) == has_adas
+    for hda2 in (True, False):
+      fingerprint = gen_empty_fingerprint()
+      if hda2:
+        cam_can = CanBus(None, False, fingerprint).CAM
+        fingerprint[cam_can] = [0x50, 0x110]  # HDA2 steering messages
+      CP = CarInterface.get_params(CAR.KIA_EV6, fingerprint, [], False, False)
+      assert bool(CP.flags & HyundaiFlags.CANFD_HDA2) == hda2
```
jyoung8607, line 54: I like that you added a test! Can't do this yet, but for the first car we merge that uses the new CAN fallback, let's duplicate the EV6 check with that car, as an example without the ADAS ECU.

opendbc/car/hyundai/tests/test_hyundai.py
```diff
@@ -44,17 +45,20 @@
...
 
     # radar available
     for radar in (True, False):
       fingerprint = gen_empty_fingerprint()
       if radar:
         fingerprint[1][RADAR_START_ADDR] = 8
-      CP = CarInterface.get_params(CAR.HYUNDAI_SONATA, fingerprint, car_fw, False, False)
+      CP = CarInterface.get_params(CAR.HYUNDAI_SONATA, fingerprint, [], False, False)
```
jyoung8607, line 61: Why this change?

jyoung8607 (changes requested):

opendbc/car/hyundai/tests/test_hyundai.py
sunnyhaibin: `car_fw` was only relevant for the `hda2` feature detection test. Since the `car_fw` is no longer relevant for determining HDA2, we should be able to just set an empty list here.

opendbc/car/hyundai/tests/test_hyundai.py
sunnyhaibin: I like this idea. Will implement.

opendbc/car/hyundai/interface.py
```diff
@@ -18,7 +18,11 @@ class CarInterface(CarInterfaceBase):
   def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experimental_long, docs) -> structs.CarParams:
     ret.carName = "hyundai"
 
-    hda2 = Ecu.adas in [fw.ecu for fw in car_fw]
+    # Check the CAM bus fingerprint for known HDA2 steering messages (0x50 or 0x110) to determine if it is an HDA2 car.
+    # The hda2 flag to initialize CanBus is not critical here since we only care about CAM bus steering messages.
+    cam_can = CanBus(None, False, fingerprint).CAM
+    hda2 = 0x50 in fingerprint[cam_can] or 0x110 in fingerprint[cam_can]
+
```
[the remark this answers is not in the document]
sunnyhaibin: Updated https://github.com/commaai/opendbc/pull/1285/commits/7b8f46c9ffe558ec7d05bb71a0e7e0b82c3d6b42

opendbc/car/hyundai/interface.py
```diff
@@ -18,8 +18,11 @@ class CarInterface(CarInterfaceBase):
   def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experimental_long, docs) -> structs.CarParams:
     ret.carName = "hyundai"
 
-    hda2 = Ecu.adas in [fw.ecu for fw in car_fw]
-    CAN = CanBus(None, hda2, fingerprint)
+    # Check the CAM bus fingerprint for known HDA2 steering messages (0x50 or 0x110) to determine if it is an HDA2 car.
+    cam_can = CanBus(None, fingerprint).CAM
+    hda2 = 0x50 in fingerprint[cam_can] or 0x110 in fingerprint[cam_can]
+
+    CAN = CanBus(None, fingerprint, hda2)
```
jyoung8607, line 25: Code is now self-explanatory and groups together logically.

```suggestion
    cam_can = CanBus(None, fingerprint).CAM
    hda2 = 0x50 in fingerprint[cam_can] or 0x110 in fingerprint[cam_can]
    CAN = CanBus(None, fingerprint, hda2)
```

sunnyhaibin:
**Offline Validation**
- [x] `pytest opendbc`
- [x] `pytest selfdrive/car/tests`
- [x] `selfdrive/test/process_replay/test_processes.py`

jyoung8607:
Looks good! PR review bot fail looks to be a Docker outage.

PR #1286, opened 2024-09-26
docs: CARS.md 2.0

### LINKED ISSUE #1144: New car support docs
This is v2 of [CARS.md](docs/CARS.md). Currently, CARS.md shows all the cars supported by upstream openpilot. The goal of this new page is to be a one stop shop to see the status of any car's support.

 It'll include:
* every single car: cars supported in upstream, dashcam cars in upstream, cars supported in forks (even if they're forever in forks), cars we know won't work (no ACC/LKAS), cars with blockers (SecOc), etc.
* super granular columns: has fuzzy fingerprinting, steering torque, etc.
* some info will be populated by openpilot, and the rest will be done manually

This will be super information dense, so I imagine we'll need to move this to a webpage instead of a GitHub Markdown file.

These docs should also help make it clear when a car port is ready to supported and merged into upstream openpilot.

Related:
* commaai/openpilot#27320
* commaai/openpilot#26384

jyoung8607:
Resolves #1144 by introducing `other_cars.py`, allowing us to generate CarDocs for cars outside of openpilot. This is WIP and still a bit trash, but illustrative enough to start the bikeshedding.

**Direct links to view the work in progress**
* CarDocs for community supported and incompatible cars, [other_cars.py](https://github.com/jyoung8607/opendbc/blob/extra-cars/opendbc/car/other_cars.py)
* New generated [CARS.md](https://github.com/jyoung8607/opendbc/blob/extra-cars/opendbc/car/docs/CARS.md)

**TODO**
- [ ] Move generation to opendbc, but save output in openpilot (?)
  - [ ] Unwind development-only move of CARS.md and assets just before shipping
- [ ] Verify this doesn't hurt any comma.ai/vehicles or shop.comma.ai flows
- [ ] Make other-than-official cars much more visually obvious (background color, italics, ?)
- [ ] Cleanup/revamp the template text in accordance with the New Way of Things 
- [ ] Remove placeholder "dashcam mode" support logic, replace with the actual reason they're in dashcam mode
- [ ] Out-of-tree cars should show N/A for other characteristics
- [x] Some complex footnotes should become intra-document #links
- [ ] Internal discussion about links to forks
- [ ] Resolve other TODOs
- [ ] Cleanup
- [ ] Tests for new support type logic


opendbc/car/docs_definitions.py
```diff
@@ -24,6 +25,16 @@ class Column(Enum):
   VIDEO = "Video"
 
 
+# TODO: Bikeshed this enum, lines between custom fork and dashcam are blurry, and not sure about legacy support
+class SupportType(Enum):
+  OFFICIAL = "Official"
```
adeebshihadeh, line 30: "official" is a pretty loaded word, let's go with "upstream"

adeebshihadeh (commented):
`other_cars.py` looks great!

PR #1287, opened 2024-09-26
Toyota: 2021-23 RAV4 Prime

jyoung8607:
Credit to @pd0wm, just bringing commaai/openpilot#31179 current after the refactor.

- [x] Find a way to pass the SecOC key down from openpilot
- [x] Don't log SecOC parameter (see mapbox token implementation)
- [x] UI warning for missing SecOC key
- [x] Make alt messages common between SecOC cars? 
- [x] Look at remaining TODOs from Willem's work
- [x] Move SecOC implementation to a common area
- [ ] ~~Tests for SecOC? TBD~~ later

TODO in a future PR:
* Refactor SecOC module API to handle reset/trip/counter state internally
* Allow storing multiple SecOC keys, by VIN


opendbc/dbc/generator/toyota/toyota_rav4_prime.dbc
```diff
@@ -0,0 +1,470 @@
...
+ SG_ RESET_CNT : 23|20@0+ (1,0) [0|65535] "" XXX
+ SG_ AUTHENTICATOR : 35|28@0+ (1,0) [0|268435455] "" XXX
+
+BO_ 37 STEER_ANGLE_SENSOR: 8 XXX
+ SG_ STEER_ANGLE : 3|12@0- (1.5,0) [-500|500] "deg" XXX
+ SG_ STEER_RATE : 35|12@0- (1,0) [-2000|2000] "deg/s" XXX
+ SG_ STEER_FRACTION : 39|4@0- (0.1,0) [-0.7|0.7] "deg" XXX
+
+BO_ 170 WHEEL_SPEEDS: 8 XXX
+ SG_ WHEEL_SPEED_FR : 7|16@0+ (0.01,-67.67) [0|250] "km/h" XXX
```
MonsterPerformance, line 48: either speed limits or offset is wrong

opendbc/dbc/generator/toyota/toyota_rav4_prime.dbc
```diff
@@ -0,0 +1,470 @@
...
+ SG_ STEER_REQUEST_2 : 0|1@0+ (1,0) [0|1] "" XXX
+ SG_ STEER_REQUEST_1 : 3|1@0+ (1,0) [0|1] "" XXX
+ SG_ COUNTER : 13|6@0+ (1,0) [0|63] "" XXX
+ SG_ STEER_ANGLE_CMD : 23|16@0- (0.0573,0) [0|65535] "" XXX
+ SG_ AUTHENTICATOR : 35|28@0+ (1,0) [0|268435455] "" XXX
+ SG_ RESET_FLAG : 37|2@0+ (1,0) [0|3] "" XXX
+ SG_ MSG_CNT_LOWER : 39|2@0+ (1,0) [0|3] "" XXX
+
+BO_ 353 DSU_SPEED: 7 XXX
+ SG_ FORWARD_SPEED : 15|16@0- (0.00390625,-30) [0|255] "km/h" XXX
```
MonsterPerformance, line 80: either speed limits or offset is wrong

opendbc/dbc/toyota_rav4_prime_generated.dbc
```diff
@@ -0,0 +1,473 @@
...
+ SG_ RESET_CNT : 23|20@0+ (1,0) [0|65535] "" XXX
+ SG_ AUTHENTICATOR : 35|28@0+ (1,0) [0|268435455] "" XXX
+
+BO_ 37 STEER_ANGLE_SENSOR: 8 XXX
+ SG_ STEER_ANGLE : 3|12@0- (1.5,0) [-500|500] "deg" XXX
+ SG_ STEER_RATE : 35|12@0- (1,0) [-2000|2000] "deg/s" XXX
+ SG_ STEER_FRACTION : 39|4@0- (0.1,0) [-0.7|0.7] "deg" XXX
+
+BO_ 170 WHEEL_SPEEDS: 8 XXX
+ SG_ WHEEL_SPEED_FR : 7|16@0+ (0.01,-67.67) [0|250] "km/h" XXX
```
MonsterPerformance, line 51: either speed limits or offset is wrong

opendbc/dbc/toyota_rav4_prime_generated.dbc
```diff
@@ -0,0 +1,473 @@
...
+ SG_ STEER_REQUEST_2 : 0|1@0+ (1,0) [0|1] "" XXX
+ SG_ STEER_REQUEST_1 : 3|1@0+ (1,0) [0|1] "" XXX
+ SG_ COUNTER : 13|6@0+ (1,0) [0|63] "" XXX
+ SG_ STEER_ANGLE_CMD : 23|16@0- (0.0573,0) [0|65535] "" XXX
+ SG_ AUTHENTICATOR : 35|28@0+ (1,0) [0|268435455] "" XXX
+ SG_ RESET_FLAG : 37|2@0+ (1,0) [0|3] "" XXX
+ SG_ MSG_CNT_LOWER : 39|2@0+ (1,0) [0|3] "" XXX
+
+BO_ 353 DSU_SPEED: 7 XXX
+ SG_ FORWARD_SPEED : 15|16@0- (0.00390625,-30) [0|255] "km/h" XXX
```
MonsterPerformance, line 83: either speed limits or offset is wrong

opendbc/dbc/toyota_rav4_prime_generated.dbc
jyoung8607: Do you have feedback on which, and how?

jyoung8607:
This should be G2G after the pandacan reference becomes available to update.

opendbc/car/toyota/carcontroller.py
```diff
@@ -59,6 +64,18 @@ def update(self, CC, CS, now_nanos):
...
+    if self.CP.flags & ToyotaFlags.SECOC.value:
+      if CS.secoc_synchronization['RESET_CNT'] != self.secoc_prev_reset_counter:
+        self.secoc_lka_message_counter = 0
+        self.secoc_lta_message_counter = 0
+        self.secoc_prev_reset_counter = CS.secoc_synchronization['RESET_CNT']
+
+        # Verify mac of SecOC synchronization message to ensure we have the right key
+        # TODO: move this to tests rather than runtime
+        # expected_mac = build_sync_mac(self.secoc_key, int(CS.secoc_synchronization['TRIP_CNT']), int(CS.secoc_synchronization['RESET_CNT']))
+        # assert(int(CS.secoc_synchronization['AUTHENTICATOR']) == expected_mac)
```
pd0wm, line 77: If you want to show an alert for having the wrong SecOC key this would be great place to detect that.

opendbc/car/toyota/carstate.py
```diff
@@ -79,7 +82,19 @@ def update(self, cp, cp_cam, *_) -> structs.CarState:
     ret.brakePressed = cp.vl["BRAKE_MODULE"]["BRAKE_PRESSED"] != 0
     ret.brakeHoldActive = cp.vl["ESP_CONTROL"]["BRAKE_HOLD_ACTIVE"] == 1
 
-    ret.gasPressed = cp.vl["PCM_CRUISE"]["GAS_RELEASED"] == 0
+    if self.CP.flags & ToyotaFlags.SECOC.value:
+      self.secoc_synchronization = copy.copy(cp.vl["SECOC_SYNCHRONIZATION"])
+      ret.gas = cp.vl["GAS_PEDAL"]["GAS_PEDAL_USER"]
```
pd0wm, line 87: `ret.gas` seems to have been removed for the other cars, so probably can go here as well.

opendbc/car/toyota/carcontroller.py
jyoung8607: Gave it some thought, but at the time I was trying to get the minimum viable port running and we can't use cloudlog here anymore. What we'd have to do is add a CarState counter, and then add some logic in openpilot `card` to surface an alert after some number of violations.

Right now the cars go to LKA fault on a missing key, but we should be showing the dashcam no-security-key startup alert on top of it. For an actually wrong key, that could use detection eventually.

Compromise for now: I'll add the CarState counter without the event logic. Quick, easy, it'll help us troubleshoot if the problem ever comes pu, and we'll get some data from the field on how to set alerting thresholds if we decide it's needed.

opendbc/car/toyota/carstate.py
jyoung8607: Not sure if it was removed, or just never added for Toyota, I'd have to walk through git blame. However, every other car port is setting `ret.gas`. ISTM we shouldn't remove it for SecOC, but (in a separate PR) add it for non-SecOC.

pd0wm: I think it was removed here https://github.com/commaai/openpilot/pull/31903/files#diff-94720bc1c4f1ce01d86ce63b3c8278d1a201cf98e0796d5db9f1c0081d5a8a15.

The `ret.gas` still being present for the RAV4 prime is probably because my PR predates that commit.

jyoung8607: Looks like it was lost in commaai/openpilot#30194, and it's still desirable to put it back someday once the right logic is found. commaCarDatasets actually makes that a lot easier.

I think it makes sense to keep it in this port.

opendbc/car/toyota/carcontroller.py
sshane: We forward logs from carlog to cloudlog: https://github.com/commaai/opendbc/blob/739117d561b265de5caf843c3f117490a9de57e6/opendbc/car/__init__.py#L14-L17

jyoung8607: sweet, will implement

jyoung8607:
@sshane What do you think about capping MAC fail logging [like this](https://github.com/commaai/opendbc/pull/1287/commits/185e14b9c88cceed17a90b846b1bc5c839dc4bfb)? Seems unwise to cloudlog fails at 100+/sec indefinitely when someone starts up with a wrong/missing key, but not sure we have similar limits for counter/checksum fails.

sshane:
Seems fine

PR #1291, opened 2024-09-26
Lexus ES TSS2: compensate while stopping

sshane:
to prevent unrecoverable brake windup. partially reverts https://github.com/commaai/opendbc/pull/1264


sshane:
I don't see nearly as much of a difference as I did before, testing after https://github.com/commaai/opendbc/pull/1271. This is potentially because we no longer underbrake coming to a stop, which is transitioned more quickly to a high brake request in the stopping state.

sshane:
Superseded by https://github.com/commaai/opendbc/pull/1301

PR #1293, opened 2024-09-27
HKG: Add FW version and update model year for Genesis GV60 2022

sunnyhaibin:
Add firmware versions for the Genesis GV60 (Performance Trim) 2022. Expand supported model-year range down to 2022.

**Route ID**
- Stock longitudinal: `b26e6868ffe5ea08/00000000--946587db7a`
- openpilot longitudinal: `b26e6868ffe5ea08/0000000a--64b0ced90d`

Thanks to the community 2022 Genesis GV60 (Performance Trim) bigal_00666 (Discord).

### DIFF
--- a/opendbc/car/hyundai/values.py
+++ b/opendbc/car/hyundai/values.py
@@ -498,7 +498,7 @@ class CAR(Platforms):
   GENESIS_GV60_EV_1ST_GEN = HyundaiCanFDPlatformConfig(
     [
       HyundaiCarDocs("Genesis GV60 (Advanced Trim) 2023", "All", car_parts=CarParts.common([CarHarness.hyundai_a])),
-      HyundaiCarDocs("Genesis GV60 (Performance Trim) 2023", "All", car_parts=CarParts.common([CarHarness.hyundai_k])),
+      HyundaiCarDocs("Genesis GV60 (Performance Trim) 2022-23", "All", car_parts=CarParts.common([CarHarness.hyundai_k])),
     ],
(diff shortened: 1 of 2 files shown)


opendbc/car/hyundai/values.py
```diff
@@ -498,7 +498,7 @@ class CAR(Platforms):
   GENESIS_GV60_EV_1ST_GEN = HyundaiCanFDPlatformConfig(
     [
       HyundaiCarDocs("Genesis GV60 (Advanced Trim) 2023", "All", car_parts=CarParts.common([CarHarness.hyundai_a])),
-      HyundaiCarDocs("Genesis GV60 (Performance Trim) 2023", "All", car_parts=CarParts.common([CarHarness.hyundai_k])),
+      HyundaiCarDocs("Genesis GV60 (Performance Trim) 2022-23", "All", car_parts=CarParts.common([CarHarness.hyundai_k])),
     ],
     CarSpecs(mass=2205, wheelbase=2.9, steerRatio=12.6),  # steerRatio: https://www.motor1.com/reviews/586376/2023-genesis-gv60-first-drive/#:~:text=Relative%20to%20the%20related%20Ioniq,5%2FEV6%27s%2014.3%3A1.
```
jyoung8607, line 503: image

This change will also affect the existing 2023 GV60, but I pulled its original test route, and it was at 16.7 and still rising toward the end, so this should be a good change for both.

```suggestion
    CarSpecs(mass=2205, wheelbase=2.9, steerRatio=17.6),
```

jyoung8607 (changes requested):

jyoung8607 (approved):

PR #1311, opened 2024-10-01
support for SecOC cars

jyoung8607:
Separating the SecOC support code from the RAV4 Prime port.


opendbc/car/structs.py
```diff
@@ -392,6 +392,13 @@ class LateralTorqueTuning:
 
   wheelSpeedFactor: float = auto_field()  # Multiplier on wheels speeds to computer actual speeds
 
+  securityConfig: 'CarParams.SecurityConfig' = field(default_factory=lambda: CarParams.SecurityConfig())
+
+  @auto_dataclass
+  class SecurityConfig:
```
adeebshihadeh, line 398: any reason to nest these fields in this struct?

jyoung8607: I felt like the list of SecOC knobs might grow. Other potential paths have been talked about for onboarding other SecOC cars, like patching out HSMs and using a static/zero key, etc. Mild preference for grouping them but if you don't think there will be parameter growth, I can take that out.

adeebshihadeh: yeah, let's unnest

PR #1318, opened 2024-10-02
Add lowSpeedAlert for dynamic minimum steering speed

sshane:


opendbc/car/structs.py
```diff
@@ -83,6 +83,7 @@ class CarState:
   carFaultedNonCritical: bool = auto_field()  # some ECU is faulted, but car remains controllable
   espActive: bool = auto_field()
   vehicleSensorsInvalid: bool = auto_field()  # invalid steering angle readings, etc.
+  lowSpeedAlert: bool = auto_field()
```
sshane, line 86: I see this being useful as a flag to signal that the car is under the minimum steering speed for cars where this is slightly dynamic, then openpilot can just add an alert based on that.

sshane:
Done in https://github.com/commaai/opendbc/pull/1334

Issue #1309, opened 2024-10-01
Build generated DBCs with SCons


adeebshihadeh:
Two steps:
* remove from git tracking
* build them with SCons

LouisVelez:
Are bounties closed manually or did I mess something with the tagging on the PR?

adeebshihadeh:
If you add "closes #<issue_number>" to the PR next time, it should close automatically.

PR #1329, opened 2024-10-02
Scons generate dbc files

bongbui321:
resolve https://github.com/commaai/opendbc/issues/1309


SConstruct
```diff
@@ -74,3 +74,4 @@ envCython["LIBS"] = python_libs
 Export('envCython')
 
 SConscript(['opendbc/can/SConscript'])
+SConscript(['opendbc/dbc/generator/SConscript'])
```
adeebshihadeh, line 77: these should go in a wrapper SConscript so openpilot only has to call one. see how the other submodules do it

opendbc/dbc/generator/SConscript
```diff
@@ -0,0 +1,4 @@
+Import('env')
+
+envDBC = env.Clone()
+envDBC.Command(None, "generator.py", "python3 $SOURCE")
```
adeebshihadeh, line 4: need to specify the dependencies and build products, otherwise it doesn't know when to build this (such as file changes)

adeebshihadeh (changes requested):

opendbc/dbc/generator/SConscript
```diff
@@ -0,0 +1,18 @@
+import os
+from opendbc.dbc.generator.generator import (generated_suffix, generator_path, opendbc_root, \
+                                  get_python_generator_scripts, get_source_dbc_files)
+Import('env')
+
```
adeebshihadeh, line 5: this looks way too complicated. why not just glob everything in generator/*?

opendbc/dbc/generator/generator.py
```diff
@@ -39,25 +40,36 @@ def create_dbc(src_dir: str, filename: str, output_path: str):
...
-  for f in glob.glob(f"{output_path}/*{generated_suffix}"):
-    os.remove(f)
+def get_python_generator_scripts(generator_path: str) -> list[str]:
+  return glob.glob(f"{generator_path}/*/*.py")
 
-  # run python generator scripts first
-  for f in glob.glob(f"{generator_path}/*/*.py"):
-    subprocess.check_call(f)
+
+def get_source_dbc_files(generator_path: str) -> defaultdict[str, list[str]]:
```
adeebshihadeh, line 47: same here, shouldn't need to touch this script

opendbc/dbc/generator/SConscript
```diff
@@ -0,0 +1,18 @@
...
+dbc_sources = []
+generated_files = []
+generator_scripts = get_python_generator_scripts(generator_path)
+for src_dir, filenames in get_source_dbc_files(generator_path).items():
+  for filename in filenames:
+    dbc_sources.append(os.path.join(src_dir, filename))
+    output_file = filename.replace('.dbc', generated_suffix)
+    generated_files.append(os.path.join(opendbc_root, output_file))
+
+envDBC = env.Clone()
```
adeebshihadeh, line 15: no need to clone since nothing's changed about the env

adeebshihadeh (changes requested):

opendbc/dbc/generator/SConscript
bongbui321: not all `.dbc` within generator dir being the sources for the generated DBCs :(https://github.com/commaai/opendbc/blob/8cf033fa3ee6191c10e5d63702f8e3211095a29b/opendbc/dbc/generator/generator.py#L57)

glob.glob(f"{generated_path/*/*.dbc}" with added conditional statements and replacing the `.dbc` suffix with `_generated.dbc` would be the same as what you referenced

it's easier abstract out the search functions to get the path to generated dbc to delete those when run with `scons -c`

PR #1333, opened 2024-10-02
docs: finish CarDocs migration to opendbc

jyoung8607:
Create opendbc/car/docs, gather docs-related stuff there. Prereq for #1286.

CARS.md will continue to live in the same place in openpilot.

- [ ] comma.ai/vehicles impact understood
- [ ] shop.comma.ai impact understood


opendbc/car/docs/generate_docs.py
```diff
@@ -1,14 +1,27 @@
...
-from opendbc.car.docs_definitions import CarDocs, Column, CommonFootnote, PartType
+from opendbc.car.docs.definitions import CarDocs, Column, CommonFootnote, PartType
 from opendbc.car.car_helpers import interfaces, get_interface_attr
 from opendbc.car.values import PLATFORMS
 
+try:
+  from openpilot.common.basedir import BASEDIR
+except ModuleNotFoundError:
+  # TODO: make sure we run in a sensible way if not checked out as a submodule
+  BASEDIR = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../../"))
```
sshane, line 19: Make a basedir file for this

opendbc/car/docs/dump_car_docs.py
sshane, line 1: This is expected to not be needed once we clean up the bot, which also only lives in openpilot, do we need to move everything?

opendbc/car/docs/generate_docs.py
jyoung8607: Turns out we already have one!

opendbc/car/docs/dump_car_docs.py
jyoung8607: If they're going away soon, I'm down with just making import fixes, just didn't make sense to leave them in selfdrive/debug long-term.

jyoung8607:
> - [ ] comma.ai/vehicles impact understood
> - [ ] shop.comma.ai impact understood

@sshane At your convenience, can you help make sure this stuff follows? Per our discussion, the API output will be kept stable even with upcoming CARS.md 2.0, but consumers will need to change their import path.

opendbc/car/docs/generate_docs.py
```diff
@@ -1,15 +1,24 @@
...
+from opendbc.car.common.basedir import BASEDIR
 from opendbc.car.structs import CarParams
-from opendbc.car.docs_definitions import CarDocs, Column, CommonFootnote, PartType
+from opendbc.car.docs.definitions import CarDocs, Column, CommonFootnote, PartType
 from opendbc.car.car_helpers import interfaces, get_interface_attr
 from opendbc.car.values import PLATFORMS
 
 
+# FIXME: make sure we run in a sensible way if not checked out as a submodule, try-except for writing to CARS.md
+CARS_MD_OUT = os.path.join(BASEDIR, "../", "../", "../", "docs", "CARS.md")
```
sshane, line 18: The path to the out file should be in openpilot and passed in

jyoung8607: > The path to the out file should be in openpilot and passed in.

I can do that.

> Since we don't write any docs in opendbc, probably best to leave everything in openpilot for now

We will. The new ALL-CARS.md (or whatever name we bikeshed) will live in opendbc. It needs to run, generate, save, and test in opendbc without having openpilot available.

CARS.md only lives in openpilot because we don't want to change the link until we have a bigger step-change in how it looks and works.

> from opendbc.car.docs import generate_cars_md, get_all_car_docs

I'll see if I can juggle things to maintain that import location.

sshane: You can just make docs.py executable and write the opendbc docs by default, then openpilot's docs.py can stay

PR #1346, opened 2024-10-05
Add Clarity

vanillagorillaa:


jyoung8607:
Clarity is a great one to link as community supported, but much like openpilot, this file is autogenerated. Have a look at #1347, and generate with opendbc/car/docs.py.

PR #1347, opened 2024-10-05
docs: update Honda works in progress

jyoung8607:
Gather some Honda work-in-progress history so that we can start them moving forward again.


opendbc/car/extra_cars.py
```diff
@@ -28,6 +28,16 @@ def init_make(self, CP: structs.CarParams):
 class CAR(Platforms):
   config: ExtraPlatformConfig
 
+  HONDA_CANFD = ExtraPlatformConfig(
+    [
+      CommunityCarDocs("Acura Integra 2024", "All"),
```
vanillagorillaa, line 33: Integra isn't actually CANFD, just regular CAN

jyoung8607: I have some minor updates to add soon, I'll re-categorize Integra and add Clarity while doing that. Thanks!

PR #1349, opened 2024-10-05
Hyundai: Add FW version for Ioniq 5 2023

sunnyhaibin:
Add firmware versions for the Hyundai Ioniq 5 2023.

Dongle ID: `a075376978f75293`

Thanks to the community Hyundai Ioniq 5 2023 owner bazerk3182 (Discord).


royjr:
This is a UK 2023 Ioniq 5 Ultimate non-HDA2 that uses a Hyundai K harness, but "Non-US only" seems to be conflicting. https://github.com/commaai/opendbc/blob/89b0f059d77d189ec771ee79fb4aa2a8d7771632/opendbc/car/hyundai/values.py#L312-L313

PR #1350, opened 2024-10-05
GM: Add Snow/Ice Mode

garrettpall:
I was live streaming cabana trying to find the LKAS button and found this.


garrettpall:
I used turn signals to grab a route for proof, but it was merged while I was driving. 
Screenshot 2024-10-05 at 3 04 33 PM

PR #1354, opened 2024-10-06
Toyota: Fix RAV4 Prime mass and wheelbase

hdoublearp:
Weight and wheelbase specifications obtained from https://web.archive.org/web/20221005134205/https://www.toyota.com/rav4prime/2022/features/mpg_other_price/4544/4550. Steering ratio obtained from https://www.toyota.com/content/dam/toyota/brochures/pdf/2022/rav4_ebrochure.pdf.

### DIFF
--- a/opendbc/car/toyota/values.py
+++ b/opendbc/car/toyota/values.py
@@ -252,7 +252,7 @@ class CAR(Platforms):
     # TODO: Enable this docs entry when it can be suppressed from openpilot CARS.md
     # [ToyotaCarDocs("Toyota RAV4 Prime 2021-23", min_enable_speed=MIN_ACC_SPEED)],
     [],
-    CarSpecs(mass=3650. * CV.LB_TO_KG, wheelbase=2.65, steerRatio=16.88, tireStiffnessFactor=0.5533),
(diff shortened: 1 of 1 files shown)


jyoung8607:
You seem to be correct on the mass/wheelbase. Your link is broken, but I rechecked online sources, and yours are closer (4,372 lbs and 105.9 inches/2.68mm). I checked this on Edmunds before merging so I must have gotten the wrong RAV4 variant.

The steerRatio needs to stay where it's at. The manufacturer will tell you the gearing ratio inside the EPS, but openpilot needs to know about actual wheel/tire movement, which includes some other suspension effects. For this, we take the value from the online params learner once it looks mostly stable. If we leave the manufacturer value in there, the car will drive wrongly/differently for the first 20 or 30 minutes of driving until the online params learner can fix things.

New reference from here: https://www.motortrend.com/reviews/2021-toyota-rav4-prime-first-test-review/

If you'll slightly tweak the mass, and round the wheelbase to 2.68mm (a standard convention in this and other vehicle ports), and revert the steerRatio change, I'll go ahead and merge. Marking as Draft pending those changes.

hdoublearp:
This makes sense to me!

It's a shame Toyota didn't keep those deep links around for long.

Will implemented your suggestion and mark for review. Thanks!

PR #1355, opened 2024-10-07
Toyota: Add FW for 2023 RAV4 Prime

jyoung8607:
Add missing firmware for the 2023 RAV4 Prime, thanks to @dan3939!


jyoung8607:
This is a verified 2023 RAV4 Prime by VIN, but the firmware format does change enough that the fuzzy fingerprint code will need to be adapted. Holding in Draft for the moment.

PR #1358, opened 2024-10-07
Lexus ES TSS2: switch braking signal

sshane:
So far my theory is ACC_BRAKING describes the brake lights and activates for engine braking, and ICBACT is friction brakes actuation. How the car gets to more than 0.5 m/s^2 of deceleration with a NEUTRAL_FORCE that says it can only engine brake ~0.1 m/s^2 is a mystery though.

If this new signal is truly when the friction brakes are being actuated, then NEUTRAL_FORCE might not be correct, or depends on the gear possibly.

There is also a weird descrepency of aEgo undershooting the ACCEL_NET brake request from the PCM, that starts only when ICBACT goes high:

image

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -164,6 +164,9 @@ class CarController(CarControllerBase):
       self.pcm_accel_compensation = rate_limit(pcm_accel_compensation, self.pcm_accel_compensation, -0.01, 0.01)
       pcm_accel_cmd = actuators.accel - self.pcm_accel_compensation
 
+      # prevent request from deviating too far from current pcm accel net
+      pcm_accel_cmd = max(pcm_accel_cmd, CS.pcm_accel_net - 0.5)
+
       # Along with rate limiting positive jerk below, this greatly improves gas response time
       # Consider the net acceleration request that the PCM should be applying (pitch included)
       if net_acceleration_request < 0.1:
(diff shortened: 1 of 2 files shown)


jyoung8607:
Side note, I see that CI breaks because the message is missing for RAV4 Prime. For the moment, if you don't mind arranging your change [like this](https://github.com/commaai/opendbc/pull/1287/commits/28c2d61ee8766e6e5103a428fbc32f7cc3e8874f) then I will fix this properly with commaai/opendbc#1319.

sshane:
The friction brakes activate before `ICBACT` goes high, so my suspicion was wrong, `ACC_BRAKING` is accurate.

> Looking only at this PR, not being deeply familiar with Toyota or with your other work, perhaps ICBACT means internal combustion engine braking active.

Possibly, but it always activates later, after requesting a higher threshold of deceleration. Which seems counter-intuitive as engine braking should be first to activate, but still looking into this! In almost all cases, this signal describes when the braking acceleration request mismatches from aEgo.

jyoung8607:
> Possibly, but it always activates later, after requesting a higher threshold of deceleration. Which seems counter-intuitive as engine braking should be first to activate, but still looking into this! In almost all cases, this signal describes when the braking acceleration request mismatches from aEgo.

Are you testing on a hybrid? If so, I'd expect the powertrain to prefer regen over compression braking, and potentially apply both when moderate decel is called for. The transmission may also need to downshift to get into a regime where no-fuel/no-ignition engine RPM is a substantial drag after accounting for torque converter slip.

Still just thinking out loud here, but when studying what these signals do, maybe throw current trans gear (1 through 8) up on a graph and see if there are interesting time and value correlations.

PR #1359, opened 2024-10-08
VW: Add more CRC data identifiers

dalathegreat:
### What
This PR adds more VAG magic bytes to the 8H2F Autosar lookup table. 

### Why
To make it easier for people reverse engineering Volkswagen CAN

### How
During some VW MEB reverse engineering, I found these decoded on [this excellent repository](https://github.com/crasbe/VW-OnBoard-Charger/tree/main/Software/Reverse%20Engineering)


jyoung8607:
Last time this came up, we decided it was okay to add messages we don't actively use in openpilot, but only if we have [test coverage](https://github.com/commaai/opendbc/blob/c78d9bbc91601abc72992faebae6e18afefbfc40/opendbc/can/tests/test_checksums.py#L93-L556). That's going to require a DBC.

Do you have one you're working with? I have *a little* info, it seems very similar to MQBevo.

dalathegreat:
@jyoung8607 Unfortunately I have no DBC, I am going straight for an implementation to use MEB platform batteries in standalone mode. PR open [here on the Battery-Emulator](https://github.com/dalathegreat/Battery-Emulator/pull/405). Once we are completed, we can start owrking on a DBC for the MEB. So feel free to reject this PR (but having the info here will help others doing reverse engineering)

jyoung8607:
If you can supply the sixteen necessary sample messages (one for each counter value) I can help you construct a DBC with placeholder entries, even if we don't know the meaning of many/most bits.

dalathegreat:
I can get back to this once we are completed!

PR #1363, opened 2024-10-12
Ford: enable radar parsing

adeebshihadeh:


sshane:
Slower than plan, but getting way too close during offramp (tracks parsing not correct?): b9d015f8ff659e1e/00000424--38a4208a5d/6

PR #1366, opened 2024-10-14
Tesla: setSpeed to 0 if not active

lukasloetkolben:
Setting the speed to anything other than 0 when we are not engaged puts the Driver Interface (DI_cruiseState) in the UNAVAILABLE state. This stops us from engaging while the car is at a standstill.


lukasloetkolben:
Make sense, because vEgo is 0 in standstill.

PR #1374, opened 2024-10-15
Toyota: improve longitudinal control for TSS2 hybrid

sshane:
Will start with just Camry TSS2 Hybrid to ship this, then have users submit reports.


sshane:
Even when rate limiting the brake request, aEgo can still overshoot the target, and it's not always visible in ACCEL_NET request or the brake amount being applied. Still looking into how to counter this

PR #1379, opened 2024-10-16
Toyota: improve start from stop for TSS2 hybrid

sshane:
In real driving and the longitudinal maneuvers, you can see a positive overshoot of aEgo on occasion, and it's sometimes shown in the accel request by the PCM and braking force application signal (`BRAKE`) and sometimes not, so that is not counteracted yet in this PR.

This PR aims to ship `PERMIT_BRAKING` in the `ACC_CONTROL` message. The current problem is any offset in the PCM between the openpilot acceleration request and `ACCEL_NET` acceleration request by the PCM when braking at a stop is maintained/not unwound when using `PERMIT_BRAKING`=False, so if you try to request 0.0 m/s^2, the car will end up accelerating at ~1.0 m/s^2, dependant on the jerk you go from a braking to a coasting request. We will try to learn the offset before moving using a PID controller.


mkd808:
longitudinal_maneuvers report for TSS 2 Corolla Hybrid - 7c7d3d5e5ff31d51/00000017--84c3f8cedf

sshane:
Thanks! https://github.com/commaai/opendbc-data/commit/a441b3358bf20ae5ca63980087c172c258a76053

sshane:
- [x] Not near a stop, we now model the coasting acceleration accurately, so we don't over-brake or under-accelerate at high speeds

What's left:

- [ ] Coming to a stop, the PCM thinks our acceleration request approaching 0 is a good reason to release brakes pre-maturely
- [ ] Leaving a stop, we need to ensure we learn the PCM gas offset before taking off or we overshoot the request.

sshane:
PCM is too complex to model simply, and the brake request overshooting is still a separate issue.

PR #1382, opened 2024-10-16
CANParser: expose `MessageState` to Cython to eliminate data copying

deanlee:
This PR greatly simplifies the update logic, resulting in reduced code complexity and eliminating unnecessary data copying.

### Main Changes:
1. Removal of `SignalValue` Struct and `CANParser::query_latest` Function: These components, previously used for copying updated values to Cython, have been removed. This change streamlines the data flow between C++ and Cython.
2. Direct Exposure of `MessageState` to Cython: `MessageState` is now directly accessible from Cython, allowing values to be read straight from the C++ object. This eliminates the need for intermediate copying, optimizing performance and simplifying the overall code structure.

Additionally, this PR includes changes from #1381, where value dictionaries are now initialized directly from DBC messages, though it does not include the `test_parser_empty_list() `function.

Below are the performance benchmark results comparing this PR against the current master branch:

**Benchmark (PR):**

```
6000 CAN packets, 10 runs
272.64 mean ms, 278.98 max ms, 268.80 min ms, 2.98 std ms
0.0454 mean ms / CAN packet
```

**Benchmark (Master):**

```
6000 CAN packets, 10 runs
335.81 mean ms, 394.27 max ms, 312.58 min ms, 24.96 std ms
0.0560 mean ms / CAN packe
```

Most of the time is spent converting and copying messages from Python to C++. so while the current results don't show a massive improvement, after #1383 is merged, this PR could reduce the time per packet from **0.0318 ms** to under **0.02 ms**. I will run additional benchmarks to confirm once that change is integrated.

### DIFF
--- a/opendbc/can/common.h
+++ b/opendbc/can/common.h
@@ -81,7 +81,7 @@ public:
             const std::vector<std::pair<uint32_t, int>> &messages);
   CANParser(int abus, const std::string& dbc_name, bool ignore_checksum, bool ignore_counter);
   std::set<uint32_t> update(const std::vector<CanData> &can_data);
-  MessageState *messageState(uint32_t address) { return &message_states.at(address); }
+  MessageState *getMessageState(uint32_t address) { return &message_states.at(address); }
 
 protected:
   void UpdateCans(const CanData &can, std::set<uint32_t> &updated_addresses);
--- a/opendbc/can/common.h
+++ b/opendbc/can/common.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <map>
+#include <set>
 #include <string>
 #include <utility>
 #include <unordered_map>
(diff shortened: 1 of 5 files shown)


sshane:
Benchmarks on a 3X:

PR:

```
6000 CAN packets, 10 runs
437.80 mean ms, 440.19 max ms, 436.00 min ms, 1.38 std ms
0.0730 mean ms / CAN packet
```

master (after bus/addr filter PR):

```
6000 CAN packets, 10 runs
395.90 mean ms, 397.14 max ms, 394.55 min ms, 0.83 std ms
0.0660 mean ms / CAN packet
```

deanlee:
@sshane : FYI, this PR also resolves [issue #1066](https://github.com/commaai/opendbc/issues/1066), as it no longer relies on timestamps for updating the last values. Instead, it updates the updated_addresses set whenever a CAN signal is successfully parsed.

```
if (state_it->second.parse(can.nanos, frame.dat)) {
    updated_addresses.insert(state_it->first);
}

```

This approach ensures the updated values are tracked more reliably, independent of the timestamp.

I believe we can close issue #1066 once this is merged.

sshane:
Nice clean up!

sshane:
@deanlee do you mind making the same change with PlotJuggler? It used `SignalValue` which we removed

deanlee:
@sshane : https://github.com/commaai/PlotJuggler/pull/67 should work, However, since I'm still unable to build PlotJuggler, feel free to take over it if you'd like.

PR #1383, opened 2024-10-16
CANParser: filter CAN messages by bus and addresses

deanlee:
This PR adds filtering for CAN messages  based on the bus and addresses that the CANParser specifically cares about.
Instead of converting and copying all incoming messages from Python to C++ , we now only parse and copy those with relevant addresses to the c++ `CanFrame` vector. This reduces the overhead from unnecessary message processing, lowering CPU and memory usage while enhancing overall performance and efficiency.

Below are the performance benchmark results comparing this PR against the current master branch:

**Benchmark (PR):**

```
6000 CAN packets, 10 runs
190.79 mean ms, 219.88 max ms, 180.86 min ms, 10.96 std ms
0.0318 mean ms / CAN packet

```

**Benchmark (Master):**

```
6000 CAN packets, 10 runs
335.81 mean ms, 394.27 max ms, 312.58 min ms, 24.96 std ms
0.0560 mean ms / CAN packe
```

### DIFF
--- a/opendbc/can/parser_pyx.pyx
+++ b/opendbc/can/parser_pyx.pyx
@@ -8,7 +8,7 @@ from libcpp.vector cimport vector
 from libc.stdint cimport uint32_t
 
 from .common cimport CANParser as cpp_CANParser
-from .common cimport dbc_lookup, SignalValue, DBC, CanData, CanFrame
+from .common cimport dbc_lookup, SignalValue, DBC, CanData
 
 import numbers
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


sshane:
on a 3X:

PR:

```
6000 CAN packets, 10 runs
437.80 mean ms, 440.19 max ms, 436.00 min ms, 1.38 std ms
0.0730 mean ms / CAN packet
```

master:

```
6000 CAN packets, 10 runs
709.38 mean ms, 729.47 max ms, 705.66 min ms, 6.78 std ms
0.1182 mean ms / CAN packet
```

sshane:
Drops Bronco card CPU usage from ~30% to ~26% w/ radar!

PR #1392, opened 2024-10-21
Lexus: closed brake loop

sshane:


sshane:
Solved by learning the offset when the PCM accel net is not accurate: https://github.com/commaai/opendbc/pull/1461

PR #1405, opened 2024-10-23
Ford: add override signal

sshane:
May prevent brake wind up when overriding while it wants to brake. The stock system even starts the brake request from aEgo which further reduces the jerk:

image


sshane:
Has no effect

PR #1406, opened 2024-10-23
Set Ford CAN radar timestep

sshane:
For https://github.com/commaai/opendbc/pull/1363

Verified with this code in the `test_radar_interface` test_models function:

```
    if self.CP.flags & 1:  # CAN FD
      assert abs(msgs - 1200) < 50  # 20 Hz
    else:
      assert abs(msgs - 2000) < 50  # 33.3 Hz
```


sshane:
Actually doesn't do anything anymore, seeing about removing

PR #1408, opened 2024-10-23
Fixup ford radar

sshane:
- `MRR_Header_InformationDetections->CAN_NUMBER_OF_DET` exactly matches our parsed points using `CAN_DET_VALID_LEVEL_xx` when assuming the scan index signal is a sync signal
- Each track message's `CAN_SCAN_INDEX_2LSB_xx` signal matches the header's when triggered on the header message
- Every cycle we now receive exactly 65 track updates when triggered on the header message


opendbc/car/ford/radar_interface.py
```diff
@@ -68,6 +71,10 @@ def update(self, can_strings):
       self._update_delphi_esr()
     elif self.radar == RADAR.DELPHI_MRR:
       self._update_delphi_mrr()
+      print('pts', len(self.pts), self.rcp.vl['MRR_Header_InformationDetections']['CAN_NUMBER_OF_DET'])
```
sshane, line 74: `CAN_NUMBER_OF_DET` was always lower by 3-6 times from our number of points, now they match exactly!

It was originally thought that `CAN_SCAN_INDEX_2LSB_xx` was a sort of multiplexing signal, but it makes more sense that along with the header message `MRR_Header_InformationDetections` it's used to ensure that the latest updates are kept up to sync, the LSB always matches the counter in the header message when waiting on the header message now.

opendbc/car/ford/radar_interface.py
```diff
@@ -55,7 +56,9 @@ def update(self, can_strings):
     vls = self.rcp.update_strings(can_strings)
     self.updated_messages.update(vls)
 
-    if self.trigger_msg not in self.updated_messages:
+    # print('updated_messages:', self.updated_messages)
+    # if self.trigger_msg not in self.updated_messages:
+    if 368 not in self.updated_messages:
```
sshane, line 61: The LSB scan index signals always match the header counter's 2 LSB when waiting for it now, before it would usually have a few messages that mismatched

sshane:
Actually it looks like the original thinking was correct! I colored scan indexes of 0 and 2 as green, and 1 and 3 as purple. They're grouped into close range tracks and far away/less accurate tracks.


https://github.com/user-attachments/assets/25122dc3-5d57-4232-a1c3-cb743fa5812e

sshane:
I then filtered for only `CAN_DET_SUPER_RES_TARGET_xx`=1 tracks:

https://github.com/user-attachments/assets/2641c330-60ea-4e94-a9f2-0f0988488ad0

sshane:
I also just realized the groups are explainable with these two signals:

image

I had some false tracks on the road with the long-range detection mode, probably can filter them out with this:

image

opendbc/car/ford/radar_interface.py
```diff
@@ -106,13 +115,45 @@ def _update_delphi_esr(self):
...
     for ii in range(1, DELPHI_MRR_RADAR_MSG_COUNT + 1):
+      # if look_index != 0:
+      #   continue
       msg = self.rcp.vl[f"MRR_Detection_{ii:03d}"]
 
       # SCAN_INDEX rotates through 0..3 on each message
       # treat these as separate points
+      # Indexes 0 and 2 have a max range of ~40m, 1 and 3 are ~170m (MRR_Header_SensorCoverage->CAN_RANGE_COVERAGE)
+      # TODO: filter out close range index 1 and 3 points, contain false positives
+      # TODO: can we group into 2 groups?
```
sshane, line 149: Not really, the flickering re-appears

sshane:
mostly in

PR #1417, opened 2024-10-24
Ford: group similar coverage range measurements

sshane:
Increases rate of points from ~8Hz to ~16Hz, but decreases points by half

### DIFF
--- a/opendbc/car/car.capnp
+++ b/opendbc/car/car.capnp
@@ -324,6 +324,10 @@ struct RadarData @0x888ad6581cf0aacb {
 
     # some radars flag measurements VS estimates
     measured @6 :Bool;
+
+    flags @7 :Int32;  # radar specific flags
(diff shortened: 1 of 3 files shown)


sshane:
Combining the two measurements from each Doppler coverage setting doubles the update frequency of each radar point from ~8 Hz to ~16 Hz, which is seen clearly in the `liveTracks` messages. However, the resulting radard lead should be largely the same due to the clustering of the duplicate points:

image

sshane:
I noticed that Ford's radar parser often deletes and re-creates tracks due to some switching logic we have. radard doesn't handle this well, as aLeadK is calculated on a per-track basis, so it's nearly always zero, causing the planner to not react well to heavily braking leads.

This PR only exacerbates this problem, best to track each scan mode's points individually in case it helps reduce track switching.

PR #1420, opened 2024-10-26
Ford: cluster radar points

sshane:
Here is https://connect.comma.ai/b9d015f8ff659e1e/000004b6--2657291be4/440/465 clustered:

https://github.com/user-attachments/assets/af031651-76c2-4e75-95ff-541247887b24

We still need to somehow correlate which clusters are which between updates.


sshane:
5m DBSCAN clustering seems to be more stable:

https://github.com/user-attachments/assets/70aa0a31-74f4-47b5-8d58-21d944660603

- [ ] Need to cluster without dependencies now

PR #1421, opened 2024-10-26
Synchronize gas and accel after accel was tweaked

blue-genie:
Route: e36b272d5679115f/0000011d--ac087d2c7d/10:12

Every time when when slowing down to a stop and the speed was below 2-3mph, the ACC would inactivate itself.

Checking PJ, noticed that right before this happens, the gas would spike and become out of sync with accel. It seems that if gas is not inactive, it has to match accel.

Look after the green and red lines in the middle chart:
image
image

After the change:
Route: e36b272d5679115f/00000120--c793cf01c4/7:9

image


sshane:
openpilot is sending messages unlike stock openpilot, plus either your panda safety is modified, or you're triggering the safety as it's blocking messages and disengaging from that.

PR #1425, opened 2024-10-29
Ford: match radar points temporally

sshane:
So far python clustering approaches are extremely slow (https://github.com/commaai/opendbc/pull/1420), so try to dynamically assign track IDs based on previous measurements


sshane:
Something's broken here (correlation not working), so going to continue here which is the original clustering idea, but it combines the clustering and temporal correlation in one step to reduce CPU: https://github.com/commaai/opendbc/pull/1427

PR #1427, opened 2024-10-29
Ford: cluster radar points

sshane:
This aims to approximate the full clustering and correlation from https://github.com/commaai/opendbc/pull/1420 while being more efficient and simple. The results aren't as temporally consistent as that PR however, but I will evaluate if this is good enough.

---

Checklist:

- [x] Doesn't match the car on the right at the start of the segment as the lead: https://connect.comma.ai/b9d015f8ff659e1e/0000055b--dcf12a3696/117/180
    image
- [x] Doesn't match the cars on the left taking off from a stop: https://connect.comma.ai/b9d015f8ff659e1e/0000055b--dcf12a3696/206/222
    image
- [x] Consistent aLeadK measurement for entirety of lead on offramp: https://connect.comma.ai/b9d015f8ff659e1e/000004b6--2657291be4/440/457
    image


sshane:
The latest changes fixed the spikes every 12 frames (8Hz), now just need to verify the clustering behavior in some edge cases:

image

Issue #1435, opened 2024-11-02
Make CANParser.vl_all not use a defaultdict


sshane:
Seems dangerous if you make a typo for a signal name, and it silently always returns an empty list.

@deanlee interested?

PR #1437, opened 2024-11-02
Make CANParser.vl_all not use a defaultdict

deanlee:
resolve: https://github.com/commaai/opendbc/issues/1435


deanlee:
Closed. The issue can be resolved by the changes in the https://github.com/commaai/opendbc/pull/1439

PR #1439, opened 2024-11-03
CANParser: optimize data handling by eliminating data copy after Update()

deanlee:
This pull request removes unnecessary data copying from MessageState to dictionaries following update() calls. By encapsulating MessageState within a Cython object and implementing a read-only dictionary for direct access to the signal data, we eliminate the need to clear `vl_all` and copy updated data to `vl`, `vl_all` and `ts_nanos`. This optimization yields approximately a 1x performance improvement on the 3X device.

With this refactor, the impact on dictionary read performance is negligible, as the read-only dictionary is a minimal wrapper around the C++ MessageState, allowing direct access to data from the C++ `unordered_map` without additional overhead.

Passes tests on the device and OpenPilot branch [#33926](https://github.com/commaai/openpilot/pull/33926).

**Key Changes:**
1. **Encapsulated MessageState in a Cython Class:** Enables direct access to signal data from the C++ MessageState, eliminating the need for data copying after `update()`.
2. **Implemented a Custom ReadOnlyDict:** Overrides `__getitem__` to retrieve data from C++ MessageState, ensuring read-only access.
3. **Adjusted `test_checksum.py`:** Removed deep copy requirements as the new implementation is read-only.

Below is the benchmark comparison on the 3X device:

**Benchmark (PR):**
```

6000 CAN packets, 10 runs
265.55 mean ms, 267.66 max ms, 264.63 min ms, 0.86 std ms
0.0443 mean ms / CAN packet
```

**Benchmark (Master):**

```
6000 CAN packets, 10 runs
491.51 mean ms, 492.63 max ms, 490.43 min ms, 0.71 std ms
0.0819 mean ms / CAN packet
```
-------------------
This PR is an improved version of https://github.com/commaai/opendbc/pull/1046

### DIFF
--- a/opendbc/can/parser_pyx.pyx
+++ b/opendbc/can/parser_pyx.pyx
@@ -39,30 +39,19 @@ cdef class MessageState:
 
 
 class ReadonlyDict(Mapping):
-  def __init__(self, state):
-    self.state = state
-    self.keys = self.state.signal_names
+  def __init__(self, state, value_accessor):
+    self._state = state
+    self._keys = state.signal_names
+    self._get_value = value_accessor
+
+  def __getitem__(self, key):
+    return self._get_value(self._state, key)
 
   def __iter__(self):
-    return iter(self.keys)
+    return iter(self._keys)
 
   def __len__(self):
-    return len(self.keys)
-
-
-class ValueDict(ReadonlyDict):
-  def __getitem__(self, key):
-    return self.state.value(key)
-
-
-class NanosDict(ReadonlyDict):
-  def __getitem__(self, key):
-    return self.state.ts_nanos(key)
-
-
-class AllValueDict(ReadonlyDict):
-  def __getitem__(self, key):
-    return self.state.all_values(key)
+    return len(self._keys)
 
 
 cdef class CANParser:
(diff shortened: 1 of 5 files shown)


deanlee:
After this, only one bottleneck remains: the conversion related to can_capnp_to_can_list_cpp. By using a NumPy array, the existing multi-step conversion:

> capnp string -> vector<CanData> -> Python list -> vector<CanData>

can be simplified to a single conversion: 

> capnp string -> NumPy array. 

This would allow the `CanParser` updates to be executed almost entirely in C++ code. Cython would serve only as a simple wrapper for calling the function. In my local draft. this optimization can further double the performance, reducing the processing time from **0.0443** ms per CAN packet to **0.0182** ms per CAN packet.

opendbc/can/parser_pyx.pyx
```diff
@@ -6,12 +6,64 @@ from libcpp.string cimport string
...
+
+
+class NanosDict(ReadonlyDict):
+  def __getitem__(self, key):
+    return self.state.ts_nanos(key)
+
+
+class AllValueDict(ReadonlyDict):
+  def __getitem__(self, key):
+    return self.state.all_values(key)
```
sshane, line 65: Anyway we can get by with `ReadonlyDict`?

deanlee: how about [ebf4696](https://github.com/commaai/opendbc/pull/1439/commits/ebf4696481c6aef07319011589c8715012c825cd)

sshane:
This brings down Bronco card CPU usage from 28% to 20%!

deanlee:
> This brings down Bronco card CPU usage from 28% to 21%!

by the way, I've also eliminate the multiple conversions from capnp can data to vector[CanFrame] while call `CanParser.update.` I'll test and submit soon. With this PR and the conversion simplification, `CanParser` and `can_capnp_to_list` will run entirely in C++ at full speed—extremely fast!

PR #1446, opened 2024-11-06
Ford: enable long by default

sshane:


opendbc/car/ford/interface.py
```diff
@@ -42,10 +42,9 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experime
       cfgs.insert(0, get_safety_config(structs.CarParams.SafetyModel.noOutput))
     ret.safetyConfigs = cfgs
 
-    ret.experimentalLongitudinalAvailable = True
-    if experimental_long:
-      ret.safetyConfigs[-1].safetyParam |= Panda.FLAG_FORD_LONG_CONTROL
-      ret.openpilotLongitudinalControl = True
+    # TODO: verify stock AEB compatibility and longitudinal limit safety before shipping to release
+    ret.safetyConfigs[-1].safetyParam |= Panda.FLAG_FORD_LONG_CONTROL
+    ret.openpilotLongitudinalControl = True
```
adeebshihadeh, line 47: only for Q3?

PR #1448, opened 2024-11-06
Toyota: set pcm_accel_net when possible

sshane:
Useful when working on porting this to other cars and having a list of cars that don't have this message


sshane:
@deanlee does your new CANParser refactor allow checking `in`? I can make this a constant list if not

deanlee:
@sshane : it support, ReadonlyDict inherits from Mapping, collections.abc.Mapping base class already provides the default implementations for many dictionary-like methods, including in, values(), items(), and get(). Therefore, we only need to implement __getitem__, __iter__, and __len__.

PR #1453, opened 2024-11-07
Toyota: learn PCM accel request

sshane:
The internal PCM acceleration request that tells us what it's going to apply is mostly correct, but has some persistent offsets we don't fully account for (due to engine creep force, road pitch, friction, general inaccuracies) and causes acceleration over/undershoot which can be really bad since we're currently not error correcting on aEgo.

This PR grounds the internal PCM acceleration request using aEgo. It's sort of like using the shape of the internal value and the magnitude of aEgo, so that it matches what we're achieving, but shows us the future state of the vehicle as well to error correct in time!

With this, we can use a PI controller for the PCM compensation without any worry of under/overshoot error as well as remove the conventional feedback control.

Corolla (a2bddce0b6747e10/0000033f--bc20a12d8b):

image

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -57,7 +57,7 @@ class CarController(CarControllerBase):
     self.pcm_accel_nets_new = []
     self.offsets = []
 
-    self.filter = FirstOrderFilter(0, 1., DT_CTRL)  # 1.5 might be okay
+    self.filter = FirstOrderFilter(0, 1.0, DT_CTRL)  # 1.5 might be okay
     self.pid = PIDController(0.5, 0.5)
 
     self.pcm_accel_compensation = 0.0
@@ -65,6 +65,7 @@ class CarController(CarControllerBase):
 
     self.packer = CANPacker(dbc_name)
     self.accel = 0
+    self.accel_prev = 0
 
     self.secoc_lka_message_counter = 0
     self.secoc_lta_message_counter = 0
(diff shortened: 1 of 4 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -163,6 +164,9 @@ def update(self, CC, CS, now_nanos):
     # *** gas and brake ***
     # For cars where we allow a higher max acceleration of 2.0 m/s^2, compensate for PCM request overshoot and imprecise braking
     if self.CP.flags & ToyotaFlags.RAISED_ACCEL_LIMIT and CC.longActive and not CS.out.cruiseState.standstill:
+      actuators_accel = min(actuators.accel, self.accel_prev + ACCEL_WINDUP_LIMIT / 6 / 3)
+      self.accel_prev = actuators_accel
```
sshane, line 168: When taking off from a stop, we now rate limit the request before everything (and slow it down) so that the PCM compensation can recognize when the internal PCM request is rising faster than us, and then allows us to compensate before we even start moving

opendbc/car/toyota/carcontroller.py
```diff
@@ -201,19 +205,23 @@ def update(self, CC, CS, now_nanos):
...
       # prevent compensation windup
-      pcm_accel_compensation = clip(pcm_accel_compensation, actuators.accel - self.params.ACCEL_MAX,
-                                    actuators.accel - self.params.ACCEL_MIN)
+      pcm_accel_compensation = clip(pcm_accel_compensation, actuators_accel - self.params.ACCEL_MAX,
+                                    actuators_accel - self.params.ACCEL_MIN)
 
-      self.pcm_accel_compensation = rate_limit(pcm_accel_compensation, self.pcm_accel_compensation, -0.01, 0.01)
-      pcm_accel_cmd = actuators.accel - self.pcm_accel_compensation
+      # TODO: we need a better way to allow the compensation to prevent the request from rising other than insane limits
+      up_limit = 0.1 if CS.out.vEgo < 5 else 0.01
```
sshane, line 222: We need to come up with a better way other than allowing insane limits to allow the request to stay low while we're ramping up to full acceleration. I think a low pass filter w/ no rate limits might do well to filter out the high frequency oscillations we see that this is really here to solve.

sshane: This is why I needed to write this for testing, the rate limit is too low to prevent the rise of the acceleration request when the PCM compensation sees it overshooting:

image

Here is it with a PI controller:

...

sshane:
This segment is an example of why we need this, even for the Lexus. It corrects for the open loop PCM accel compensation being sometimes wrong, especially if the pitch estimate is wrong:

https://connect.comma.ai/57048cfce01d9625/00000323--2b5c73370b/5710/5741

image

sshane:
done in https://github.com/commaai/opendbc/pull/1494

PR #1455, opened 2024-11-07
Raven PARTY

robbederks:
New Model S Raven port based from the PARTY bus instead of the more classic CHASSIS bus


sshane:
Do we plan on merging this soon?

opendbc/car/tesla/values.py
```diff
@@ -47,32 +47,65 @@ class CAR(Platforms):
...
   TESLA_MODEL_Y = TeslaPlatformConfig(
     [
       TeslaCarDocsHW3("Tesla Model Y (with HW3) 2020-23"),
       TeslaCarDocsHW4("Tesla Model Y (with HW4) 2024"),
      ],
     CarSpecs(mass=2072., wheelbase=2.890, steerRatio=12.0),
+    {Bus.party: 'tesla_model3_party'},
+  )
+  TESLA_MODEL_S_RAVEN = PlatformConfig(
+    [CarDocs("Tesla Model S Raven", "All")],
```
sshane, line 61: What year is this for?

robbederks: iirc it's 2019-2021

adeebshihadeh:
This still the port you're driving on?

robbederks:
Roughly the same yes, but lukas merged it more with the M3 port. Feel free to close

adeebshihadeh:
Yeah let’s revisit when we have a pp!

PR #1460, opened 2024-11-08
Toyota: learn PCM accel request v2

sshane:
Hopefully a shippable version of https://github.com/commaai/opendbc/pull/1453

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -181,6 +181,7 @@ class CarController(CarControllerBase):
         # find closest value to aEgo in pcm_accel_net_deque
         arr = np.array(self.pcm_accel_net_deque) - accel_due_to_pitch
         idx = (np.abs(arr - CS.out.aEgo)).argmin()
+        # TODO: move into 33hz update loop!
         self.cur_idx = clip(idx, self.cur_idx - 1/3, self.cur_idx + 1/3)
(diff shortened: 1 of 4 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -154,10 +167,36 @@ def update(self, CC, CS, now_nanos):
...
+        # self.pid.reset()
+        new_pcm_accel_net = CS.pcm_accel_net
+      else:
+        # TODO: decide a good value for past pcm accel net
+        # find closest value to aEgo in pcm_accel_net_deque
+        arr = np.array(self.pcm_accel_net_deque) - accel_due_to_pitch
+        idx = (np.abs(arr - CS.out.aEgo)).argmin()
+        self.cur_idx = clip(idx, self.cur_idx - 1/3, self.cur_idx + 1/3)
+        self.debug3 = round(float(self.cur_idx))
+        print(self.cur_idx)
```
sshane, line 186: We need this dynamic indexing to solve innacuracies under high jerk. Here is assuming a 0.4s offset:

image

Here is the dynamic indexing:

image

sshane: And here is both gas and brake w/ dynamic indexing (from maneuver report):

image

Basically, a hard-coded offset that worked for one case/car didn't work for another

PR #1463, opened 2024-11-08
Revert "Toyota: learn offset for PCM acceleration request"

sshane:
Reverts commaai/opendbc#1461

Need to get the aEgo delay thing figured out for high jerk scenarios, but on the right track. Reverting until it works in all cases


gregjhogan:
High jerk may also cause the pitch of the device changing independently from the pitch of the road due to suspension movement, which I found was significant and annoyingly difficult to account for when trying to use pitch to calculate desired engine torque output for the RAM.

sshane:
Done in https://github.com/commaai/opendbc/pull/1494

Issue #1462, opened 2024-11-08
Lexus ES: overshoots start from stop maneuvers


sshane:
I cannot reproduce the nice starts shown in the latest Lexus ES report I gathered before merging https://github.com/commaai/opendbc/pull/1303:

https://commaai.github.io/opendbc-data/longitudinal_reports/LEXUS_ES_TSS2_57048cfce01d9625_00000208--c387e67187.html

What the route in the report is doing is setting the permit braking bit when the slow rate limited accel request > 0, but on master it sets permit braking immediately once the maneuver starts, causing a bunch of overshoot (similar to our Corolla! are they more similar than I had thought?).

Revert back to this behavior, or https://github.com/commaai/opendbc/pull/1453 should fix this.

sshane:
I see what might have happened. I must have gotten the maneuver report while on the first commit of https://github.com/commaai/opendbc/pull/1301 which sets permit braking based on the rate limited acceleration, but then realized it should be based on pitch and switched it before merging. This is fundamentally different than what I tested though, as that accel request signal isn't rate limited.

See the difference:

![image](https://github.com/user-attachments/assets/51382367-a79b-48a3-91d8-28a420236411)

PR #1465, opened 2024-11-09
Lexus ES TSS2: fix acceleration overshoot from a stop

sshane:
See https://github.com/commaai/opendbc/issues/1462


sshane:
Doing the rate limit before the PCM compensation also allows us to potentially correct for overshoot before it happens.

PR #1475, opened 2024-11-13
Toyota: add an accel winddown rate limit

sshane:
Matched Corolla's ACCEL_NET winddown limit:

image

Also might be crazy, but I observed the Corolla not overshooting in normal sng with an acceleration request that heavily oscillated within 0.1 m/s^2 or so, does that somehow prevent internal PCM wind up seen here? Try it!


sshane:
The rate limit on the brakes didn't prevent windup, but they should prevent the PCM compensation from winding up during the step response.

PR #1477, opened 2024-11-13
Toyota: use a filter for PCM compensation

sshane:
We mainly want to reduce the high frequency oscillations and feedback from the PCM signals, not restrict the rate. This should do that. Split from https://github.com/commaai/opendbc/pull/1475


sshane:
Just the filter change, looks pretty good!

57048cfce01d9625/0000021f--9761f9d007/2

image

Issue #1296, opened 2024-08-26
Volkswagen B harness


jyoung8607:
**New Harness Checklist**

- [x] which car is this for? Volkswagen in general newer than MY2022, including MQBevo (but not MEB)
- [ ] schematic - Push to neo repo once confirmed working
- [ ] beta sent to tester
- [ ] in shop
- [ ] Added to openpilot docs

**Additional Info**

This harness is intended to work for newer MQB with the 2Q0 camera, and also MQBevo with the 5WA camera. The wiring between the two types isn't the same, but for our purposes, they are compatible enough to combine.

The male connector housing is `8W0 972 112 A` ([example source](https://www.ecstuning.com/b-genuine-volkswagen-audi-parts/housing/8w0972112a/)). It's pricey, but there also seems to be a Taobao seller.

The female connector housing has not been found on the market. @vanillagorillaa went so far as to purchase and disassemble the camera, and the PCB mount housing had no identifying markings. @lukasloetkolben has a 3D printable model.

While comma production is pending, a limited number of prototype kits are available [directly from Lukas](https://www.ebay.com/itm/395616600662).

**Harness Wire Mappings**

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/jyoun/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/jyoun/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

Terminal | MQB 2Q0 (New, Tan) | MQBevo (Tan) | Harness Mapping
-- | -- | -- | --
1 | Extended CAN High | ADAS CAN High | CAN 0/2 High
2 | Extended CAN Low | ADAS CAN Low | CAN 0/2 Low
3 | Fusion CAN High |   | CAN 1 High
4 | Fusion CAN Low |   | CAN 1 Low
5 | Camera heater supply | Camera heater supply | Passthrough
6 | Camera heater ground | Camera heater ground | Passthrough
7 | Ground | Ground | Ground
8 | Switched 12V | Switched 12V | Switched 12V
9 |   |   |  
[...]

PR #1487, opened 2024-11-14
Remove VW footnote from website

sshane:
@jyoung8607 does VW_MQB_A0 make sense to also hide or no?


sshane:
We can also remove since we don't sell camera harness at this time, correct?

jyoung8607:
CarFootnote `VW_EXP_LONG` is technically correct but admittedly confusing, and has outlived most (not all) of its use.

CarFootnote `VW_MQB_A0` can go away when you ship #1296.

PR #1491, opened 2024-11-15
Toyota: improve new longitudinal tune

sshane:


sshane:
~During high jerk events, aEgo will often lag behind the ACCEL_NET request as the physical brake lag isn't captured in that signal, so we learn an incorrect offset. We will mitigate this by adjusting the learning rate when jerk of ACCEL_NET is high~

nvm, hybrids respond so quickly that they overshoot the brake request right as it goes negative, and we want to learn the overshoot there with that high jerk. Instead, we'll low pass filter the PCM ACCEL_NET and then learn the offset on that, which should reduce overshoot and incorrect offset learning on both hybrid (fast brakes) and ICE (slow brakes)

debug is pcm accel net, debug2 is filtered at 0.25s:

Camry Hybrid:

image

Lexus ES ICE:

You can see here how we were requesting more deceleration because we thought it was undershooting, even though what it was applying was physically rate limited and it eventually overshooted

image

sshane:
Looking at 4 ICE and 4 hybrids maneuver reports from https://commaai.github.io/opendbc-data, it's clear that there's roughly a 2x increase in down jerk, so we might want a variable time constant for the ACCEL_NET filter to provide the best results on both platform variants. Up jerk looks mostly the same

-1 m/s^2 step response means (ICE):
- TOYOTA_COROLLA_TSS2: 1.12s
- LEXUS_ES_TSS2: 1.73s
- LEXUS_ES_TSS2: 1.62s
- LEXUS_NX_TSS2: 0.71s

-1 m/s^2 step response means (hybrid):
- TOYOTA_COROLLA_TSS2: 0.57s
- TOYOTA_CAMRY_TSS2: 0.5s
- TOYOTA_HIGHLANDER_TSS2: 0.55s
- TOYOTA_PRIUS_TSS2: 0.57s

sshane:
PCM comp was removed with addition of https://github.com/commaai/opendbc/pull/1559

PR #1494, opened 2024-11-16
Toyota: adaptive ACCEL_NET for new long tune

sshane:
Split the learning from https://github.com/commaai/opendbc/pull/1491, wanted to keep it as history and that PR tests several changes I can ship separately.

This includes driving improvements in terms of following the acceleration request, as well as a straight up bug fix: when the PCM accel compensation was wrong (from bad pitch for example) and you were either at max brake or max gas, there was no way of correcting for it, now we ground it on aEgo with a low pass filter.

### DIFF
--- a/opendbc/car/__init__.py
+++ b/opendbc/car/__init__.py
@@ -2,7 +2,7 @@
 import logging
 from collections import namedtuple
 from dataclasses import dataclass, field
-from enum import IntFlag, ReprEnum, EnumType
+from enum import IntFlag, ReprEnum, StrEnum, EnumType, auto
 from dataclasses import replace
 
 from panda import uds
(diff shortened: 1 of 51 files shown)


sshane:
Maneuver report generated on this PR from route: 57048cfce01d9625/00000339--2de8bb149d

Compared against: https://commaai.github.io/opendbc-data/longitudinal_reports/LEXUS_ES_TSS2_57048cfce01d9625_00000316--44084725e9.html

Summary
---

The 20 mph brake maneuvers seem to overshoot more compared to the previous report, however I only have a report from before pitch compensation was added. Pitch in brake maneuvers causes more overshoot, so this isn't new with this PR.

No longer overshoots as badly and for as long for start from stop:

image

No longer overshoots accel and decel on the creep maneuver:

image

-1 /s^2 maneuver seems about the same:

image

-4 m/s^2 maneuver seems to overshoot slightly more:

image

+1 m/s^2 overshoots less:

image

+4 m/s^2 seems to be about the same.

PR #1495, opened 2024-11-16
Toyota: PERMIT_BRAKING shouldn't use rate limited decel

sshane:
fixes some overshoot due to the [down rate limit](https://github.com/commaai/opendbc/pull/1481) we merged recently

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -203,9 +203,10 @@ class CarController(CarControllerBase):
 
         # Along with rate limiting positive jerk above, this greatly improves gas response time
         # Consider the net acceleration request that the PCM should be applying (pitch included)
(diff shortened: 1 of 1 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -203,9 +203,10 @@ def update(self, CC, CS, now_nanos):
 
         # Along with rate limiting positive jerk above, this greatly improves gas response time
         # Consider the net acceleration request that the PCM should be applying (pitch included)
-        if net_acceleration_request < 0.1 or stopping or not CC.longActive:
+        net_acceleration_request_min = min(actuators.accel + accel_due_to_pitch, net_acceleration_request)
```
sshane, line 206: we can just use `actuators.accel + accel_due_to_pitch` once we are ready to use it for better start from stops

PR #1498, opened 2024-11-16
Toyota: improve start response for new tune

sshane:
Split from https://github.com/commaai/opendbc/pull/1491

Needs to be enabled on all cars first

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -19,8 +19,7 @@ VisualAlert = structs.CarControl.HUDControl.VisualAlert
 
 # The up limit allows the brakes/gas to unwind quickly leaving a stop,
 # the down limit roughly matches the rate of ACCEL_NET, reducing PCM compensation windup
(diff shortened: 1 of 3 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -227,10 +226,10 @@ def update(self, CC, CS, now_nanos):
 
         # Along with rate limiting positive jerk above, this greatly improves gas response time
         # Consider the net acceleration request that the PCM should be applying (pitch included)
-        net_acceleration_request_min = min(actuators.accel + accel_due_to_pitch, net_acceleration_request)
-        if net_acceleration_request_min < 0.1 or stopping or not CC.longActive:
+        net_acceleration_request_raw = actuators.accel + accel_due_to_pitch
+        if net_acceleration_request_raw < 0.1 or stopping or not CC.longActive:
```
sshane, line 230: Switches from setting PERMIT_BRAKING to 0 on the delayed, rate limited accel to the accel from openpilot which goes >0 much quicker

PR #1499, opened 2024-11-16
Toyota: enable new tune for all TSS2 which support it

sshane:
- [ ] hybrid has unexplainable PCM ACCEL_NET overshoot still, is there some physical effect we need to model as well?
  - [ ] corolla overshoot looks very similar to hybrid?!
- [x] corolla lets off brakes when coming to a final stop sometimes - https://github.com/commaai/opendbc/pull/1508
- [ ] Lexus once set PERMIT_BRAKING down early due to lack of neutral accel consideration: https://connect.comma.ai/57048cfce01d9625/00000331--8b51b57abb/558/642


sshane:
PCM comp was removed with addition of https://github.com/commaai/opendbc/pull/1559

devimauz:
hybrid has unexplainable PCM ACCEL_NET overshoot still, is there some physical effect we need to model as well?
 corolla overshoot looks very similar to hybrid?!

is this solved as well?

sshane:
It should be mostly, yes!

PR #1503, opened 2024-11-17
GM: SDGM CarSpecs to GMCarSpecs

garrettpall:


adeebshihadeh:
Anything we can do to make this bug harder to write in the future?

garrettpall:
I think my PR was open since before that change and nothing broke using it. Since all are updated now it shouldn't happen when someone copy/pastes.

PR #1508, opened 2024-11-18
Toyota: don't release brakes coming to a stop w/ new tune

sshane:
here's a come to stop scenario where Corolla released brakes too much due to bad creep accel logic:

green: previous actuators output
orange: new actuators output

image

PR #1509, opened 2024-11-19
Toyota: more stopping fixes for new long tune

sshane:
Potentially fixes https://connect.comma.ai/57048cfce01d9625/00000331--8b51b57abb/558/642 where we set permit braking to 0 despite still needing to brake to come to a stop. This is because we were on a 3 degree incline which caused us to go over the threshold, without considering we need to brake extra for the creep force

Builds upon and includes changes from https://github.com/commaai/opendbc/pull/1508 for testing

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -80,6 +80,7 @@ class CarController(CarControllerBase):
     lat_active = CC.latActive and abs(CS.out.steeringTorque) < MAX_USER_TORQUE
 
     if len(CC.orientationNED) == 3:
+      # self.pitch.x = CC.orientationNED[1]
       self.pitch.update(CC.orientationNED[1])
 
     # *** control msgs ***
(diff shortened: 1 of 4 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -233,6 +241,12 @@ def update(self, CC, CS, now_nanos):
         # Along with rate limiting positive jerk above, this greatly improves gas response time
         # Consider the net acceleration request that the PCM should be applying (pitch included)
         net_acceleration_request_min = min(actuators.accel + accel_due_to_pitch, net_acceleration_request)
+
+        # Prevents releasing brakes while stopping by considering engine creep force
+        # TODO: should this be if CS.neutral_accel > 0 instead?
+        if CS.acc_braking:
+          net_acceleration_request_min -= CS.neutral_accel
```
sshane, line 248: can use actuatorsOutput, but we want to eventually not include the rate limit for better start from stop perf so that's a non starter

sshane: this also regresses start from stop performance, fix that

PR #1512, opened 2024-11-20
Toyota: add 2024 EU RAV4 Hybrid fingerprint

peanut-hu:
With the latest nightly build and my firmware versions added to fingerprints, the car is not "unrecognized" anymore, but immediately starts to drop PCS/LTA errors when OpenPilot tries to use the sytem :(
Fingerprinting alone does not seem to help working with this EU XA50 RAV4 Hybrid model.


sshane:
What's your dongle id?

peanut-hu:
> What's your dongle id?

918ceea575036e0f

ErichMoraga:
Noting today's conversation in Discord...
https://discord.com/channels/469524606043160576/524327905937850394/1309200410568949861

sshane:
Can you put your device on WiFi to upload logs?

peanut-hu:
> openpilot is reporting the PRE_COLLISION message's checksum is invalid. Can you put your device on WiFi to upload logs?

Of course! Just tell me step-by-step how to do it - or point me to the right part of wiki, and i will follow.

I checked the Toyota harness but it seems to be the correct one as referenced here: https://github.com/commaai/neo/blob/master/car_harness/v3/Toyota_A_Harness.pdf
Also the CANBUS error happens with and without the OBD cable plugged in, so that is ruled out also.

PR #1515, opened 2024-11-21
Toyota: prevent overshoot w/ new tune

sshane:
Compensates for overshoot during high jerk maneuvers better, since there will always be some delay to the PCM ACCEL_NET.

To be clear, this has little effect on the Lexus ES since its jerk limit is so low, it doesn't overshoot that much, but it does fix other cars like the Corolla with 2x higher jerk and most likely the Camry Hybrid with 4x higher jerk.

Split from https://github.com/commaai/opendbc/pull/1513


sshane:
Has little effect on Lexus due to low jerk limits, but does fix quite a bit of overshoot on Corolla. Lexus ES report: 57048cfce01d9625/00000360--192e9b2a0a

PR #1516, opened 2024-11-21
CI: fix macOS test running

LouisVelez:
Addresses the issues found in [#1493](https://github.com/commaai/opendbc/pull/1493) (Mac OS not running)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,19 +22,15 @@ jobs:
         #run: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
     steps:
     - uses: actions/checkout@v4
-    - name: setup python 3.12
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.12
-    - name: Install coreutils on macOS
+    - name: nproc alias for macOS
       if: matrix.os == 'macos-latest'
       run: |
-        brew update
-        brew install coreutils
+        # Create alias for nproc using sysctl
+        echo 'alias nproc="sysctl -n hw.ncpu"' >> $GITHUB_ENV
     - name: install uv and dependencies
       run: |
         curl -LsSf https://astral.sh/uv/install.sh | sh
-        uv venv
+        uv venv --python 3.12
         # TODO: remove setuptools once https://github.com/cython/cython/issues/5751 is fixed
         uv pip install -e .[testing,docs] setuptools
     - run: uv run scons -j$(nproc)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,18 +22,13 @@ jobs:
         #run: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
     steps:
     - uses: actions/checkout@v4
-    - name: nproc alias for macOS
-      if: matrix.os == 'macos-latest'
-      run: |
-        # Create alias for nproc using sysctl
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -60,8 +61,9 @@ envCython["CCFLAGS"] += ["-Wno-#warnings", "-Wno-shadow", "-Wno-deprecated-decla
 envCython["CCFLAGS"].remove("-Werror")
 
 python_libs = []
-if arch == "Darwin":
-  envCython["LINKFLAGS"] = ["-bundle", "-undefined", "dynamic_lookup"]
+# macOS (both Intel and Apple Silicon)
+if sys.platform == "darwin":
```
adeebshihadeh, line 65: is `arch` not equivalent?

.github/workflows/tests.yml
```diff
@@ -22,7 +22,16 @@ jobs:
...
-    - name: setup python
+    - name: setup python 3.12
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.12
+    - name: Install coreutils on macOS
+      if: matrix.os == 'macos-latest'
+      run: |
+        brew update
+        brew install coreutils
```
adeebshihadeh, line 33: why do we need coreutils?

.github/workflows/tests.yml
```diff
@@ -22,7 +22,16 @@ jobs:
...
     - uses: actions/checkout@v4
-    - name: setup python
+    - name: setup python 3.12
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.12
+    - name: Install coreutils on macOS
+      if: matrix.os == 'macos-latest'
+      run: |
+        brew update
```
adeebshihadeh, line 32: do we need to update? seems like it'll just waste time

.github/workflows/tests.yml
```diff
@@ -22,7 +22,16 @@ jobs:
         #run: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]
     steps:
     - uses: actions/checkout@v4
-    - name: setup python
+    - name: setup python 3.12
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.12
```
adeebshihadeh, line 28: uv is supposed to handle this. is there a problem with the uv python?

.github/workflows/tests.yml
LouisVelez: There's no nproc by default on Mac so coreutils adds it but now I'm thinking that I could replace the brew code with  something like: 
`echo 'alias nproc="sysctl -n hw.ncpu"' >> $GITHUB_ENV`.  Will try it out later today.

SConstruct
LouisVelez: On Intel-based systems, it returns x86_64, while on Apple Silicone systems, it returns arm64. sys.platform returns darwin. envCython is the same for both, so I figured this was a nice quick way to do both without complicating logic.

.github/workflows/tests.yml
LouisVelez: fixed

.github/workflows/tests.yml
LouisVelez: I need to investigate this one a bit more. Github doesn't like the alias, and I'm a bit rusty on the Actions CI. 

worse case I add this back and take out the `brew update` I must note that if brew finds itself out of date it will auto-run `brew update`.

adeebshihadeh: oh it's not worth getting brew involved just for nproc. let's just get rid of nproc

adeebshihadeh:
Thanks for the PR!

pbassut:
@LouisVelez thanks for this!

PR #1521, opened 2024-11-23
ci: don't timeout if the runner is slow to pick up the job

maxime-desroches:


.github/workflows/tests.yml
```diff
@@ -51,3 +56,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-pre-commit-
     - run: uv run pre-commit run --all
+    - uses: ./.github/workflows/timeout
```
adeebshihadeh, line 59: other actions have something like a "Post checkout" that runs at the end. can we do that here?

maxime-desroches: Only possible with JS actions. Created a really simple one here: https://github.com/commaai/timeout
We now only have to do this once:
```
- uses: commaai/timeout@v1
```
or to specify a timeout other than 60 seconds:
```
- uses: commaai/timeout@v1
  with:
    timeout: 20
```

PR #1522, opened 2024-11-23
Toyota: simple long tuning improvement

sshane:
My goal is to ship a measurably improved long experience without regressing anything. Performance doesn't need to be perfect, just noticeably improved. This is becoming harder with the PCM acceleration request compensation stuff which is getting pretty complicated with all the filters and controllers.

This is an attempt to match the behavior of the new tune without relying on any not-fully-understood CAN signals.

Here's an example of the Corolla not braking enough on a down hill due to bad pitch, bad ACCEL_NET signal, and slow aEgo compensation: https://connect.comma.ai/a2bddce0b6747e10/00000503--6377595dba/154/209


sshane:
Calculating jEgo and extrapolating into the future by the long actuator delay actually matches what the new tune does, without needing to compensate for issues with ACCEL_NET:

Corolla w/ new tune on master on left. Corolla w/ this branch on right:

image

image

sshane:
Done in https://github.com/commaai/opendbc/pull/1559

PR #1527, opened 2024-11-25
Honda: Odyssey RC5(JPN) support

leonunix:
add  Odyssey RC5 support
this is a car information
1.  bosch B connect. 
2. radarless car like civic 2022
3. use special scm_button and scm_feedback like nidec. not same as civic

- [route with openpilot:  ] a3fa73a8f1cd8e8d/0000000c--439df2ef4a
- [route with stock acc:  ] a3fa73a8f1cd8e8d/0000000b--777636dc85


leonunix:
@sshane Can you review this merge request? It work perfect now.

PR #1531, opened 2024-11-26
Hyundai: re-enable Optima Hybrid

JoshRomero:


adeebshihadeh:
Got a dongle ID and route? Can you upload the rlogs for the route too?

adeebshihadeh:
Please reopen with a test route.

PR #1533, opened 2024-11-27
Toyota: improve ego acceleration near stop

sshane:
Split from https://github.com/commaai/opendbc/pull/1522, enabled by https://github.com/commaai/opendbc/pull/1532

Removes the overshoot of aEgo near a stop due to kalman filter and high speed threshold wheel speed sensors. Has tiny impact on control now, but enables us to trust aEgo more in the future.

livePose.accelerationDevice.x would be more ideal as it can measure acceleration far earlier than this signal


sshane:
Here the times are similar, but you can see the lack of overshooting when stopping/starting:

image

PR #1534, opened 2024-11-27
Add KIA EV6 2025 (with HDA2)

taylor224:
Add KIA EV6 2025 (with HDA2) car fingerprint.

Car Info

KIA EV6 2025 (facelifted)
with HDA2
Earth trim in South Korea


taylor224:
It still not working. I think it is related with LFA2

taylor224:
It seems applied steer by angle command system from EV6 2025(I saw this on web news).
It looks like same with EV9. 
I applied EV9 pull request code to EV6.
I will test this code on my vehicle as soon as possible.

PR #1536, opened 2024-11-27
Toyota: error correct on future acceleration

sshane:
Split from https://github.com/commaai/opendbc/pull/1522

Toyota's longitudinal response includes some momentum that must be accounted for. Specifically, it will overshoot the request even if it is a flat step (see plot in https://github.com/commaai/opendbc/pull/1535), so you must wind down your request before aEgo reaches it.


sshane:
See before and after with this PR (the car in the [route](https://github.com/commaai/opendbc/pull/1535) is receiving the previous, green request):

image

sshane:
Not using a filter allows the jerk calculation to be more responsive:

image

sshane:
Done in https://github.com/commaai/opendbc/pull/1559

PR #1385, opened 2024-10-18
Toyota: Add SECOC longitudinal control

chrispypatt:
* Add counter for secoc acc messages like the other secoc messages
* Add new acc message creation for second ACCEL_CMD 
* Remove SECOC `openpilotLongitudinalControl` disable 

This will be accompanied by a related panda safety PR.

PR #1543, opened 2024-11-28
Draft PR, Update Chrispypatt's RAV4 Prime Long Fork

tmaidment:


adeebshihadeh:
Looks like this is just a duplicate of #1385?

tmaidment:
> Looks like this is just a duplicate of #1385?

PR #1559, opened 2024-12-07
Toyota: error correct on future acceleration

sshane:
From https://github.com/commaai/opendbc/pull/1536

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -75,6 +75,8 @@ class CarController(CarControllerBase):
     self.error_rate = FirstOrderFilter(0.0, 0.5, DT_CTRL * 3)
     self.prev_error = 0.0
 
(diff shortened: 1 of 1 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -42,7 +42,7 @@ def get_long_tune(CP, params):
   kdBP = [0.]
   kdV = [0.]
   if CP.carFingerprint in TSS2_CAR:
-    kiV = [0.5]
+    kiV = [0.25]
```
sshane, line 45: We can bump this back up to 0.5 once we figure out a good way to winddown integral for step changes using error rate). See this case where it shouldn't have wound down integral:

6483a59cd49b6650/000001cb--9815f176ae/1

image

sshane:
This reduces integral windup/overshoot, but the car still overshoots the request quite a bit. This *isn't* described by any of the PCM ACCEL_NET, so we still need some extra compensation for this:

image

Issue #1557, opened 2024-12-05
./test.sh: line 8: uv: command not found


Ssh1y:
Hello, I encountered the following problem, what is the reason?

```bash
r1pple@r1pple-virtual-machine:~/work/opendbc$ proxychains pip3 install -e .[testing,docs]
[proxychains] config file found: /etc/proxychains4.conf
[proxychains] preloading /usr/lib/x86_64-linux-gnu/libproxychains.so.4
[proxychains] DLL init: proxychains-ng 4.16
Defaulting to user installation because normal site-packages is not writeable
Obtaining file:///home/r1pple/work/opendbc
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting pandacan@ git+https://github.com/commaai/panda.git@master (from opendbc==1.0.0)
  Cloning https://github.com/commaai/panda.git (to revision master) to /tmp/pip-install-vl48v07n/pandacan_14fc4ccb55884e9ebe82eb9f71b2ea51
  Running command git clone --filter=blob:none --quiet https://github.com/commaai/panda.git /tmp/pip-install-vl48v07n/pandacan_14fc4ccb55884e9ebe82eb9f71b2ea51
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] Strict chain  ...  192.168.110.75:7890  ...  github.com:443  ...  OK
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] Strict chain  ...  192.168.110.75:7890  ...  github.com:443  ...  OK
  [proxychains] DLL init: proxychains-ng 4.16
  [proxychains] DLL init: proxychains-ng 4.16
  Resolved https://github.com/commaai/panda.git to commit c7cc2deaf046403899b5bcc964b9f48bfd508534
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: scons in /home/r1pple/.local/lib/python3.11/site-packages (from opendbc==1.0.0) (4.8.1)
[...]

PR #1565, opened 2024-12-09
check if uv is installed

royjr:
possible fix https://github.com/commaai/opendbc/issues/1557

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -7,7 +7,7 @@ cd $DIR
 # check if uv is installed
 if ! command -v uv &>/dev/null; then
     echo "'uv' is not installed. Installing 'uv'..."
-    pip install uv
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -4,6 +4,12 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
 cd $DIR
 
+# check if uv is installed
+if ! command -v uv &>/dev/null; then
+    echo "'uv' is not installed. Installing 'uv'..."
+    pip install uv
```
adeebshihadeh, line 10:
```suggestion
   curl -LsSf https://astral.sh/uv/install.sh | sh
```

PR #1577, opened 2024-12-14
Added new variant for Ioniq 6

FML128:


sshane:
What's your dongle id?

PR #1578, opened 2024-12-14
Toyota: JDM Yaris with SecOC

AlexandreSato:
The Japanese toyota yaris has the port (lateral only) almost ready, I believe that just like the rav4 prime should remain excluded from cardocs? 

0b54d0594d924cd9/00000041--51e7d1ea2e

user https://github.com/gonzalles125 
discord gon0822


jyoung8607:
Is this the one that needs the radar points parser disabled?

AlexandreSato:
> Is this the one that needs the radar points parser disabled?

Yes is a continental new radar

AlexandreSato:
- [ ] apply Toyota: Standard SecOC platform config (https://github.com/commaai/opendbc/pull/1619)

gonzalles125:
0b54d0594d924cd9/000000d3--aa729fc253/9
The route taken during the test drive after the update.

sshane:
This seems to be the hybrid variant? Can you fix the docs, or are all Yaris hybrid now?

AlexandreSato:
> This seems to be the hybrid variant? Can you fix the docs, or are all Yaris hybrid now?

Yes this is hybrid! And Yaris ICE exists in japan, my bad. Opened: https://github.com/commaai/opendbc/pull/1641

PR #1579, opened 2024-12-14
Move safety code into opendbc

nworb-cire:
Still a bit WIP, will mark ready once I've fully validated the panda end. Still need to figure out the best way of moving the safety code tests into opendbc.

I did modify a couple of the safety files, diffs as follows:
```diff 
# safety_hyundai.h
+   // forward cam to ccan and vice-versa, except lkas cmd
-   // forward cam to ccan and viceversa, except lkas cmd
```
to comply with codespell and
```diff
# safety_subaru_preglobal.h
+   const int SUBARU_PG_PARAM_REVERSED_DRIVER_TORQUE = 4;
-   const int SUBARU_PG_PARAM_REVERSED_DRIVER_TORQUE = 1;
```
to keep all of the Subaru panda flags in the same class.

Other changes include moving the UDS script into opendbc and the examples and `panda_runner.py` file out of opendbc. I moved them to openpilot, so expect a PR there once this is ready. Shouldn't be strictly necessary for basic functionality.


adeebshihadeh:
> I did modify a couple of the safety files, diffs as follows:

Can you get any diff merged into panda first such that we can just `mv panda/board/safety/ openpdbc_repo/opendbc/safety/`? It's a ton easier to review that way.

opendbc/car/ford/values.py
```diff
@@ -42,6 +42,11 @@ def __init__(self, CP):
     pass
 
 
+class FordPandaFlags(IntFlag):
```
adeebshihadeh, line 45: this is something we should be able to split out and merge on its own

opendbc/car/panda_runner.py
adeebshihadeh, line 1: why is this deleted?

adeebshihadeh (commented):
Thanks for picking this up. Happy to provide feedback as often as you'd like.

Per usual, it's best if we can break things up into small, self-contained PRs. In this case, I think we can start with the Python constants.

opendbc/car/__init__.py
```diff
@@ -284,6 +284,7 @@ class PlatformConfigBase(Freezable):
   flags: int = 0
 
   platform_str: str | None = None
+  safety_file: str = "safety_defaults.h"
```
sshane, line 287: shouldn't be a need for this, safetyParam should be all you need

sshane:
we don't just want to copy paste the files (that is the end goal however), but we also need to build a standalone library and test it here, going to close since this is just a placeholder PR

PR #1583, opened 2024-12-15
Toyota: raise long ki again

sshane:


AlexandreSato:
I used it for two days, for a few hours I reversed the commit to compare. I convinced myself that 0.5 got better, more responsive

sshane:
With the future aEgo in the PID, this is too noisy. We can decrease the extrapolation we due, which we're starting in: https://github.com/commaai/opendbc/pull/1665

PR #1587, opened 2024-12-17
Update fingerprints.py for Hyundai Ioniq 2017 EU

BroodPitt:
i add my fingerprint values for the Hyundai Ioniq 2017 EU model

route: 45aa7b8203ee032e/00000002--2e650caa1c/1
dongleid: 45aa7b8203ee032e


sshane:
It looks like the radar disable doesn't work for your car, it wakes back up, however we don't support alpha long in release so I am going to merge this.

PR #1589, opened 2024-12-17
Move car-specific panda flags into opendbc

nworb-cire:
Requires https://github.com/commaai/panda/pull/2115 due to merging the Subaru platform flags


sshane:
thanks!

PR #1591, opened 2024-12-18
Publish to pypi

nworb-cire:
Invokes [this action](https://github.com/marketplace/actions/python-auto-release-pypi-github) if tests pass on master branch. Needs Comma to create some repo secrets.

Note that I'm generally unable to thoroughly test if this works, due to the nature of CI.


sshane:
done in https://github.com/commaai/opendbc/pull/1707 with direct publishing instead

PR #1595, opened 2024-12-22
Remove STEER_MAX limit for Genesis G90

Maddog1929:
The 2018 Genesis G90 should not be subject to the 255 steer max limit as it is fully capable of requesting the default 384 without error. I have been driving with this change since February with over 7k miles without issue. Attaching most recent route.

Car
2018 Genesis G90

Route
590a1a98ea35681c/2024-12-20--16-07-15/0

Including debug of car commanding and holding steer value of 384 without backing off in a tight freeway exit turn
image


Maddog1929:
Including debug of car commanding and holding steer value of 384 without backing off in a tight freeway exit turn
image

PR #1598, opened 2024-12-22
`cp panda/python/uds.py opendbc/car/uds.py`

nworb-cire:
Merge https://github.com/commaai/panda/pull/2118 first for the commit message to be true


adeebshihadeh:
Let's put this in `opendbc/car/`?

Issue #1184, opened 2024-08-27
pytest: slow collection for `car/tests/`


adeebshihadeh:
`can/tests/` is 20x faster to collect than `car/tests/`. 
```
batman:opendbc$ pytest --co can/tests/
============ 22 tests collected in 0.05s ============ 
batman:opendbc$ pytest --co car/tests/
============ 1627 tests collected in 1.04s ============ 
```

My first suspicion is the use of [parameterized](https://pypi.org/project/parameterized/). Parameterizing is super important for running all the tests on different car models, so it's worth confirming that it's this and making it faster. Is it faster to just use a for loop?

roelofvandijk:
It's all the imports.

E.g.
`pyinstrument -m pytest --cache-clear --collect-only opendbc/car/tests/test_lateral_limits.py`

```
================================================================================ 430 tests collected in 0.29s =================================================================================

  _     ._   __/__   _ _  _  _ _/_   Recorded: 22:02:12  Samples:  358
 /_//_/// /_\ / //_// / //_'/ //     Duration: 0.401     CPU time: 0.398
/   _/                      v5.0.0

Program: pytest --cache-clear --collect-only opendbc/car/tests/test_lateral_limits.py

0.354 <module>  pytest/__main__.py:1
└─ 0.354 console_main  _pytest/config/__init__.py:194
      [108 frames hidden]  _pytest, pluggy, inspect, importlib, ...
         0.201 import_module  importlib/__init__.py:73
         ├─ 0.141 AssertionRewritingHook.exec_module  _pytest/assertion/rewrite.py:144
         │  └─ 0.139 <module>  opendbc/car/tests/test_lateral_limits.py:1
         │     └─ 0.136 <module>  opendbc/car/car_helpers.py:1
         │        ├─ 0.064 <module>  opendbc/car/fingerprints.py:1
         │        │  ├─ 0.039 <module>  opendbc/car/interfaces.py:1
         │        │  │  └─ 0.039 <module>  opendbc/car/values.py:1
         │        │  │     ├─ 0.014 compile  <built-in>
         │        │  │     └─ 0.004 <module>  opendbc/car/volkswagen/values.py:1
         │        │  └─ 0.022 get_interface_attr  opendbc/car/interfaces.py:378
         │        │     └─ 0.015 compile  <built-in>
         │        ├─ 0.059 load_interfaces  opendbc/car/car_helpers.py:16
[...]

PR #1599, opened 2024-12-23
Improve get_interface_attr speed through caching

fahhem:
Inspired by https://github.com/commaai/opendbc/issues/1184 but only a small step.

To me, it looks like the car/ module does a lot of (potentially unnecessary) work at import time, such as defining dataclasses for documentation of a car. As a result, collecting the tests has to not only import class and function definitions, but also do real work.

In this case, `interfaces.get_interface_attr()` walks a folder structure ignoring some folders, dynamically imports modules, does some optional dict-merging, and returns. During test collection, it's called 3 times, each time walking the directory structure and importing the same module.

It seems to me that actually closing #1184 would entail a dozen or more of PRs like this: finding a small improvement to the import-time work or moving it either to runtime or maybe even "compilation".

### DIFF
--- a/opendbc/car/interfaces.py
+++ b/opendbc/car/interfaces.py
@@ -375,28 +375,42 @@ INTERFACE_ATTR_FILE = {
 
 # interface-specific helpers
 
+_brand_values = {}  # attr_file => brand_name => module
+_brand_names = None
 def get_interface_attr(attr: str, combine_brands: bool = False, ignore_none: bool = False) -> dict[str | StrEnum, Any]:
+  global _brand_names
   # read all the folders in opendbc/car and return a dict where:
   # - keys are all the car models or brand names
   # - values are attr values from all car folders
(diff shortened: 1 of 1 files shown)


sshane:
Did you test the difference of this PR? I see a very little difference (1.05s to 0.98s) for me. Also, we can likely just use functools.cache if this were significant

fahhem:
Hi @sshane,

Yes, I did test it, and it was minimal. I implied it when I mentioned "It seems to me that actually closing https://github.com/commaai/opendbc/issues/1184 would entail a dozen or more of PRs like this." since the performance target that was set is so far from where you are, and these PRs will each do incremental improvements.

The roughly 6% (or 60ms) this saves is definitely small, but there was no smoking gun in the traces, it's a hundred small cuts.

sshane:
Can you switch it to functools.cache?

PR #1601, opened 2024-12-24
Update fingerprints.py for 2024 Hyundai Tucson

Curtrahmel:
Added " b'\xf1\x00NX4 FR_CMR AT USA LHD 1.00 1.00 99211-CW020 14Z', "
   --> to " CAR.TUCSON_4TH_GEN: {
    (Ecu.fwdCamera, 0x7c4, None): [ "


Curtrahmel:
I don't know what is blocking it, but before they switched the location of the fingerprint file, it was easy to make the changes to it so I could take advantage of the improvements. Now, due to inexperience, I am not able to use my device on the latest os. Any help would be greatly appreciated.

staeiou:
Looking into it more. This seems to be failing because there is a blacklist against the "CW" part of the fingerprint. There is a comment: " # Asserts no ECUs known to be shared across platforms exist in the database.
    # Tucson having Santa Cruz camera and EPS for example" https://github.com/commaai/opendbc/blob/bc2143ba93ec2d78d04f99eed27f2677b0d665dd/opendbc/car/hyundai/tests/test_hyundai.py#L94

Curtrahmel:
Is there anything that can be done to fix it? I have used that for over 19000 miles with no major issues.

staeiou:
The PR that added the blacklist is https://github.com/commaai/openpilot/pull/28318 

@sshane @sunnyhaibin it looks like you were involved in this, can you help us out? There are two of us that have 2024 Tucsons with a fingerprint that has the "CW" in the blacklist.
 
Route: 6d9d843a1b27bc98|00000003--6558424d55

PR #1605, opened 2024-12-30
scons failure fixed

MikeBusuttil:
- building no longer fails
  - by setting `PYTHONPATH` env var within scons (where it matters)
- docs simplified

#### Bug Description:
```bash
$ scons                   # ❌ fails
$ PYTHONPATH=$(pwd) scons # ✅ succeeds
```
image

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -3,11 +3,13 @@ import subprocess
 import sysconfig
 import platform
 import numpy as np
+from pathlib import Path
 
 arch = subprocess.check_output(["uname", "-m"], encoding='utf8').rstrip()
 if platform.system() == "Darwin":
   arch = "Darwin"
 
+os.environ['PYTHONPATH'] = str(Path(sysconfig.get_paths()['data']).parent)
 python_path = sysconfig.get_paths()['include']
 cpppath = [
   '#',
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -17,15 +11,15 @@ uv sync --all-extras
...
+uv run scons -j$(nproc 2>/dev/null || sysctl -n hw.logicalcpu)
 
 # *** lint ***
 # TODO: pre-commit is slow; replace it with openpilot's "op lint"
 #uv run pre-commit run --all-files
 uv run ruff check .
 
 # *** test ***
-uv run pytest -n8
+uv run pytest -n$(nproc 2>/dev/null || sysctl -n hw.logicalcpu)
```
adeebshihadeh, line 22: this was intentional. the tests actually run slower with more cores since pytest has tons of overhead on distributing the tests

SConstruct
```diff
@@ -3,11 +3,13 @@ import subprocess
 import sysconfig
 import platform
 import numpy as np
+from pathlib import Path
 
 arch = subprocess.check_output(["uname", "-m"], encoding='utf8').rstrip()
 if platform.system() == "Darwin":
   arch = "Darwin"
 
+os.environ['PYTHONPATH'] = str(Path(sysconfig.get_paths()['data']).parent)
```
adeebshihadeh, line 12: this isn't the right fix. this should already be set before you get to this point

PR #1614, opened 2025-01-07
Improve get_interface_attr speed through caching

fahhem:
From #1599:

> Inspired by https://github.com/commaai/opendbc/issues/1184 but only a small step.
> 
> To me, it looks like the car/ module does a lot of (potentially unnecessary) work at import time, such as defining dataclasses for documentation of a car. As a result, collecting the tests has to not only import class and function definitions, but also do real work.
> 
> In this case, interfaces.get_interface_attr() walks a folder structure ignoring some folders, dynamically imports modules, does some optional dict-merging, and returns. During test collection, it's called 3 times, each time walking the directory structure and importing the same module.
> 
> It seems to me that actually closing https://github.com/commaai/opendbc/issues/1184 would entail a dozen or more of PRs like this: finding a small improvement to the import-time work or moving it either to runtime or maybe even "compilation".

I switched to `functools.cache` as requested by @sshane. Speed is still very close, might be slightly faster, but definitely in the noise. Sometimes 0.88s, sometimes 0.93s, but on master I get a similar spread. I'm not sure it's worth analyzing the time much further.


opendbc/car/interfaces.py
```diff
@@ -375,28 +375,43 @@ def update(self, CC: structs.CarControl, CS: CarStateBase, now_nanos: int) -> tu
...
+
+
+@cache
+def _get_brand_values_for_attr(brand_name: str, attr: str) -> Any | None:
+  attr_file = INTERFACE_ATTR_FILE.get(attr, "values")
+  try:
+    return __import__(f'opendbc.car.{brand_name}.{attr_file}', fromlist=[attr])
+  except (ImportError, OSError):
+    pass
+  return None
```
sshane, line 390: You could import each values file once instead of several per brand, further reducing the import calls.

fahhem: As in, look up the `attr_file` and then cache based on that rather than the raw `attr`? Because if the `attr_file` is different, then I assume the imports aren't overlapping, right?

sshane:
closing since this adds a bunch of lines without much speedup atm

PR #1616, opened 2025-01-09
2022 Honda Civic Torque Tuning

MarcoTheDingo:
Updated the 11th gen with better tuning and torqued live tuning. Then changed actuator delay to 0.5 to make it match planner. Much better responsiveness overall, but especially at low speeds.

Stock route: 5130484aa8069bad/00000000--5da80373ec

Long torqued route with 0.4 delay: 5130484aa8069bad/0000004e--ab53601964
Short torqued route with 0.5 delay: 5130484aa8069bad/00000002--a454a60ba7

Requires: https://github.com/commaai/openpilot/pull/34341


jyoung8607:
I'm not sure about this change yet. Some notes:

Your actuator delay of 0.5 seems very excessive. The prior delay of 0.1 looks just about right, even 0.2 looks to be too much. I could believe that your actuator delay setting might produce a subjectively appealing turn-in experience, but at the level of controls tuning, we need to faithfully execute the lateral plan we get from the model.

image

steerRatio is wandering around, often quite far from your initial default value, and tireStiffnessFactor is running away as described in commaai/openpilot#33074. This is not your fault, but it's going to make tuning a bit of a challenge.

The torque learner is applying a value considerably higher than your configured value of 1.25, so in the real world this is going to drive a lot different for the first 30 minutes or so after a reset/reinstall. I've found that applying the live-learned values at this stage of tuning may be counterproductive. For now I'd suggest removing that change in openpilot to make it so that its learned value isn't applied live, but it's always running and observing anyway, and you can update the manual config based on its logs later.

**TL;DR:** I think this one needs to go back to the lab. FWIW, I think the lateral accel torque controller (and maybe paramsd as well) may just not get along well with progressive ratio steering. Volkswagen is also stuck on the old PIF controller due to basically the same problems you're having. I have ideas for how we might adapt the controller to work better for us, but it's going to take some work.

MarcoTheDingo:
@jyoung8607 made a new branch https://github.com/MarcoTheDingo/opendbc/tree/civic-tune with just the better tuning parameters if that looks good enough to merge or advice on how to improve. Still has some oscillation at lower speeds, but generally a bit better. Thanks!
5130484aa8069bad/0000000d--c2f48762b0

jrgiacone:
@jyoung8607 Is this something we can revisit and maybe someone could help guide me in the right direction on? I have a 26 civic hybrid and love to tinker

PR #1622, opened 2025-01-12
GV70 2025 Electric

morrislee:
Adding fingerprints for 2025 GV70 fingerprint by @phynicle


morrislee:
I believe this is all is needed to add support for GV70 2025 described by phynicle

sshane:
What is the dongle id?

morrislee:
> What is the dongle id?

waiting for a route, I am getting mixed info. @phynicle says just needs fingerprint, sunny says it needs more. Need a route to confirm

PR #1623, opened 2025-01-12
Car Port add : Genesis GV70 2.5T 2025 korea (HDA1)

chanhojung:


adeebshihadeh:
What's your dongle ID?

PR #1624, opened 2025-01-12
GM: FPv3 based on VIN

garrettpall:


sshane:
Have we evaluated if we need the VIN to match and not the camera FW versions we now gather?

opendbc/car/gm/values.py
```diff
@@ -58,6 +58,37 @@ def __init__(self, CP):
...
+    _2016 = "G"
+    _2017 = "H"
+    _2018 = "J"
+    _2019 = "K"
+    _2020 = "L"
+    _2021 = "M"
+    _2022 = "N"
+    _2023 = "P"
+    _2024 = "R"
+    _2025 = "S"
```
sshane, line 90: We should probably make this common (not live in gm/) as it's standard for almost all VINs, or use a small VIN library.

opendbc/car/gm/values.py
```diff
@@ -58,6 +58,37 @@ def __init__(self, CP):
...
+    class CHEVROLET:
+        TRUCK = {"1GC", "2GC", "3GC"}
+        MPV = {"1G1", "1GN", "2G1", "2GN", "3G1", "3GN", "KL1", "KL7"}
+
+    class GMC:
+        TRUCK = {"1GT", "2GT", "3GT"}
+        MPV = {"1GK", "2GK", "3GK"}
+
+    class CADILLAC:
+        MPV = {"1G6", "1GY", "2G6", "3GY"}
```
sshane, line 71: Two spaces for indents

opendbc/car/gm/values.py
```diff
@@ -147,39 +182,80 @@ class CAR(Platforms):
       GMCarDocs("Chevrolet Bolt EV 2022-23", "2LT Trim with Adaptive Cruise Control Package"),
     ],
     GMCarSpecs(mass=1669, wheelbase=2.63779, steerRatio=16.8, centerToFrontRatio=0.4, tireStiffnessFactor=1.0),
+    wmis = WMI.CHEVROLET.MPV,
+    fourth_digits = {"F"},
```
sshane, line 186: What is the fourth digit? Why do we sometimes require a match, and not other times? If it's a platform code, we should call it that.

opendbc/car/gm/values.py
```diff
@@ -58,6 +58,37 @@ def __init__(self, CP):
...
+
+    class CADILLAC:
+        MPV = {"1G6", "1GY", "2G6", "3GY"}
+
+class ANY:
+  ALL = {
+    "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M",
+    "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z",
+    "0", "1", "2", "3", "4", "5", "6", "7", "8", "9"
+    }
```
sshane, line 78: ANY isn't a great name, but can we just specify all the "fourth digits" that may match per platform? That is safer, preventing matching with a completely new/different vehicle.

opendbc/car/gm/values.py
sshane, line 71: Do we need to separate these into subbrands? VW gets away without it.

opendbc/car/gm/values.py
```diff
@@ -147,39 +182,80 @@ class CAR(Platforms):
...
+    fourth_digits = {"R"},
+    fifth_digits = ANY.ALL,
+    years = {YEARS._2019},
   )
   CHEVROLET_TRAVERSE = GMSDGMPlatformConfig(
     [GMCarDocs("Chevrolet Traverse 2022-23", "RS, Premier, or High Country Trim")],
     GMCarSpecs(mass=1955, wheelbase=3.07, steerRatio=17.9, centerToFrontRatio=0.4),
+    wmis = WMI.CHEVROLET.MPV,
+    fourth_digits = ANY.ALL,
+    fifth_digits = {"R", "V"},
```
sshane, line 238: What is fifth digits (better name)?

sshane (changes requested):

garrettpall:
> Have we evaluated if we need the VIN to match and not the camera FW versions we now gather?

Currently there is no way for me to look at the firmware’s collected as the public database hasn’t been updated since after the query has been implemented. 

I put this together based off a comment from Jason: https://discord.com/channels/469524606043160576/1282121586002104402/1327798185460240457

I like the idea of fingerprinting off the vin because it is something that can be decoded and has a meaning behind it instead of firmware versions being in a lookup database.

Once the public dataset is updated, I can look into firmware versions.

opendbc/car/gm/values.py
garrettpall: The forth digit is the platform code for smaller vehicles while the five digit is the platform code for larger vehicles (4th is then weight class). I can look into renaming these and making that logic cleaner.

opendbc/car/gm/values.py
garrettpall: I was trying to find the lowest possible uniquely identifying digits to avoid. Having some cars not fingerprint. I can see if this is consistent once the public dataset is updated.

opendbc/car/gm/values.py
garrettpall: Without year and the WMI, the same model code could be all the following:
GMC Acadia Awd '10-'16, Acadia Limited Awd '17, Saturn Outlook Awd '10, Buick Enclave Awd '10-'17, Chevy Traverse Awd '10-'17, Buick Enclave Awd '18-'24, Chevy Traverse Awd '18-'23.

Brand is useful here to determine platform. Truck vs MPV is needed so small vehicles using the forth digit platform code don’t match as the forth digit can be the weight class for a truck. Year helps us control some things, but I would be open to leaving that more forgiving.

morrislee:
Current master/release does not recognize a 2022 Trailblazer, this PR picks up and fingerprints properly as a Trailblazer correctly

Here is a route for a 2022 Trailblazer working:
f97ecb0c37e04cc2/00000000--7ce0807334/0

morrislee:
Do we need more validation for this to be merged? if so, what else is needed?

garrettpall:
> Do we need more validation for this to be merged? if so, what else is needed?

I need the comma public segments to get updated to make sure this works for all GMs.

opendbc/car/gm/values.py
```diff
@@ -147,39 +182,80 @@ class CAR(Platforms):
...
   )
   GMC_YUKON = GMPlatformConfig(
     [GMCarDocs("GMC Yukon 2019-20", "Adaptive Cruise Control (ACC) & LKAS")],
     GMCarSpecs(mass=2490, wheelbase=2.94, steerRatio=17.3, centerToFrontRatio=0.5, tireStiffnessFactor=1.0),
   )
 
+def match_fw_to_car_fuzzy(live_fw_versions, vin, offline_fw_versions) -> set[str]:
+  candidates = set()
+# Check the WMI and chassis code to determine the platform
+  wmi = vin[:3]
```
sshane, line 249: Use our VIN class

opendbc/car/gm/values.py
```diff
@@ -58,6 +58,37 @@ def __init__(self, CP):
     self.BRAKE_LOOKUP_BP = [self.ACCEL_MIN, max_regen_acceleration]
     self.BRAKE_LOOKUP_V = [self.MAX_BRAKE, 0.]
 
+class WMI:
+    class CHEVROLET:
+        TRUCK = {"1GC", "2GC", "3GC"}
+        MPV = {"1G1", "1GN", "2G1", "2GN", "3G1", "3GN", "KL1", "KL7"}
```
sshane, line 64: we use 2 spaces for indent

opendbc/car/gm/fingerprints.py
```diff
@@ -45,27 +42,6 @@
   CAR.CADILLAC_ESCALADE_ESV_2019: [{
     715: 8, 840: 5, 717: 5, 869: 4, 880: 6, 289: 8, 454: 8, 842: 5, 460: 5, 463: 3, 801: 8, 170: 8, 190: 6, 241: 6, 201: 8, 417: 7, 211: 2, 419: 1, 398: 8, 426: 7, 487: 8, 442: 8, 451: 8, 452: 8, 453: 6, 479: 3, 311: 8, 500: 6, 647: 6, 193: 8, 707: 8, 197: 8, 209: 7, 199: 4, 455: 7, 313: 8, 481: 7, 485: 8, 489: 8, 249: 8, 393: 7, 407: 7, 413: 8, 422: 4, 431: 8, 501: 8, 499: 3, 810: 8, 508: 8, 381: 8, 462: 4, 532: 6, 562: 8, 386: 8, 761: 7, 573: 1, 554: 3, 719: 5, 560: 8, 1279: 4, 388: 8, 288: 5, 1005: 6, 497: 8, 844: 8, 961: 8, 967: 4, 977: 8, 979: 8, 985: 5, 1001: 8, 1017: 8, 1019: 2, 1020: 8, 1217: 8, 510: 8, 866: 4, 304: 1, 969: 8, 384: 4, 1033: 7, 1009: 8, 1034: 7, 1296: 4, 1930: 7, 1105: 5, 1013: 5, 1225: 7, 1919: 7, 320: 3, 534: 2, 352: 5, 298: 8, 1223: 2, 1233: 8, 608: 8, 1265: 8, 609: 6, 1267: 1, 1417: 8, 610: 6, 1906: 7, 611: 6, 612: 8, 613: 8, 208: 8, 564: 5, 309: 8, 1221: 5, 1280: 4, 1249: 8, 1907: 7, 1257: 6, 1300: 8, 1920: 7, 563: 5, 1322: 6, 1323: 4, 1328: 4, 1917: 7, 328: 1, 1912: 7, 1914: 7, 804: 3, 1918: 7
   }],
-  CAR.CHEVROLET_BOLT_EUV: [{
```
sshane, line 48: let's keep these around for now, remove in follow up PR

opendbc/car/gm/values.py
sshane, line 64: can you add a link to some documentation for these codes? wikipedia, forum, etc.

opendbc/car/gm/values.py
sshane, line 90:
```suggestion
    S_2025 = "S"
```

this if anything. note Rivian has the same class, so we should make it common

opendbc/car/gm/values.py
```diff
@@ -85,6 +116,10 @@ class GMPlatformConfig(PlatformConfig):
     Bus.radar: 'gm_global_a_object',
     Bus.chassis: 'gm_global_a_chassis',
   })
+  wmis: set[str] = field(default_factory=set)
+  years: set[str] = field(default_factory=set)
+  fourth_digits: set[str] = field(default_factory=set) # Passanger cars , See https://en.m.wikibooks.org/wiki/Vehicle_Identification_Numbers_(VIN_codes)/GM/VIN_Codes
```
sshane, line 121:
```suggestion
  fourth_digits: set[str] = field(default_factory=set)  # Passenger cars, see https://en.m.wikibooks.org/wiki/Vehicle_Identification_Numbers_(VIN_codes)/GM/VIN_Codes
```

opendbc/car/gm/values.py
```diff
@@ -85,6 +116,10 @@ class GMPlatformConfig(PlatformConfig):
...
+  fifth_digits: set[str] = field(default_factory=set) # SUV/Truck
```
sshane, line 122:
```suggestion
  fifth_digits: set[str] = field(default_factory=set)  # SUV/Truck
```

morrislee:
Do we have any plans in the near future? This has been over a year, hoping we can get some more accurate fingerprinting in place to fix the rest of Gm fleet controls

jyoung8607 (changes requested):
Reopened since there's been some recent development for GM lateral controls, and I think this PR was blocking application of those improvements for some platforms.

@garrettpall Haven't gone line-by-line yet, but overall this looks reasonable.

I agree with the move to generalize ModelYear, and the way you've done it. Can you open a standalone PR with just that refactor and the Rivian port touches? We can review and merge that quickly. Then, this PR will only have GM FPv3 changes.

morrislee:
what do we need to do to validate this and have it merged? I am hoping to not have the code change too much and this is out of date

morrislee:
Following up on this seeing what I can do to help to get this merged

PR #1626, opened 2025-01-12
Got rid of openpilot.common.numpy_fast as per https://github.com/commaai/openpilot/issues/34331

Sammohana:


opendbc/car/__init__.py
```diff
@@ -9,7 +9,7 @@
 from opendbc.car import structs
 from opendbc.car.can_definitions import CanData
 from opendbc.car.docs_definitions import CarDocs, ExtraCarDocs
-from opendbc.car.common.numpy_fast import clip, interp
+from numpy import clip, interp
```
adeebshihadeh, line 12: let's stay consistent with `import numpy as np`

adeebshihadeh (changes requested):

opendbc/car/__init__.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
 # functions common among cars
 import logging
```
adeebshihadeh, line 3:
```suggestion
# functions common among cars
import logging
import numpy as np
```

PR #1627, opened 2025-01-12
Toyota: Add secoc distance message from PCM

chrispypatt:


sshane:
Can you provide a drive where you press the distance button?

chrispypatt:
Are you looking for a public route added in the PR? Or can I share one just with you for the purpose of this PR?

chrispypatt:
Here’s a route where I press the button several times 41ba5b181f29435d/00000036--10383e82d4

sshane:
This is not sent from the camera, so we cannot send it to increment the distance. `cp.vl["PCM_CRUISE_2"]["PCM_FOLLOW_DISTANCE"]` still works to get the value, so this seems un-needed

sshane:
You can rename the message to be prefixed like PCM_CRUISE_X and we can merge that though

chrispypatt:
Yeah. We won't be able to use this the same way it is used today for other cars. It also looks like the message has secoc but it's a little harder to tell because the freq of the message is much lower than other secoc messages. Still worth documenting the message in the DBC IMO. I will update the name per your suggestion

chrispypatt:
Even if we can't send this message to change the distance bars, would it be acceptable to still hookup this button press to change personality? It could get out of sync with the dash but at least would give the ability to change personality without fishing through the settings menu.

opendbc/dbc/generator/toyota/toyota_secoc_pt.dbc
```diff
@@ -67,6 +67,17 @@ BO_ 418 CRUISE_RELATED: 8 XXX
  SG_ CRUISE_ACTIVE : 7|1@0+ (1,0) [0|1] "" XXX
  SG_ CHECKSUM : 63|8@0+ (1,0) [0|255] "" XXX
 
+BO_ 589 PCM_CRUISE_4: 8 XXX
```
sshane, line 70: Not certain, but better than DSU/camera or XXX

```suggestion
BO_ 589 PCM_CRUISE_4: 8 PCM
```

sshane: Could also be the radar, but seems odd if the camera is sending ACC messages

opendbc/dbc/generator/toyota/toyota_secoc_pt.dbc
```diff
@@ -138,7 +149,6 @@ CM_ SG_ 1014 ADJACENT_ENABLED "when BSM is enabled in settings, this is on along
 CM_ SG_ 1014 L_APPROACHING "vehicle approaching from left side of car. enabled above 10mph, regardless of ADJACENT_ENABLED or APPROACHING_ENABLED";
 CM_ SG_ 1014 R_APPROACHING "vehicle approaching from right side of car. enabled above 10mph, regardless of ADJACENT_ENABLED or APPROACHING_ENABLED";
 CM_ SG_ 1014 APPROACHING_ENABLED "when BSM is enabled in settings, this is on along with ADJACENT_ENABLED. this controls bsm alert visibility";
```
sshane, line 151:
```suggestion
CM_ SG_ 1014 APPROACHING_ENABLED "when BSM is enabled in settings, this is on along with ADJACENT_ENABLED. this controls bsm alert visibility";

```

sshane:
That won't work, any changes in the UI won't be reflected in the UI/take effect

chrispypatt:
But we could instead have comma listen to the follow distance and update the UI based on the current cars state instead right?

PR #1635, opened 2025-01-14
Toyota: remove longitudinal derivative

sshane:
Doesn't do much. The future estimated aEgo we use for error correction provides a similar effect to derivative but is more pronounced and less noisy at larger scales, so we'll just keep that.

On the bottom of each plot, orange is derivative **removed** (this PR), green is the real route with current derivative.

Highway coasting: image
Highway accel: image

Medium speed accel: image
Medium speed braking: image


sshane:
Step response effect is there, but minimal. Compare orange line on bottom plots (left is d, right is no d)

image

PR #1636, opened 2025-01-14
Tesla: add model years and TeslaCarDocs

lukasloetkolben:


sshane:
know if 2025 is supported?

lukasloetkolben:
> thanks! know if 2025 is supported?

Yes, it's very likely since the Model 3 hasn't changed. Someone already bought the harness for a 2025 Model 3, but it is not yet confirmed. The new Model Y Juniper, released recently, should also work with the current code.

PR #1653, opened 2025-01-22
Revert Yaris SecOC for now

sshane:
missing test route and lateral tune is questionable. we should only merge new cars on the torque tune unless there's a valid reason.


sshane:
@AlexandreSato feel free to re-open this PR with some fixes

AlexandreSato:
> @AlexandreSato feel free to re-open this PR

opened #1668

PR #1660, opened 2025-01-23
Add European (2020) HYUNDAI_KONA_HEV fp

twilsonco:
This is for @elruidomepeta's (on Discord) 2020 Hyundai Kona HEV in Spain. 

Route of successful recognition on `master` with these changes: 69e57686816b982c/00000010--18ca001831/0


twilsonco:
So the fingerprint works fine, but OP has no long control. Inspecting logs, it sends commands but the car doesn't respond. 

Comparing commands sent by SCC with those sent by OP, I found some differences that may be responsible for it not responding to OP ACC commands. I'll see if I can get that resolved, then this PR with coincide with one to make the necessary changes to hyundaican.py.

twilsonco:
I wasn't able to resolve the issues with long control by trying to better mimic the SCC commands the car was using when in dashcam mode. 

I'm submitting this as is so that @elruidomepeta can run `master`. After that, I'll work with him to record fresh routes to capture the problem and we'll open a bug report.

PR #1664, opened 2025-01-25
Toyota: fix acceleration discontinuity when stopping

sshane:
Regression from https://github.com/commaai/opendbc/pull/1532

Due to Toyota PCM's overzealous processing of ACC acceleration requests, the small jump that can occur when stopping can cause the brake controller to overreact and let go of the brakes entirely. This is only possible if the PID controller is applying a smaller acceleration than feedforward.

This issue at least affects hybrids, as seen in this case:

https://connect.comma.ai/d9b97c1d3b8c39b2/00000020--177f5572ab/1346/1362

image

This usually does not happen as [longcontrol.py starts with the last output accel when stopping](https://github.com/commaai/openpilot/blob/540c45bfec37266d943a1433bff82c9f88068fb5/selfdrive/controls/lib/longcontrol.py#L72), but Toyota is the only brand that has its final PID controller in carcontroller due to custom tuning/logic.

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -220,14 +220,15 @@ class CarController(CarControllerBase):
         j_ego = (self.aego.x - prev_aego) / (DT_CTRL * 3)
         a_ego_future = a_ego_blended + j_ego * 0.5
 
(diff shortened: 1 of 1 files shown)


sshane:
After:

image

The integral unwind we just added slowly corrects for the offset which is desireable.

PR #1575, opened 2024-12-14
HKG: ccNC port (non-HDA2)

royjr:
ccnc

> [!NOTE]
> This port supports the following ccNC-equipped vehicles:
>
> - Hyundai Kona (without HDA II) 2024-25
> - Hyundai Kona Hybrid (without HDA II) 2024
> - Hyundai Kona Electric (without HDA II) 2024
> - Hyundai Santa Cruz (without HDA II) 2025
> - Hyundai Sonata (without HDA II) 2024-25
> - Hyundai Sonata Hybrid (without HDA II) 2024-25
> - Hyundai Tucson Hybrid (without HDA II) 2025
> - Kia K4 (without HDA II) 2025
> - Kia K5 (without HDA II) 2025
> - Kia Sorento (without HDA II) 2024-25

PR #1671, opened 2025-01-28
HKG: ccNC messages

royjr:
New messages required by https://github.com/commaai/opendbc/pull/1575 and used by newer hyundai cars


opendbc/dbc/hyundai_canfd.dbc
```diff
@@ -160,6 +160,7 @@ BO_ 298 LFA: 16 ADRV
  SG_ LKA_ASSIST : 62|1@1+ (1,0) [0|1] "" XXX
  SG_ LKA_MODE : 24|3@1+ (1,0) [0|7] "" XXX
  SG_ NEW_SIGNAL_2 : 70|2@0+ (1,0) [0|3] "" XXX
+ SG_ NEW_SIGNAL_4 : 72|4@1+ (1,0) [0|15] "" XXX
```
sshane, line 163: Can we give these semi-relevant, self-describing names?

opendbc/dbc/hyundai_canfd.dbc
```diff
@@ -658,6 +662,139 @@ BO_ 1264 LOCAL_TIME: 8 XXX
  SG_ MINUTES : 21|6@0+ (1,0) [0|63] "" XXX
  SG_ SECONDS : 31|8@0+ (1,0) [0|59] "" XXX
 
+BO_ 353 MSG_161: 32 CCNC
```
sshane, line 665: Same

sshane: May be a good idea  to start thinking about separating the DBCs, and import hyundai_canfd if it's a superset for architecture clarity. For now you can give these a good suffix/prefix to differentiate them

opendbc/dbc/hyundai_canfd.dbc
royjr: The added [`NEW_SIGNAL_4`](https://github.com/commaai/opendbc/pull/1575/files#diff-256b6a0bc4c776172e296de10551d95f65f5013dbbf8364478d35f07fffc0828R47) & [`NEW_SIGNAL_5`](https://github.com/commaai/opendbc/pull/1575/files#diff-256b6a0bc4c776172e296de10551d95f65f5013dbbf8364478d35f07fffc0828R174) are like the other `NEW_SIGNAL_`'s defined, just signals we set to match stock. Don't have anything other than that at the moment.

opendbc/dbc/hyundai_canfd.dbc
royjr: > Same

Done

opendbc/dbc/hyundai_canfd.dbc
sshane: I know, but they should ideally be as descriptive as possible, or comments added in the DBC to hint at what we suspect t hey might do.

royjr: Added comments, values seem to always be the same, graph shows car on, off, then on again. We can remove if this isn't good enough. Screenshot 2025-01-27 at 23 10 45

PR #1675, opened 2025-01-28
2022 Honda Civic PI Tuning

MarcoTheDingo:
Updated the 11th gen with better tuning while on the PI controller. Main goal was to reduce the wobble at lower speeds (<30 MPH) as well as better corners (less corrections and turning into the turn on time). Changed kpV from 0.8 to 0.5, then kiV from 0.24 to 0.125.

Managed to get it 90% of the way there. Most wobbles are gone, but there still is a small amount of wobble at <20 MPH, but possibly could be due to the model. @jyoung8607 also suggests potentially using a breakpoint to change the tuning values at different speeds. I am not sure how to implement it, but if an openpilot maintainer has any suggestions, please let me know.

Stock route: 5130484aa8069bad/00000000--5da80373ec
New route: 5130484aa8069bad/00000032--bcc862d657


MarcoTheDingo:
@jyoung8607 are you able to review this or determine if there is another maintainer who can? I would love for this to be merged so I can re-contribute my routes to openpilot. Thanks!

MarcoTheDingo:
@adeebshihadeh as discussed on the weekly PR review a few weeks ago, I have had several users test this and all report better turning and less wobbles at slow speed. Works with 22 Civic (mine) as well as 25 Civic Hybrids. Routes updated above if you can review.

adeebshihadeh:
Added to the release milestone. @sshane want to give this a quick check before merging?

PR #1676, opened 2025-01-29
HKG: Car Port Hyundai Ioniq 5 PE (HDA2 & LFA2)

devtekve:
> [!NOTE]
> * Needs: https://github.com/commaai/opendbc/pull/2420 
> * Needs: https://github.com/commaai/opendbc/pull/2422

This update adds angle control functionality specifically for the Hyundai Ioniq 5 PE model. Changes were made across several files and variables to accommodate this additional feature. This control type provides enhanced handling and steering performance depending on the driving conditions, given its ability to adjust steering angle precisely.

* IONIQ 5 PE EU route with OP running this branch: `e1107f9d04dfb1e2/0000025f--8df0049cd4`
* Route of stock HKG with LFA (at the second half of the route I explicitly disabled LKAS): `e1107f9d04dfb1e2/00000096--02ecca61a6`

> [!NOTE]
>  Because we've seen the torque have different scaling on different vehicles @sunnyhaibin looked into possible signals to detect vehicles and adjust safety limits based on it. To that, we found a signal `DATC` that seems a good candidate. For the record here are the values observed from 2 EU ioniq5 and 1 USA one 

image

### DIFF
--- a/opendbc/car/hyundai/carcontroller.py
+++ b/opendbc/car/hyundai/carcontroller.py
@@ -131,7 +131,7 @@ class CarController(CarControllerBase):
         adaptive_ramp_rate = max(torque_delta / self.angle_torque_override_cycles, 1)  # Ensure at least 1 unit per cycle
         self.lkas_max_torque = max(self.lkas_max_torque - adaptive_ramp_rate, self.params.ANGLE_MIN_TORQUE)
       else:
-        # Hardcoding to a "sensible" value until we can get a better dynamic value with the new controller above.
+        # EU vehicles have been seen to "idle" at 96, while US vehicles have been seen idling at "230" for LFA.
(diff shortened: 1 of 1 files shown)


devtekve:
On stock HKG, I noticed that for my car, the steering angle for LFA caps out at 120º, while for LKAS it caps out at 180º
`e1107f9d04dfb1e2/00000096--02ecca61a6/0`

image

For that test I was with LFA on, but on a traffic light, and I turned the steering all the way to the left and then all the way to the right.

devtekve:
@sshane FYI angle steering via LKAS msg has a hard limit at 180. Anything more than 180 is ignored by the ADAS DRV ECU. There's no fault, just simply capped

PR #1679, opened 2025-01-30
added firmware for my new car

Victor4156:
'23 G70 2.0t sport prestige


Victor4156:
This was an accidental pull request, Please disregard because I haven't tested if the fingerprinting will work.

PR #1681, opened 2025-01-30
IsoTpMessage: support CAN FD single frames

sshane:
Moved from https://github.com/commaai/panda/pull/1524

Will merge by hackathon! Just need some time to make sure this is done correctly w/ reasonable assertions

### DIFF
--- a/opendbc/car/uds.py
+++ b/opendbc/car/uds.py
@@ -469,14 +469,21 @@ class IsoTpMessage:
 
     if rx_data[0] >> 4 == ISOTP_FRAME_TYPE.SINGLE:
       assert self.rx_dat == b"" or self.rx_done, "isotp - rx: single frame with active frame"
-      self.rx_len = rx_data[0] & 0x0F
+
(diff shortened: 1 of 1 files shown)


opendbc/car/uds.py
```diff
@@ -469,15 +469,29 @@ def _isotp_rx_next(self, rx_data: bytes) -> ISOTP_FRAME_TYPE:
 
     if rx_data[0] >> 4 == ISOTP_FRAME_TYPE.SINGLE:
       assert self.rx_dat == b"" or self.rx_done, "isotp - rx: single frame with active frame"
-      self.rx_len = rx_data[0] & 0x0F
-      assert self.rx_len < self.max_len, f"isotp - rx: invalid single frame length: {self.rx_len}"
-      self.rx_dat = rx_data[1:1 + self.rx_len]
+
+      # CAN FD single frame:
```
sshane, line 473:
```suggestion
```

PR #1691, opened 2025-02-04
ISO-TP: set up logging

sshane:
Now you just need to touch one logging object instead of passing down a parameter through a bunch of classes


opendbc/car/carlog.py
```diff
@@ -0,0 +1,12 @@
...
+
+# set up logging
+LOGPRINT = os.environ.get('LOGPRINT', 'INFO').upper()
+carlog = logging.getLogger('carlog')
+carlog.setLevel(LOGPRINT)
+carlog.propagate = False
+
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter('%(message)s'))
+carlog.addHandler(handler)
```
sshane, line 12:
```suggestion
```

not sure why panda has this

PR #1696, opened 2025-02-06
VW MEB: Platform support

jyoung8607:
Support for the [Volkswagen ID family and other MEB vehicles](https://en.wikipedia.org/wiki/Volkswagen_Group_MEB_platform#Production_vehicles). Thanks and credit to @infiniteCable.

- [ ] Harness: Volkswagen C (issue link TBD)
- [ ] Route with stock Travel Assist: `81dd9e9fe256c397/0000002f--bff24fca36`
- [ ] Route with openpilot, stock long: `81dd9e9fe256c397/0000001f--97c42cf98d`
- [ ] Process replay route: `TBD`

**Safety layer TODO**
- [ ] commaai/panda#2160
- [x] #1931
- [x] Verify CAN ignition is actually necessary
- [x] Move to DEBUG until comma releases from dashcam
- [x] Look for opportunities to consolidate code with MQB and MQBevo
  - [x] checksum support
  - [ ] ~rx steering angle, driver input torque~
- [ ] Lateral control safety
  - [ ] ~curvature to lateral accel ISO limits~ (switching to power control exclusively)
  - [ ] driver override power control
- [x] CI tests
  - [x] General (wheel speeds, gas/brake, driver input torque, ACC buttons, ACC state)
  - [ ] ~curvature to lateral accel ISO limits~ (switching to power control exclusively)
  - [ ] driver override power control

**Driving layer TODO**
- [x] Switch initial vehicle support from Cupra Born to Volkswagen ID.4, for testing at comma_hack
- [x] #1816
- [x] #1836
- [x] #1847
- [x] #1849
- [x] Lateral control
  - [x] commaai/openpilot#34679
  - [x] Refactor torque-vs-curvature carcontroller code into separate functions
  - [x] Power-percentage control on driver torque input
  - [x] Investigate power boost possibility
  - [x] Fix LKAS fault logic at startup/standstill
  - [ ] ~Injection testing, probe control limits~
  - [ ] CI tests for ~curvature and~ power safety limits
- [x] Look for opportunities to consolidate code with MQB and MQBevo
  - [x] ~Use DBC generator for common messages~ (too many subtle differences)
  - [x] carstate
  - [x] mqbcan/mebcan
  - [x] checksum support
- [x] Camera integration (in addition to existing gateway integration)
- [x] Fix BSM false firing
- [x] Fix comfort turn signals
- [ ] EA mitigation
- [x] Common gearshift signal

**De-scoped from initial merge TODO:**
- [ ] Explicit opendbc checksum tests (deferred because CAN-FD messages are doing something weird)
- [ ] EA mitigation, current solution works but see if simulating capacitive touch is better
- [x] Alpha openpilot longitudinal control
  - [ ] Test radar UDS knockout from camera
  - [x] Fix cruise fault logic at startup/standstill
  - [ ] Fix nonAdaptive bug
  - [ ] Verify nature/need of the ACC API fault-avoidance carcontroller code
  - [ ] Verify nature/need of special HUD code (temporarily set aside in 42b708bbceb3facd129fd56ed02b31ae51849ddd)
  - [ ] Verify radar code is doing what we need
  - [ ] Panda safety tests (written, just backed out)


jyoung8607:
@sshane When you get a chance, let me know your impression of how the ID.4 drives compared to other comma cars. Feel free to adjust the STEER_POWER constants if you feel the need, I was thinking about lowering the MIN slightly.

I did not clearly understand the way angle control maps to curvature, particularly the different ways that VW would use it as compared to Ford, so I tried my own implementation of curvature and power limits. If you're okay with that method (assuming we add some tests) the way it's done can replicated entirely within Panda.

This method does shortchange you a bit when taking a banked curve; you'll notice that on the comma test loop. However, I don't think the Ford-style angle control would help us here either. It would be nice if there were a way to communicate openpilot's localizer roll down to Panda, that way we could compensate.

The only other problem I'm aware of is that stock ACC sometimes doesn't want to reengage immediately after a disengage. I didn't have time to study why that was, or whether it affects stock. The problem only lasts a couple seconds.

Let me know if you run into anything else!

sshane:
comma hack 5 device: 81dd9e9fe256c397

PR #1698, opened 2025-02-07
Add buttonEnable carState field

sshane:
To remove logic that lives in openpilot now for handling the brands' various implementation of button enabling. Pt. 1 of deleting [car_specific.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/car_specific.py)

### DIFF
--- a/opendbc/car/gm/carstate.py
+++ b/opendbc/car/gm/carstate.py
@@ -35,11 +35,10 @@ class CarState(CarStateBase):
 
   def update_button_enable(self, buttonEvents: list[structs.CarState.ButtonEvent]):
     """ The ECM allows enabling on falling edge of set, but only rising edge of resume """
-    if not self.CP.pcmCruise:
-      for b in buttonEvents:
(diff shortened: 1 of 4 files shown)


opendbc/car/gm/carstate.py
```diff
@@ -33,6 +33,15 @@ def __init__(self, CP):
 
     self.distance_button = 0
 
+  def update_button_enable(self, buttonEvents: list[structs.CarState.ButtonEvent]):
+    """The ECM allows enabling on falling edge of set, but only rising edge of resume"""
+    if not self.CP.pcmCruise:
+      for b in buttonEvents:
+        if (b.type == ButtonType.accelCruise and b.pressed) or \
```
sshane, line 40:
```suggestion
    if not self.CP.pcmCruise:
      for b in buttonEvents:
        The ECM allows enabling on falling edge of set, but only rising edge of resume
        if (b.type == ButtonType.accelCruise and b.pressed) or \
```

PR #1702, opened 2025-02-09
VW: Don't treat the initializing state as a fault

jyoung8607:
In rare configurations, the EPS reverts back to INITIALIZING when auto start-stop has stopped the engine at standstill. It might be related to EPS configuration options that control whether EPS power assist is available with ignition-on but the engine not running.

Route demonstrating the problem: `ba88d39b49c5e6b8/0000000b--afe45956d1`

The other fault states give us the coverage we need.

### DIFF
--- a/opendbc/car/volkswagen/carstate.py
+++ b/opendbc/car/volkswagen/carstate.py
@@ -269,10 +269,10 @@ class CarState(CarStateBase):
     return ret
 
   def update_hca_state(self, hca_status):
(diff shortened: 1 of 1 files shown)


JaCzekanski:
@jyoung8607 I usually drive with Start&Stop off, but on rare occasions while in traffic jam I tried switching it back on, openpilot always greeted me with fault.

Thanks for fixing it.

PR #1703, opened 2025-02-09
Added '23 G70 2.0

Victor4156:
Model: 2023 Genesis G70 Sport Prestige 2.0T
I'm having trouble understanding why some of the firmware versions are so different from the rest of the brand. Also for the extra Ecus I wasn't sure what to do with them.


sshane:
What is your dongle id?

PR #1712, opened 2025-02-12
VW PQ: Add FW for 2012 Jetta Sportwagen TDI

insertwittyusername:
Just wanted to add in my firmware so that it could get upstreamed into master-ci for my car.

### DIFF
--- a/opendbc/car/volkswagen/fingerprints.py
+++ b/opendbc/car/volkswagen/fingerprints.py
@@ -352,9 +352,11 @@ FW_VERSIONS = {
   CAR.VOLKSWAGEN_JETTA_MK6: {
     (Ecu.srs, 0x715, None): [
       b'\xf1\x875C0959655M \xf1\x890726\xf1\x82\t00NB1108--------24',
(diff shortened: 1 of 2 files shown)


sshane:
What's your dongle id?

insertwittyusername:
> What's your dongle id?

My dongle id is a0d99b85ff06857b.

insertwittyusername:
I spoke with JYoung on the Discord, my car has the same chassis code "AJ" as the Mk6 Jetta sedan (my car is the Mk6 Jetta wagon) despite having a different wheelbase, but this setup works regardless, and there isn't resources to try to fix the issue of having two different vehicles with the same chassis code, so I just thought I would add my firmware to the pre existing fingerprint.

insertwittyusername:
Here's a route that shows it working:

a0d99b85ff06857b/00000039--7c828e39aa/11

jyoung8607:
Note for later: The Jetta Sportwagen wheelbase is ~7cm shorter than Jetta because the USA Jetta Sportwagen is really a Golf wagon. Unfortunately Golf and Jetta share the same chassis code so there's not much we can do about it right away, and we're already carrying some worse wheelbase issues (Transporter and Crafter).

PR #1713, opened 2025-02-12
Move car safety modes to opendbc

sshane:
from [panda](https://github.com/commaai/panda/pull/2143)

todo:
- [x] rename libpanda to libsafety
- [ ] car ignition tests, not required
- [x] ensure header files aren't duplicated (ignition header is silently not used)
- [x] misra
- [x] misra mutation tests
- [x] safety mutation tests
- [x] no mac safety testing, we want?
- [ ] install cppcheck once without cache

to remove (in board/):
- [x] config.h
- [x] ~can.h~ - need this, it has code to calculate addr, len, dat from CANPacket_t
- [x] health.h
- [x] ~faults.h~ - TODO: relay malfunction fault seems reasonable for libsafety, figure out how to blend this with panda's faults
- [x] faults_declarations.h~ - same
- [x] libc.h
- [x] boards/*
- [x] main_definitions.h - we currently set `hw_type` to 6 and 7 for the safety tests, but this variable isn't used in board/safety at all
- [x] main_declarations.h
- [x] drivers/*
- [x] comms_definitions.h
- [x] can_comms.h


sshane:
Safety mutation tests should pass once we merge as it uses the git diff. The original PR says this is expected: https://github.com/commaai/panda/pull/2040

PR #1719, opened 2025-02-13
Ford: increase message size for MAC

incognitojam:
```
from openpilot.tools.lib.logreader import LogReader

route = '5a1844a261142c75/00000006--569f26efa4/0'
lr = LogReader(route)

sizes = {}
for msg in lr:
  if msg.which() != "can": continue
  for m in msg.can:
    if m.src >= 64 or m.address > 1400: continue
    sizes[m.address] = len(m.dat)

for address, size in sorted(sizes.items(), key=lambda x: x[0]):
  if 16 <= size < 32:
    print(address, size)
```


sshane:
We want to enforce message lengths soon, this would fault that check.

PR #1722, opened 2025-02-15
Honda: Car Port Acura Integra

vanillagorillaa:
Checklist

- [ ]  added entry to CarInfo in selfdrive/car/*/values.py and ran selfdrive/car/docs.py to generate new docs
- [x]  test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ]  route with openpilot:
- [x]  harness Type: Bosch B

Requires
- [ ] https://github.com/commaai/panda/pull/2152/files


opendbc/car/honda/carstate.py
```diff
@@ -21,7 +21,7 @@
 
 def get_can_messages(CP, gearbox_msg):
   messages = [
-    ("ENGINE_DATA", 100),
+    ("ENGINE_DATA", 0), # TODO : not on every car (missing from Integra)
```
jyoung8607, line 24: We definitely can't disable the frequency check for other cars, this would have to become a conditional.

opendbc/safety/safety/safety_honda.h
```diff
@@ -80,7 +81,8 @@ static void honda_rx_hook(const CANPacket_t *to_push) {
   int bus = GET_BUS(to_push);
 
   // sample speed
-  if (addr == 0x158) {
+  // 0x158 used for all suported Hondas except Integra (use 0x309 car_speed message)
+  if ((addr == 0x158) || (addr == 0x309)){
```
jyoung8607, line 85: The `0x309` case and its corresponding RxCheck aren't exercised in tests. I'm not sure we can depend on it anyway, so it may be irrelevant.

opendbc/car/honda/carstate.py
```diff
@@ -136,10 +136,13 @@ def update(self, can_parsers) -> structs.CarState:
     # ******************* parse out can *******************
     # STANDSTILL->WHEELS_MOVING bit can be noisy around zero, so use XMISSION_SPEED
     # panda checks if the signal is non-zero
-    ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 1e-5
+    if self.CP.carFingerprint == CAR.ACURA_INTEGRA:
+        ret.standstill = cp.vl["CAR_SPEED"]["CAR_SPEED"] < 1e-5
```
jyoung8607, line 140: I'm definitely not in love with this message, it's almost certainly intended for instrument cluster display. Low frequency, high arb ID, and has a considerable response lag.

jyoung8607 (commented):
The under-2kmh speed issue is annoying. I see how you're trying to solve it, and while I'm not in love with your solution, I looked in Cabana myself for a bit and I'm not sure I see a better one.

I'm curious how stock deals with it. What's the control envelope for the factory camera-based ACC? Will it slow to a stop, and if so, will it resume? Does openpilot longitudinal control work?

opendbc/car/honda/carstate.py
vanillagorillaa: Certainly not ideal, but given that "STEER_MOTOR_TORQUE" doesn't have a frequency check, I thought it would be better than the additional lines for a conditional check. I can change it though.

opendbc/car/honda/carstate.py
vanillagorillaa: I can take another look to see what we have, or can use

opendbc/safety/safety/safety_honda.h
vanillagorillaa: will leave for the moment. Maybe we can find another message for speed

opendbc/car/honda/carstate.py
jyoung8607: It's a nice-to-have bit of information on Honda, but isn't used in carcontroller or safety like it is on certain other brands. Our vehicle speed reference is definitely safety critical.

opendbc/car/honda/carstate.py
jyoung8607: In a pinch, there's a message that has ABS wheel speed sensor impulse integral, and that doesn't have the 2 kmh cutoff. We'd have to calibrate it against the main reference speed, but there's precedent and support for briefly suppressing engagement for steering angle sensor calibration and the like.

The reason I asked about stock ACC is that I'd be surprised if it does FtS/SnG if its own reference source has the 2kmh cutoff.

mvl-boston: Does anyone have a route from a CVT (automatic) Integra?

The manual Integras do not do SnG (they disable all stock ACC under 25mph because it is a manual).  And I presume cruisespeed is good enough for manuals as comma won't be engaging on a manual's standstill.

No Honda/Acuras do factory SnG, you have to always startup with a button press (which comma spoofs for their SnG). 

But the CVT Intregra probably will support comma button-press SNG as it does cruise down to zero, and may have xmission speed or an alternative still active if they want reliable cruise to zero inputs.

Yoshiofthewire: My '23 Integra with CVT has what they call "Slow speed follow". It will stop and go for up to 90 seconds, after which you have to press the gas to get a resume.  Using this branch, Stop and Go seems to just work for me, but I have yet to test heavily.

mvl-boston: Can you post a route log?   It will help to see if the CVT has a better signal than the TypeS did for determining standstill.

Yoshiofthewire: [Integra23SnG.zip](https://github.com/user-attachments/files/20244275/Integra23SnG.zip)
17:37:00 the car resumes forward in a stop and go.

Yoshiofthewire: bc9b5ddf81aa9f46/00000024--2beaa6fc4c/0

mvl-boston: There are very high sensitivity curves in 20E.   I can't figure out the exact cutoffs, appears to be a cycling odometer, but it is very sensitive to standstill, perhaps standstill can be written as new20E = prior20E.

mvl-boston:
Reopened as https://github.com/commaai/opendbc/pull/2432

PR #1726, opened 2025-02-16
Custom rx sub_addr

robbederks:
BMW returns a different sub address when talking to the ECU


opendbc/car/uds.py
```diff
@@ -345,9 +346,9 @@ def _recv_buffer(self, drain: bool = False) -> None:
             carlog.debug(f"CAN-RX: {hex(rx_addr)} - 0x{bytes.hex(rx_data)}")
 
             # Cut off sub addr in first byte
```
sshane, line 348:
```suggestion
            # Cut off sub addr in first byte
            sub_addr = self.rx_sub_addr if self.rx_sub_addr is not None else self.sub_addr
```

do something like this?

sshane (approved):

PR #1731, opened 2025-02-17
Add Tesla Model 3/Y ignition signal

sshane:
https://github.com/commaai/panda/pull/2153


opendbc/dbc/tesla_model3_party.dbc
```diff
@@ -230,7 +263,35 @@ BO_ 923 DAS_status: 8 PARTY
```
sshane, line 265:
```suggestion


```

PR #1751, opened 2025-02-19
2023 F-150  5.0 - FIngerprint

blue-genie:
route: e36b272d5679115f/000002fc--34c96aa127


blue-genie:
I'm not sure if anything else is needed to add for the fingerprint PR, but I placed this py on my device and my truck is now fingerprinted - just getting ready for testing this weekend

I also removed Dashcam for my test

PR #1753, opened 2025-02-20
Rivian: add Safety

lukasloetkolben:


sshane:
Can you check the safety tests?

lukasloetkolben:
Yes, I am currently traveling. I will have a look this evening.

opendbc/safety/safety/safety_rivian.h
```diff
@@ -0,0 +1,161 @@
...
+    if(addr == 0x208){
+      float speed =  (((GET_BYTE(to_push, 6) << 8) | GET_BYTE(to_push, 7)) * 0.01);
+      vehicle_moving = speed != 0;
+      UPDATE_VEHICLE_SPEED(speed / 3.6);
+    }
+
+    // Driver torque
+    if (addr == 0x380) {
+      int torque_driver_new = ((((GET_BYTE(to_push, 2) << 4) | (GET_BYTE(to_push, 3) >> 4)) * 0.1) - 205);
+      update_sample(&torque_driver, torque_driver_new);
```
sshane, line 22: `driver_torque_allowance` is an integer, but you're converting this to a float (that's getting truncated to an int with update_sample), are you sure this is doing anything?

sshane: so subtract `205/0.1=2050` here

opendbc/safety/tests/test_rivian.py
```diff
@@ -7,7 +7,7 @@
 from opendbc.safety.tests.common import CANPackerPanda
 
 
-class TestRivianSafety(common.PandaCarSafetyTest, common.DriverTorqueSteeringSafetyTest):
+class TestRivianSafetyBase(common.PandaCarSafetyTest, common.DriverTorqueSteeringSafetyTest, common.LongitudinalAccelSafetyTest):
```
sshane, line 10: Stock long should be the default

sshane:
The Rivian we tested achieved more lateral accel going left vs. right. The max lateral accel with 350 is ~4 m/s^2 without driver torque limiting.

The routes without driver torque safety where cmd torque was allowed to wind up to max, were able to achieve a bit more lateral accel than the routes with driver torque safety. This is likely because of the reduced steering friction/higher momentum, but isn't totally understood. For this PR we are limiting the car the worst case for 3.0 m/s^2 (so 80 mph, left turn, no driver torque safety).

Once we understand why we see more lateral accel without driver torque safety, we can increase these limits.

Injection test routes with driver torque safety:
- bc095dc92e101734/000000d8--2200ab212a

The average lateral accel we saw was 2.35 m/s^2 with a max of 2.7 m/s^2 at 73 mph.

image

Injection test routes without driver torque safety:

- bc095dc92e101734/000000d1--9e1eb61a47  # 350 MAX_STEER
- bc095dc92e101734/000000d3--19b9024311  # 250 MAX_STEER

The maximum lateral acceleration we hit without driver torque limiting at 250 units was ~3.7 m/s^2 with a ~3.1 average

The maximum lateral accel we saw was 3.7 m/s^2 at 80 mph.

PR #1755, opened 2025-02-20
namespace runners for all

sshane:


.github/workflows/tests.yml
```diff
@@ -13,17 +13,15 @@ jobs:
...
+        os: ['namespace-profile-amd64-8x16','namespace-profile-macos-8x14']
     steps:
     - uses: commaai/timeout@v1
     - uses: actions/checkout@v4
     - run: ./test.sh
 
   safety_tests:
     name: safety
-    runs-on: ${{ ((github.repository == 'commaai/opendbc') && ((github.event_name != 'pull_request') || (github.event.pull_request.head.repo.full_name == 'commaai/opendbc'))) && 'namespace-profile-amd64-8x16' || 'ubuntu-latest' }}
+    runs-on: 'namespace-profile-amd64-8x16'
```
adeebshihadeh, line 24: forks don't have namespace (but fork PRs to upstream can)

sshane: got it

PR #1761, opened 2025-02-20
Tesla: remove old fault prevention code

sshane:
Unclear if this does anything on newer cars. FSD is seen at 35+, and during high curvature rate turns, this effectively limits the rate much lower than intended.

https://connect.comma.ai/2c912ca5de3b1ee9/00000061--464f9c85c3

image

See the up and down limits are effectively the same because of this:

image

### DIFF
--- a/opendbc/car/tesla/carcontroller.py
+++ b/opendbc/car/tesla/carcontroller.py
@@ -28,8 +28,8 @@ class CarController(CarControllerBase):
         # Angular rate limit based on speed
         apply_angle = apply_std_steer_angle_limits(actuators.steeringAngleDeg, self.apply_angle_last, CS.out.vEgo, CarControllerParams)
 
-        # To not fault the EPS
-        apply_angle = float(np.clip(apply_angle, CS.out.steeringAngleDeg - 20, CS.out.steeringAngleDeg + 20))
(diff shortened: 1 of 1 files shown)


opendbc/car/tesla/carcontroller.py
```diff
@@ -28,8 +28,8 @@ def update(self, CC, CS, now_nanos):
         # Angular rate limit based on speed
         apply_angle = apply_std_steer_angle_limits(actuators.steeringAngleDeg, self.apply_angle_last, CS.out.vEgo, CarControllerParams)
 
-        # To not fault the EPS
-        apply_angle = float(np.clip(apply_angle, CS.out.steeringAngleDeg - 20, CS.out.steeringAngleDeg + 20))
+        # # To not fault the EPS
+        # apply_angle = float(np.clip(apply_angle, CS.out.steeringAngleDeg - 40, CS.out.steeringAngleDeg + 40))
```
sshane, line 32:
```suggestion
        # FSD has been seen with deltas up to ~40
        apply_angle = float(np.clip(apply_angle, CS.out.steeringAngleDeg - 40, CS.out.steeringAngleDeg + 40))
```

PR #1784, opened 2025-02-22
Rivian: fix resume from standstill

lukasloetkolben:


sshane:
did we test this change tonight or was that master?

lukasloetkolben:
> did we test this change tonight or was that master?

We tested this change tonight

sshane:
can you rebase?

PR #1786, opened 2025-02-22
Tesla: fix longitudinal override jerk

sshane:


maxime-desroches:
trigger-test-models

maxime-desroches:
trigger-test-models

sshane:
does not work yet

PR #1787, opened 2025-02-22
Rivian: update torque params

sshane:
image


sshane:
Max lat accel is still roughly 2.5 m/s^2 (sometimes slightly less):

image

PR #1795, opened 2025-02-22
Run misra test in test.sh

adeebshihadeh:


adeebshihadeh:
Too slow, will investigate later

PR #1801, opened 2025-02-22
Tesla: forward Summon

sshane:
summon and autopark: https://connect.comma.ai/2c912ca5de3b1ee9/000000b1--77e4f8b8bf


sshane:
Unfortunately Autopark uses the normal Tesla angle steering and ACC, so we can't tell if it wants to enable for Autopark or any other reason (like user engaging via stalk)

sshane:
forward and then come to point summon: 2c912ca5de3b1ee9/0000024b--f77e7d4bcc

PR #1810, opened 2025-02-23
Ford: enable Q4 safety in release

adeebshihadeh:


sshane:
done in https://github.com/commaai/opendbc/pull/1896

PR #1823, opened 2025-02-23
Reducing testing time of car/tests to be about 0.3 seconds (link to #1184)

tanyamalik2000:
I was able to reduce the testing time of car/tests to about 0.3 seconds just by removing the parameterized import in all the files and replacing it with for loops. This links to #1184. 

I can check to see if I can reduce it further to less than 0.1 seconds according to the bounty, but this pull request can either be used as a checkpoint (or completion if the reduction is not possible).


sshane:
The tests are failing. Can we instead just write a fast parameterized?

Issue #1794, opened 2025-02-22
Ford Q4 fingerprinting issue - take 2


blue-genie:
Route: e36b272d5679115f/0000030f--340e67d898

More than 6 minutes long - Dashcam with unrecognized car

`git diff` output:

```
comma@comma-e11dd515:/data/openpilot$ git diff
diff --git a/opendbc_repo b/opendbc_repo
--- a/opendbc_repo
+++ b/opendbc_repo
@@ -1 +1 @@
-Subproject commit 4185f81635edab3e943a881edde7e1a03d43ec9c
+Subproject commit 4185f81635edab3e943a881edde7e1a03d43ec9c-dirty
```

```
comma@comma-e11dd515:/data/openpilot/opendbc_repo$ git diff
diff --git a/opendbc/car/ford/interface.py b/opendbc/car/ford/interface.py
index 26a18ea..1b22341 100644
--- a/opendbc/car/ford/interface.py
+++ b/opendbc/car/ford/interface.py
@@ -20,7 +20,7 @@ class CarInterface(CarInterfaceBase):
   @staticmethod
   def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experimental_long, docs) -> structs.CarParams:
     ret.brand = "ford"
-    ret.dashcamOnly = bool(ret.flags & FordFlags.CANFD)
+#    ret.dashcamOnly = bool(ret.flags & FordFlags.CANFD)

     ret.radarUnavailable = Bus.radar not in DBC[candidate]
     ret.steerControlType = structs.CarParams.SteerControlType.angle
```

To fix it, while driving I just rebooted the device. The next route `e36b272d5679115f/00000310--6760d21c98` did a good fingerprinting.

adeebshihadeh:
I don't understand this issue?

sshane:
Try this branch for a few days: https://github.com/commaai/opendbc/tree/ford-queries

The query for the EPS didn't come back the second time, unclear why the panda or gateway is dropping messages silently (we see them on the other buses, but not the one we requested it on).

blue-genie:
route: e36b272d5679115f/00000319--0e9ee82ed2 - using the `ford-queries` branch. 

Rebooted, and route 31a worked fine. 

```
/data/openpilot
++ git diff
diff --git a/opendbc_repo b/opendbc_repo
index 4ffe21a51..0ebf0fbe4 160000
--- a/opendbc_repo
+++ b/opendbc_repo
[...]

PR #1824, opened 2025-02-23
Ford: add F-150 Lightning 2022 FW

incognitojam:
Note: route is missing EPS fw...

Route: `0b2c0bec9a28eb0f/00000000--d6637dad8d`
Second route: `0b2c0bec9a28eb0f/00000001--c766afec05` (eps still missing)


incognitojam:
seeing same issue as #1794 in these routes

incognitojam:
we could just merge this as-is?

PR #1828, opened 2025-02-24
Hyundai CAN-FD Safety: Refactor TX message assignments

sunnyhaibin:
**Description**
- Refactored Hyundai CAN-FD TX message assignments to use reusable and descriptive macros.
- Consolidated similar TX message configurations into grouped definitions for better clarity.
- Reduced code redundancy by replacing inline message lists with predefined macros.
- No functional changes were introduced, focusing solely on code organization.


sunnyhaibin:
Validation: https://github.com/commaai/openpilot/actions/runs/13494287208

PR #1830, opened 2025-02-24
Hyundai CAN-FD Safety: Refactor bus detections

sunnyhaibin:


sunnyhaibin:
Validation: https://github.com/commaai/openpilot/actions/runs/13495629170

sunnyhaibin:
Validation: https://github.com/commaai/openpilot/actions/runs/13495813588

PR #1835, opened 2025-02-24
Ford CANFD Radar Parser

blue-genie:
Radar enabled with the same logic as for CAN - see below for more details: 

> e36b272d5679115f/00000308--b61d26edc1/0

Radar Enabled - with final version of the code:

> e36b272d5679115f/0000033b--8d4b7c8da5/0
> e36b272d5679115f/0000033c--d7640794bc/0

I started with using the same logic used for CAN Fords, adjusting the code for the missing signals in CANFD, and I had some artifacts like:
https://github.com/user-attachments/assets/f4537216-9a6d-43a3-b973-a7bca28b8ff9

Check the lead chevron how it bounces.
https://github.com/user-attachments/assets/dfd9ff2a-e680-4d3f-afd5-2f34bd463ee9
I'm expecting the old radar points to remain in the right and new ones to come on the left.

The road curves to the right and the radar points cross the lanes, when there is noting on the lanes.
image

CAN logic uses `yRel = -sin(azimuth) * dist` and I used the same for CANFD and I got the above artifacts, then I switched to `yRel = sin(azimuth) * dist` and I got:

https://github.com/user-attachments/assets/fdac539c-1d03-4db0-b1bd-b9a85ed4e4a9

Note how nice old points are left on the right and new ones come from the left.
https://github.com/user-attachments/assets/d2120681-f914-4c7c-b338-2d0f11cf1e35

Note how nice the radar points match with the left edge of the road.
image

I will add some more comments in the code.


opendbc/car/ford/interface.py
```diff
@@ -34,6 +34,10 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experime
       # MRR_Header_Timestamps->CAN_DET_TIME_SINCE_MEAS reports 61.3 ms
       ret.radarDelay = 0.06
 
+    if not ret.radarUnavailable and DBC[candidate][Bus.radar] == RADAR.DELPHI_MRR_64:
```
blue-genie, line 37: Didn't really understood how the 60ms was calculated, but I tried to apply a similar logic for CANFD

opendbc/car/ford/radar_interface.py
```diff
@@ -217,6 +234,67 @@ def _update_delphi_mrr(self):
...
+        dist = msg[f"CAN_DET_RANGE_{ii:02d}_{iii:02d}"]  # m [0|255.984]
+        if scanIndex in (1, 3) and dist < DELPHI_MRR_MIN_LONG_RANGE_DIST:
+          valid = False
+
+        if valid:
+          print(f"CAN_DET_HOST_VEH_CLUTTER: {msg[f"CAN_DET_HOST_VEH_CLUTTER_{ii:02d}_{iii:02d}"] }") if msg[f"CAN_DET_HOST_VEH_CLUTTER_{ii:02d}_{iii:02d}"] > 0 else None
+          azimuth = msg[f"CAN_DET_AZIMUTH_{ii:02d}_{iii:02d}"]              # rad [-3.1416|3.13964]
+          distRate = msg[f"CAN_DET_RANGE_RATE_{ii:02d}_{iii:02d}"]          # m/s [-128|127.984]
+          dRel = cos(azimuth) * dist                                        # m from front of car
+          yRel = sin(azimuth) * dist                                        # in car frame's y axis, right is positive
```
blue-genie, line 286: This is the major change between CAN and CANFD, I can't say if it's correct not not the one for CAN, but this one makes the most sense for CANFD. See the PR details for more explanation

coffee-cake-isaac:
🚀

opendbc/car/ford/radar_interface.py
```diff
@@ -229,6 +245,79 @@ def _update_delphi_mrr(self, ret: structs.RadarData):
...
+
+          self.points.append([dRel, yRel * 2, distRate * 2])
+
+    # Update once we've cycled through all 4 scan modes
+    if headerScanIndex != 3:
+      return True, [] # MRR_Detection_* messages in CANFD are at 20Hz, services.py expects liveTracks to be at 20Hz - we'll send messages to meet the 20Hz
```
blue-genie, line 315: CANFD MRR is at 20MHz
`liveTracks` is at 20Hz in `services.py` - https://github.com/commaai/openpilot/blob/master/cereal/services.py#L33

Sending for 2 scans out of 4 does not mark `liveTracks` as invalid, but sporadically some `liveTracks` don't make it to `radard` which causes `radard` alerts - since it's marked as invalid as well.

adeebshihadeh:
Gonna merge, then we can do final validation once we use it for OP long. For now, it'll only affect FCW.

opendbc/car/ford/interface.py
```diff
@@ -38,10 +37,16 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
...
+    ret.radarUnavailable = Bus.radar not in DBC[candidate]
+    ret.radarDelay = {
       # average of 33.3 Hz radar timestep / 4 scan modes = 60 ms
       # MRR_Header_Timestamps->CAN_DET_TIME_SINCE_MEAS reports 61.3 ms
-      ret.radarDelay = 0.06
+      RADAR.DELPHI_MRR: 0.06,
+
+      # average of 20 Hz radar timestep / 4 scan modes = 100 ms
+      RADAR.DELPHI_MRR_64:  0.1
+    }.get(Bus.radar, 0.1)
```
sshane, line 49: just use an if statement, dict.get never looks clean

sshane: you're comparing an auto enum against a set of unrelated strings

opendbc/car/ford/radar_interface.py
sshane: Is this even needed for CAN FD? This was only for clustering since the track IDs didn't match positionally that well, and radard requires consistent tracks in time

blue-genie: So, based on what you said, I should track how a track changes over time. I should check how its position shifts from one frame to the next, how smooth the graph of the track is. Is this right?

My understanding was that you didn't want to over load the system with multiple clustering.

I'll graph the tracks and see how they change over time.

opendbc/car/ford/interface.py
blue-genie: I'll bring back the `if` statements in a clean way.

opendbc/car/ford/radar_interface.py
sshane: No, the clustering here isn't very clean and should ideally go into radard. If we can avoid another platform that uses it in the meantime, that's better.

sshane:
> I'll collect a new route with clustering on `headerScanIndex` 2 and 3, not only on 3 and see how everything looks.

Make sure you understand what each is

sshane:
Can you provide a route? The two in the description don't have rlogs

blue-genie:
> Can you provide a route? The two in the description don't have rlogs

I'm guessing the old routes were archived.

These are routes on a clean master:
e36b272d5679115f/00000478--81ffd3e440
e36b272d5679115f/00000477--46bd01ebcd
e36b272d5679115f/00000476--442fdcaba9
e36b272d5679115f/00000369--a3e8499a85

All 4 routes have empty GitDiff. See if they help.

Issue #1838, opened 2025-02-24
[$100 bounty] safety: tx hook should be filtered by address


sshane:
We already do this for the rx hook, should do it for the tx hook

$100 bounty for a clean, mergable, reasonable, no huge review changes requested PR

ccdunder:
A related behavior that seems strange to me is that tx hooks currently run if there's a relay_malfunction. We don't tx, fwd, or run fwd hooks, though. Seems like a bug unless I'm missing something?

PR #1851, opened 2025-02-26
Only run tx hooks on msgs in safety_config.tx_msgs.

ccdunder:
https://github.com/commaai/opendbc/issues/1838

renamed whitelisted to allowed for consistency with safety_rx_hook code.


sshane:
safety tests are failing

sshane:
Are our tests wrong?

ccdunder:
@sshane nooutput tx hook will never get exercised after this change. i don't think modifying prod code to fix that missing coverage is the right approach, but i don't spot a convention for excluding lines from coverage — **is there one? what do you generally do for cases like this?** alternatively i could concoct a "test" to exercise that line but that seems convoluted.

ccdunder:
running this locally passes. ugh.. i guess i'll go see what the difference between local & the github runners is

ccdunder:
also this seems like something that should _really_ have a test but there doesn't seem to be a simple way to do right now with adding new test infra. i'm assuming (& hoping) you want that in the same PR, but just calling out that **it will make it larger.**

ccdunder:
Saving the best for last: Just want to call out that it's certainly possible that some car models rely on the existing bad behavior. **What's your rollout policy for potentially breaking changes?**

1. are harness failures acceptable in master? (anticipating they get reported before release)
2. do you have a way to run this over all recent car segments to check for a regression?
3. otherwise, it could be done as a dark launch. my hunch is this is overkill, but really can't say without knowing your policy.

ccdunder:
> also this seems like something that should _really_ have a test but there doesn't seem to be a simple way to do right now with adding new test infra. i'm assuming (& hoping) you want that in the same PR, but just calling out that **it will make it larger.**

Just to be clear, I believe making it larger is the right thing to do based on my principles as follow. Feel free to push back if you want something else:

i'm **not** backfilling existing missing tests (in hyundai_canfd) (since they're already missing so this strictly improves the codebase regardless). i **am** adding a test for the behavior this modifies (since not doing so strictly worsens the codebase by making bugs like this possible).

ccdunder:
TIL the coverage report uses a different tool (lcov) than the coverage test (gcov). gcov doesn't support exclusion markers easily. gcovr does.

then i remembered that i misra disallows unreachable code, so i'll step back and evaluate if another approach makes more sense.

opendbc/safety/safety/safety_hyundai_canfd.h
```diff
@@ -198,9 +198,12 @@ static bool hyundai_canfd_tx_hook(const CANPacket_t *to_send) {
       violation |= longitudinal_accel_checks(desired_accel_val, HYUNDAI_LONG_LIMITS);
     } else {
       // only used to cancel on here
+      // GCOV_EXCL_START
+      // TODO: Cover this with tests.
```
sshane, line 202: Are you sure our tests isn't accidentally sending this on the wrong bus? We should definitely hit this

opendbc/safety/safety/safety_rivian.h
```diff
@@ -80,12 +80,8 @@ static bool rivian_tx_hook(const CANPacket_t *to_send) {
...
     // Longitudinal control
     if (addr == 0x160) {
-      if (rivian_longitudinal) {
-        int raw_accel = ((GET_BYTE(to_send, 2) << 3) | (GET_BYTE(to_send, 3) >> 5)) - 1024U;
-        if (longitudinal_accel_checks(raw_accel, RIVIAN_LONG_LIMITS)) {
-          tx = false;
-        }
-      } else {
+      int raw_accel = ((GET_BYTE(to_send, 2) << 3) | (GET_BYTE(to_send, 3) >> 5)) - 1024U;
+      if (longitudinal_accel_checks(raw_accel, RIVIAN_LONG_LIMITS)) {
```
sshane, line 84: Is this because 0x160 is now blocked since it's not in the tx messages?

ccdunder: correct. the case where addr == 0x160 && !rivian_longitudinal becomes unreachable with this change.

opendbc/safety/safety/safety_hyundai_canfd.h
ccdunder: oh yeah that's right, rings a bell. the first time i noticed this tx_hooks behavior was after refactoring hyundai_canfd tests and being puzzled that one of the buses was backwards (0<->2). great observation! i'll look into it.

ccdunder: Turns out the only test cases that hit this path before this change are `test_spam_can_buses` and `test_relay_malfunction`.

To me that means adding coverage for this is out-of-scope for this PR. We should absolutely follow up on fixing it, though. (That way we can ship something that strictly improves the codebase incrementally and avoid scope creep / trying to boil the ocean.)

ccdunder:
Am I missing something or is there no support for unit-testing safety.h currently? e.g. testing this specific behavior (tx hooks run after the filter)

Best solution to that I see is to make a mock safety_hooks, but I don't think we want it in the const safety_hook_registry. That can be solved by splitting set_safety_hooks apart. But lack of existing test support (and tests) seems like an issue & PR on its own (if you agree it's something worth having). @sshane let me know what you think

adeebshihadeh:
> Just want to call out that it's certainly possible that some car models rely on the existing bad behavior. **What's your rollout policy for potentially breaking changes?**

We monitor master quite closely, but we don't merge any potential regressions that are fully testable offline. In this case, this should be fully verified by CI, and we should add coverage if it isn't. openpilot's `test_models` should cover this case.

opendbc/safety/safety/safety_hyundai_canfd.h
sshane: It seems that our Hyundai tests should test this path. Maybe the tx message is missing from the list for the stock long param

sshane: this should fix this coverage issue: https://github.com/commaai/opendbc/pull/1898

sshane:
thanks!

sshane:
you can open a ticket here for the bounty: https://commasupport.zendesk.com/hc/en-us/requests/new?ticket_form_id=29233607274775

PR #1852, opened 2025-02-26
Ford Q4: limit max curvature from vehicle speed

sshane:
For now we will allow ~2.4 m/s^2 max lateral accel due to not knowing the roll and assuming a ~3.4 degree road roll at all time.


sshane:
Injection test route: https://connect.comma.ai/8279cfddf735a11e/0000000c--ac5563bb50

sshane:
The single safety mutation failure is expected:

```
[info] Survived mutants (2/14):
/home/runner/work/opendbc/opendbc/opendbc/safety/safety.h:755:30: warning: Survived: Replaced >= with > [cxx_ge_to_gt]
      if (desired_angle_last >= 0) {
                             ^
```

1. If desired_angle_last is 0, then we will hit the rate limits before the max curvature limit up to at least 268 mph

PR #1855, opened 2025-02-26
safety: tx hook filtered by address

BrainLessPea:
sovles https://github.com/commaai/opendbc/issues/1838


sshane:
safety and misra tests are failing

sshane:
This removes two critical safety functions

PR #1857, opened 2025-02-26
Rivian: rename and cleanup DBC

lukasloetkolben:


sshane:
Does the name more closely represent which bus it refers to?

lukasloetkolben:
> Does the name more closely represent which bus it refers to?

Yes we are connected to the primary actuator bus. Since there are so many busses in the rivian, I think it makes sense to name them properly.

PR #1862, opened 2025-02-27
Tesla: reduce jerkiness after gas overwrite

lukasloetkolben:
Ramping up the jerk after overriding the gas helps to make it feel less jerky.

### DIFF
--- a/opendbc/car/tesla/teslacan.py
+++ b/opendbc/car/tesla/teslacan.py
@@ -5,6 +5,7 @@ from opendbc.car.tesla.values import CANBUS, CarControllerParams
 class TeslaCAN:
   def __init__(self, packer):
     self.packer = packer
+    self.jerk = 0
 
   @staticmethod
   def checksum(msg_id, dat):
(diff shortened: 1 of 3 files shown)


opendbc/car/tesla/teslacan.py
```diff
@@ -24,14 +25,16 @@ def create_steering_control(self, angle, enabled, counter):
     values["DAS_steeringControlChecksum"] = self.checksum(0x488, data[:3])
     return self.packer.make_can_msg("DAS_steeringControl", CANBUS.party, values)
 
-  def create_longitudinal_command(self, acc_state, accel, cntr, active):
+  def create_longitudinal_command(self, acc_state, accel, cntr, active, override):
+    # ramp up jerk after a gas overwrite
+    self.jerk = 0 if override else self.jerk + CarControllerParams.JERK_RATE_UP
```
sshane, line 30: I suggested this, but @robbederks thought it was hacky. I'm going to take another look again in a few days to see if we're doing anything wrong first before merging

lukasloetkolben: I did a few tests and it feels much smoother.
Maybe you can do a test drive.
Maybe increase the rate to 0.04-0.05

lukasloetkolben: I agree, but for now it's probably better to mask the bug than show it?

sshane:
Are you sure this works? I set both jerks to 0 and it still happens

opendbc/car/tesla/teslacan.py
sshane: Looks like there's a combination of things the stock system does that we don't. On gas override it reduces the jerk and ramps it back up, then the up accel limit is near the ISO max, but because the set speed is negative it doesn't actually actuate it, just allow it. I think since we set to 0 on inactive, it thinks we're violating the accel limit and with a very high jerk it brakes to go within it (0 m/s^2):

image

sshane: Well look at this, you can roughly calculate what the setSpeed should be from vehicle speed :eyes:, it matches pretty closely! Except the estimate was wrong when I put in the down jerk:

image

image

sshane: Oh I was overriding during those accelerations so that may be the reason

sshane: It looks much less similar on an FSD route :disappointed:

lukasloetkolben:
> Are you sure this works? I set both jerks to 0 and it still happens

It's definitely working for me. I have to test setting jerk to 0.

adeebshihadeh:
Still relevant?

lukasloetkolben:
> Still relevant?

The PR fixed my jerky gas override, but apparently wasn't working on the Model Y HW4.
i closed the PR for now

PR #1867, opened 2025-02-27
Replace gcov & lcov with gcovr

ccdunder:
This is in order to support exclusion markers e.g. #1851.

This also replaces lcov, since it's one less dependency, is smaller, and can be installed via uv (or apt).

Compared to gcov, gcovr:
➖ txt reports take 180ms instead of 8ms, so test.sh takes 2% longer to run on my machine.

Compared to lcov, gcovr:
➕ html reports take 300ms instead of 800ms, so test.sh is 6% faster.
➕ generated report size is 2.8 MB vs 4.2 MB
➕ transitive install size is 2.15 MB vs 32.07 MB
➕ report is faster to navigate, contains more info (e.g. branch info)
gcovr sample: [link](https://ccdunder.github.io/gcovr-comparison/gcovr/)
lcov sample: [link](https://ccdunder.github.io/gcovr-comparison/lcov/)

### DIFF
--- a/opendbc/safety/tests/test.sh
+++ b/opendbc/safety/tests/test.sh
@@ -19,17 +19,16 @@ pytest -n8
 
 # generate and open report
 if [ "$1" == "--report" ]; then
-  geninfo ./libsafety/ -o coverage.info
-  genhtml coverage.info -o coverage-out
+  mkdir -p tests/coverage-out
+  gcovr -r . --html-details tests/coverage-out/index.html
   sensible-browser coverage-out/index.html
 fi
 
 # test coverage
-GCOV_OUTPUT=$(gcov -n ./libsafety/safety.c)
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Nice! No drawbacks to the switch?

ccdunder:
Only drawback is test.sh takes 2% longer to run. I added a detailed comparison to the top comment.

sshane:
this breaks the report
image

ccdunder:
> this breaks the report

Where is the screenshot from? gcovr produces output in the same folder as before (`opendbc/safety/tests/coverage-out`) containing an `index.html` same as lcov.

ccdunder:
Ugh no i see what it is. One extra level of path. Will send a fix.

ccdunder:
Fix: https://github.com/commaai/opendbc/pull/1897

PR #1895, opened 2025-03-03
safety replay: warm up

sshane:
From https://github.com/commaai/opendbc/pull/1852

should test_models do this instead?


sshane:
Can do this better in https://github.com/commaai/openpilot/pull/34720

PR #1897, opened 2025-03-03
Fix output path of coverage report.

ccdunder:
Fixes https://github.com/commaai/opendbc/pull/1867#issuecomment-2693450867


sshane:
added back the nesting here: https://github.com/commaai/opendbc/pull/1900

Issue #1839, opened 2025-02-24
[Bounty] safety counter and checksum tests should be true by default


sshane:
$100 bounty for a clean, mergable, reasonable, no huge review changes requested PR

PR #1902, opened 2025-03-03
test(safety): counter and checksum tests true by default 

brymut:
The following PR is attempting to address https://github.com/commaai/opendbc/issues/1839

- Refactored checksum validation to gracefully handle missing checksum functions
- Refactored counter validation to ensure no segfault if get_counter is NULL but max_counter > 0
- Added a new test case to verify that the safety check doesn't crash when get_counter is NULL.

Hi @sshane could you please have a look?


sshane:
Safety tests are failing, can you check why?

PR #1905, opened 2025-03-03
docs: cleanup extra cars

sshane:


sshane:
@jyoung8607 I don't have time for this now, but this is a good proof of concept. Want to take it over? Does this look reasonable to you?

You just need to ignore != SupportType.UPSTREAM on the site jinja template

opendbc/car/docs.py
```diff
@@ -11,20 +10,14 @@
...
 EXTRA_CARS_MD_OUT = os.path.join(BASEDIR, "../", "../", "docs", "CARS.md")
 EXTRA_CARS_MD_TEMPLATE = os.path.join(BASEDIR, "CARS_template.md")
 
-ExtraPlatform = Platform | EXTRA
-EXTRA_BRANDS = get_args(ExtraPlatform)
-EXTRA_PLATFORMS: dict[str, ExtraPlatform] = {str(platform): platform for brand in EXTRA_BRANDS for platform in brand}
-
 
 def get_params_for_docs(platform) -> CarParams:
   cp_platform = platform if platform in interfaces else MOCK.MOCK
```
sshane, line 23: we can remove this once done

jyoung8607:
> @jyoung8607 I don't have time for this now, but this is a good proof of concept. Want to take it over? Does this look reasonable to you?
> 
> You just need to ignore != SupportType.UPSTREAM on the site jinja template

It's more than the CARS.md jinja template, it's the functions consumed by comma.ai/vehicles and shop.comma.ai. At the time I wasn't free to alter that interface at all. Hence the EXTRA_STUFF, so those original functions could continue filtering down to the set of release-supported cars.

https://github.com/commaai/opendbc/blob/01c1a42e99b766747d48b14171a4dd5ae897ff4c/opendbc/car/docs.py#L65-L69

https://github.com/commaai/opendbc/blob/01c1a42e99b766747d48b14171a4dd5ae897ff4c/opendbc/car/docs.py#L84-L93

PR #1908, opened 2025-03-03
Ford CAN FD: SecOC is dashcam

sshane:


blue-genie:
So you think we can have something extra, like maybe a log message or anything to easily identify why the car is still in dash-cam ?

sshane:
these will get added to the qlog

PR #1910, opened 2025-03-04
Rivian: cancel command

sshane:
This PR works by sending an "unavailable" ACC status to the ACM (ADAS ECU) from the VDM (receiver of ACC commands, what we think of the PCM on other cars).

If you keep the ACC status low ("unavailable") the ACM will fault as it thinks something is wrong never seeing the VDM respond to its enable request. However, if you send a few frames of "available" and then cause a falling edge back to "unavailable" the ACM respects this and ACC cancels as if its a user disable (no dash message, just normal disengage tone).

Thanks to Rivian user @gerrylum for testing!


sshane:
So far sending at 100Hz faulted the ACM, and 4Hz was ignored. We may need to forward the ACC message through openpilot

PR #1912, opened 2025-03-04
Rivian: stock AEB safety

sshane:
Might be needed if https://github.com/commaai/opendbc/pull/1910 works


adeebshihadeh:
Still relevant?

PR #1913, opened 2025-03-04
Rivian: add another ACC fault signal

sshane:
https://connect.comma.ai/2d2da5dea63b5c5e/00000002--e4de380ab0

Found while testing https://github.com/commaai/opendbc/pull/1910

image


sshane:
Need to either check gear or propulsion status, it goes high in these cases:

image

sshane:
While this could be useful to log, there's no signal to describe when it's valid, sometimes it stays high after going into D or propulsion status is active, and rises before going into park:

image

PR #1914, opened 2025-03-04
HKG: Add FW for 2020 Hyundai Ioniq HEV

royjr:
Hyundai Ioniq Hybrid 2020 (non plug-in)
Hyundai H
f72173595de5d8a7/0000006c--5a83cac5de


jyoung8607:
LGTM, mostly, and `pytest selfdrive/car` passes locally. @royjr last thing before merging, can you confirm what harness they used? Would be nice to get rid of this TODO.

https://github.com/commaai/opendbc/blob/01807965274aed3564d578230909cfabcaa62ed9/opendbc/car/hyundai/values.py#L224

royjr:
Hyundai H harness / 2020 confirmed, could we assume 2021 and remove todo completely?

PR #1918, opened 2025-03-04
Rivian: add VDM fault signal

sshane:
Found when testing https://github.com/commaai/opendbc/pull/1910. A value of 3 can cancel stock ACC (ACM/Autonomous eXperience Module), but likely permanently faults it after some time:

2d2da5dea63b5c5e/00000004--c739b2e2a0

image


sshane:
Only seen at 1 in all of our local Rivian segments (1000 segments)

PR #1922, opened 2025-03-05
add Ford Escape 2023-24 and Kuga 2024

incognitojam:
- Ford Escape received a facelift in 2023 ("Mk 4.5"), is now on the CAN FD platform
- The Kuga only received the facelift from 2024, and now there isn't a supported ICE variant (ACC/LCA only on auto transmission, which are only HEV/PHEV now)
- No SecOC on a 2024 Escape

Theoretically the 2025 MY is the same? but maybe we should wait to see one

Harness: `Ford Q4`
prev route `56574443e0c3783c|2024-04-12--22-10-02` (on a fork, will get another route running this branch)
Test route: `56574443e0c3783c/00000002--c00bd0fe69`

EDIT: Just seen that there was already a PR, https://github.com/commaai/opendbc/pull/1094


opendbc/car/ford/values.py
```diff
@@ -124,7 +124,14 @@ class CAR(Platforms):
...
     [
       FordCarDocs("Ford Escape 2020-22", hybrid=True, plug_in_hybrid=True),
-      FordCarDocs("Ford Kuga 2020-22", "Adaptive Cruise Control with Lane Centering", hybrid=True, plug_in_hybrid=True),
+      FordCarDocs("Ford Kuga 2020-23", "Adaptive Cruise Control with Lane Centering", hybrid=True, plug_in_hybrid=True),
+    ],
+    CarSpecs(mass=1750, wheelbase=2.71, steerRatio=16.7),
+  )
+  FORD_ESCAPE_MK4_5 = FordCANFDPlatformConfig(
+    [
+      FordCarDocs("Ford Escape 2023-24", hybrid=True, plug_in_hybrid=True),
```
incognitojam, line 133: There are two packages, `Co-Pilot360 Assist+ (68B)` and `Co-Pilot360 Assist 2.0 (68C)`, which both have Adaptive Cruise and Lane Centering.
2023 - https://www.mavericktruckclub.com/forum/attachments/2023-escape-order-guide-pdf.116038/
2024 - https://dealerimages.dealereprocess.com/image/upload/v1700240087/2024_Escape_obwtvs.pdf

The cars come with `Co-Pilot360 features` as standard but they only include Lane Keeping and AEB.

opendbc/car/ford/values.py
```diff
@@ -124,7 +124,14 @@ class CAR(Platforms):
...
+      FordCarDocs("Ford Kuga 2024", "Adaptive Cruise Control with Lane Centering Assist", hybrid=True, plug_in_hybrid=True),
```
incognitojam, line 134: The new Kuga in the UK says the following on higher trims:
> Adaptive Cruise Control (with Stop & Go with self-charging and plug-in hybrid powertrains)
> Lane-Keeping Aid with Lane-Departure Warning. Lane Centering Assist included on all self-charging and plug-in hybrid powertrain
They make a distinction between hybrid/non-hybrid, presumably the non-hybrids are only available with manual gearboxes and ACC with Stop and Go and Lane Centering is no available for these

incognitojam: TODO: check whether all HEV/PHEV have LCA, or if you can get them with lower trim

incognitojam: I think all HEV/PHEV have the required packages

sshane: By default make sure to use the US advertising/packages, but Kuga is not in the US?

incognitojam: the Kuga isn't in the US, and I'm just using US sources for the Escape

incognitojam: oh, none of this applies to the Escape, you can still get hybrids without ACC

opendbc/car/ford/values.py
incognitojam: I think it's fine to leave this as `Co-Pilot360 Assist+` like the Q3 Escape

opendbc/car/tests/routes.py
```diff
@@ -19,6 +19,7 @@
 # FIXME: add routes for these cars
 non_tested_cars = [
   MOCK.MOCK,
+  FORD.FORD_ESCAPE_MK4_5,
```
adeebshihadeh, line 22: needs a test route before we can merge

opendbc/car/tests/routes.py
```diff
@@ -19,6 +19,7 @@
...
   FORD.FORD_F_150_MK14,
```
adeebshihadeh, line 23: @sshane we should fix this missing route too

adeebshihadeh (changes requested):

incognitojam:
Need to check harness parts :mag:

incognitojam:
Oh, comma already sell an extension for the comma power so that can be added to the harness parts. https://comma.ai/shop/comma-power-extender

PR #1792, opened 2025-02-22
Tesla: copy stock longitudinal control

sshane:
image

image

PR #1923, opened 2025-03-05
Tesla: setSpeed is vEgo while disabled

sshane:
simple correctness fix from https://github.com/commaai/opendbc/pull/1792


lukasloetkolben:
With this change it's not possible to engage at low speeds.

https://github.com/commaai/opendbc/pull/1366

sshane:
This is not my experience in our Model Y. Before and after this change it denied enabling under seemingly random conditions, e.g. try to enable at 2 mph, doesn't work, try again in a few seconds, works. Do you have a route of before it allowing you to enable and now not?

PR #1925, opened 2025-03-05
docs: start extra cars clean up

sshane:
https://github.com/commaai/website/pull/32


sshane:
@jyoung8607 hopefully this makes it more clear

PR #1930, opened 2025-03-06
Replace @parameterized.expand with @pytest.mark.parametrize to improve test speed

Aniurm:
This change reduces the `car/tests/` collection time to ~0.4s.

```
~/opendbc$ pytest --cache-clear --collect-only opendbc/car/tests/

================================ 1735 tests collected in 0.42s ===========================
```

This links to https://github.com/commaai/opendbc/issues/1184.

PR #1932, opened 2025-03-06
perf: refactor to lazy interface loading

roelofvandijk:
Relates to https://github.com/commaai/opendbc/issues/1184

By moving the `car_helpers.py` imports to a `pytest.fixture`, we can shorten the collection time.

Refactors the `car` tests to use:
* a pytest `fixture` for the interface loading (this import is not executed during collection)
* ~`pytest.mark.parametrize` instead of the `parametrize` in select places to enable properly using the fixture.~ coincidentally already fixed in https://github.com/commaai/opendbc/pull/1930

I tried to keep the diff as small as possible.

There are 3 big chunks of time spent during the pytest collection on a fresh run (after compilation):
* first import of `CANDefine`, aka the packer_pyx compiled library (only slow the first run after compilation)
* import of the interfaces in `car_helpers.py` (always slow)
* import of pytest plugins

"Fresh" denotes a recompilation of the `cython` modules.

## Collect only 

`pytest opendbc/car/tests --collect-only` 

| State  | Fresh           | Second Run         |
|--------|-----------------|-------------------|
| master  | 0.56s  | 0.27s  |
| PR | 0.16s  | 0.13s  |

compare to

`pytest opendbc/can/tests --collect-only` 

| State  | Fresh           | Second Run         |
|--------|-----------------|-------------------|
| master  | 0.32s | 0.02s  |
| PR | 0.29s  | 0.04s  |

## Collection + execution

`pytest opendbc/car/tests`

| State  | Fresh           | Second Run         |
|--------|-----------------|-------------------|
| master  | 6.42s   | 6.19s |
| PR | 5.92s | 5.76s |

### DIFF
--- a/opendbc/car/tests/conftest.py
+++ b/opendbc/car/tests/conftest.py
@@ -1,0 +1,9 @@
+from opendbc.car.car_helpers import _get_interface_names, load_interfaces
+import pytest
+
+
+# imports from directory opendbc/car/<name>/
+@pytest.fixture(scope="module")
+def interfaces():
+    interface_names = _get_interface_names()
+    return load_interfaces(interface_names)
--- a/opendbc/car/tests/test_car_interfaces.py
+++ b/opendbc/car/tests/test_car_interfaces.py
@@ -7,7 +7,6 @@ from collections.abc import Callable
 from typing import Any
 
 from opendbc.car import DT_CTRL, CanData, gen_empty_fingerprint, structs
-from opendbc.car.car_helpers import interfaces
 from opendbc.car.fingerprints import FW_VERSIONS
 from opendbc.car.fw_versions import FW_QUERY_CONFIGS
 from opendbc.car.interfaces import get_interface_attr
@@ -53,7 +52,7 @@ class TestCarInterfaces:
   @settings(max_examples=MAX_EXAMPLES, deadline=None,
             phases=(Phase.reuse, Phase.generate, Phase.shrink))
   @given(data=st.data())
-  def test_car_interfaces(self, car_name, data):
+  def test_car_interfaces(self, car_name, data, interfaces):
     CarInterface = interfaces[car_name]
 
     args = get_fuzzy_car_interface_args(data.draw)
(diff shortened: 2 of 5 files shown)


opendbc/car/tests/test_car_interfaces.py
```diff
@@ -49,11 +48,11 @@ def get_fuzzy_car_interface_args(draw: DrawType) -> dict:
 class TestCarInterfaces:
   # FIXME: Due to the lists used in carParams, Phase.target is very slow and will cause
   #  many generated examples to overrun when max_examples > ~20, don't use it
-  @pytest.mark.parametrize("car_name", sorted(PLATFORMS))
+  @pytest.mark.parametrize("car_name", sorted(PLATFORMS) + [MOCK.MOCK])
```
roelofvandijk, line 51: I believe the was mistakenly removed in https://github.com/commaai/opendbc/pull/1930

sshane: It should exist in PLATFORMS already

opendbc/car/tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+
+from opendbc.car.car_helpers import _get_interface_names, load_interfaces
+import pytest
+
```
sshane, line 4:
```suggestion


```

opendbc/car/tests/conftest.py
```diff
@@ -0,0 +1,9 @@
...
```
sshane, line 1:
```suggestion
```

opendbc/car/tests/test_lateral_limits.py
```diff
@@ -19,53 +18,54 @@
 # jerk is measured over half a second
 JERK_MEAS_T = 0.5
 
+@pytest.fixture
```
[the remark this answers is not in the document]
roelofvandijk: If this is to make the review easier, I reduced the diff.

It seems nonsensical, because we are not parametrizing anymore over the test class, but directly over the test functions.

roelofvandijk:
@sshane rebased, minimized diff.

PR #1939, opened 2025-03-07
safety: checksum and counter are explicitly opt-out

sshane:


opendbc/safety/safety_declarations.h
```diff
@@ -121,10 +121,11 @@ typedef struct {
   const int addr;
   const int bus;
   const int len;
-  const bool check_checksum;         // true is checksum check is performed
+  const bool ignore_checksum;        // checksum check is not performed when set to true
+  const bool ignore_counter;         // counter check is not performed when set to true
```
adeebshihadeh, line 125: this name isn't quite correct. the name should make it clear that it doesn't have a checksum or counter

sshane: Ford implements checksum and counters for some messages we ignore due to those checksums and counters occasionally being wrong, but that's the exception.

PR #1946, opened 2025-03-07
draft - Acura RDX 2022-2023

mvl-boston:
draft PR for Acura RDX 2022-2023

work in process

Current state:
- working without longitudinal in 2022, longitudinal throws canbus errors.   (also a test of 098 release-staging does the same thing, and didn't fail on 097, so I think this is a core openpilot problem)   Trying to identify changes between 098 and 097.
- 2023 fingerprint is not yet added, community member is planning to buy a comma and will add fingerprint once tested


mvl-boston:
Made a mistake and corrupted commit history with github desktop.    Will create a brand new PR.

PR #1953, opened 2025-03-08
Toyota: Cleanup/fix safety RxChecks for SecOC

jyoung8607:


sshane:
Some extra messages were in SECOC so I moved them

Issue #1950, opened 2025-03-08
safety fwd hook should be a config/struct


sshane:
re-merge https://github.com/commaai/opendbc/pull/1785 but with everything

- [ ] dynamic AEB (or generic message?) forwarding

sshane:
done

PR #1903, opened 2025-03-03
safety: only run rx hooks on whitelisted msgs

sshane:
https://github.com/commaai/opendbc/pull/1851 but for rx

catches a few missing rx checks:

```
FAILED test_models.py::TestCarModel_14_CHEVROLET_BOLT_EUV::test_panda_safety_carstate@CarTestRoute(route='f08912a233c1584f|2022-08-11--18-02-41', car_model=<CAR.CHEVROLET_BOLT_EUV>, segment=1) - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'regenBraking': 312}
FAILED test_models.py::TestCarModel_20_CHEVROLET_VOLT::test_panda_safety_carstate@CarTestRoute(route='c950e28c26b5b168|2018-05-30--22-03-41', car_model=<CAR.CHEVROLET_VOLT>, segment=None) - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'regenBraking': 30}
FAILED test_models.py::TestCarModel_43_GENESIS_GV60_EV_1ST_GEN::test_panda_safety_carstate@CarTestRoute(route='b5d6dc830ad63071|2022-12-12--21-28-25', car_model=<CAR.GENESIS_GV60_EV_1ST_GEN>, segment=12) - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'controlsAllowed': 9}
FAILED test_models.py::TestCarModel_183_RIVIAN_R1_GEN1::test_panda_safety_carstate_fuzzy@CarTestRoute(route='bc095dc92e101734/000000db--ee9fe46e57', car_model=<CAR.RIVIAN_R1_GEN1>, segment=None) - IndexError: list index out of range
FAILED test_models.py::TestCarModel_241_TOYOTA_RAV4_PRIME::test_panda_safety_carstate@CarTestRoute(route='20ba9ade056a8c7b|2021-02-08--21-57-35', car_model=<CAR.TOYOTA_RAV4_PRIME>, segment=None) - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'gasPressed': 349}
FAILED test_models.py::TestCarModel_247_TOYOTA_SIENNA::test_panda_safety_carstate@CarTestRoute(route='eb6acd681135480d|2019-06-20--20-00-00', car_model=<CAR.TOYOTA_SIENNA>, segment=None) - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'brakePressed': 184}
```

PR #1954, opened 2025-03-08
safety: fwd hook config

sshane:
Closes https://github.com/commaai/opendbc/issues/1950

v2 of https://github.com/commaai/opendbc/pull/1785

removes turing completeness

haven't checked all cases, but we might be able to just do the relay malfunction part of https://github.com/commaai/opendbc/pull/1903 with nothing more than `blocked` and a few more lines (plus checking a bunch more addresses automatically)!


opendbc/safety/safety_declarations.h
```diff
@@ -13,12 +13,14 @@
   do { \
     (config).rx_checks = (rx); \
     (config).rx_checks_len = sizeof((rx)) / sizeof((rx)[0]); \
+    (config).fwd_buses = NULL; \
```
sshane, line 16: this is needed since we commonly do `safety_config ret;` everywhere which doesn't zero initialize this to NULL.

TODO: change to a forwarding boolean since everything uses 0: 2, 2: 0 except all_output which is based on a dynamic param

sshane:
blockers:
- [ ] Hyundai CAN FD: needs clean up, and we forward/block when bus is 1?
- [ ] Subaru: clean up, some messages that we either block or don't block are in common
- [ ] Honda: dynamic AEB fwding
- [ ] Tesla: dynamic AEB fwding

opendbc/safety/safety_declarations.h
```diff
@@ -49,8 +52,14 @@ typedef struct {
   int addr;
   int bus;
   int len;
+  bool blocked;
```
sshane, line 55:
```suggestion
  bool block_fwding;  // or no_fwding
```

makes this more clear

opendbc/safety/safety.h
```diff
@@ -253,7 +253,22 @@ bool safety_tx_hook(CANPacket_t *to_send) {
 }
 
 int safety_fwd_hook(int bus_num, int addr) {
-  return (relay_malfunction ? -1 : current_hooks->fwd(bus_num, addr));
+  const int destination_bus = bus_num == 0 ? 2 : (bus_num == 2 ? 0 : -1);
```
adeebshihadeh, line 256: i'd spread this over a few lines of if/else for readability

opendbc/safety/safety_declarations.h
```diff
@@ -151,6 +160,7 @@ typedef struct {
   int rx_checks_len;
   const CanMsg *tx_msgs;
   int tx_msgs_len;
+  bool block_fwding;
```
adeebshihadeh, line 163: I'd call this `enable_forwarding` or `disable_forwarding`

adeebshihadeh (commented):
Looks great!

sshane:
Done in Done in https://github.com/commaai/opendbc/pull/2095

PR #1959, opened 2025-03-09
safety: relay malfunction config

sshane:
split from https://github.com/commaai/opendbc/pull/1954


adeebshihadeh:
Why can't this be the same config as blocked?

sshane:
It will be, I'm just trying to merge the simplest thing first.

Issue #1637, opened 2025-01-14
Honda: Third-gen facelift Acura MDX needs a separate car port


jyoung8607:
The [Acura MDX got a facelift for model-year 2022](https://en.wikipedia.org/wiki/Acura_RDX#2021_facelift), which came with a new camera/radar architecture. This architecture is also shared by newer Honda Odyssey.

Correctly supporting these cars requires a new model port with minor platform changes. The CAN messaging is nearly identical, but there are architectural changes that affect the minimum speed for lateral control and cluster display of the cruise setpoint. A port was started in commaai/openpilot#25481 by @csouers but wasn't quite completed or merged.

In the meantime, starting with commaai/openpilot#27812, we inadvertently started recognizing them as pre-facelift cars. So we need the MY2022 port **and** then we need to disentangle all the firmware that was added to ACURA_RDX_3G. This will be more than just reverting the one PR, the fuzzy fingerprinting bot may have added more on top.

- [ ] Temporarily revoke documented support for MY2022 Acura MDX
- [ ] Merge a working port for MY2022 and forward
- [ ] Disentangle UDS firmware between Acura MDX 3G and 3G-facelift
- [ ] Restore MY2022 and forward Acura MDX as officially supported

mvl-boston:
Thank you, for merging.   Let me know if you would like me to propose code for this.   In the past week I have looked through all the items mentioned and can follow everything except the fuzzy logic, and can support as needed.

There is also a problem with the longitudinal acceleration/develeration tuning parameters and I do not know if that is unique to the 2022 RDX or also impacts other bosch models.

mvl-boston:
Issue 1638 links to the RDX information.

The MDX is a separate Acura vehicle that has its own separate incompatibility issues.    The subject line for this issue and linked issue #1638 should say RDX.

jyoung8607:
My mistake, I'll sort it out shortly. Thanks!

mvl-boston:
Port is about 75% complete:
- stable (but not properly fingerprinted 0.97 openpilot) code - https://github.com/mvl-boston/openpilot/tree/RDX_stable_097
- 0.98 opendbc port in process - https://github.com/mvl-boston/opendbc

I don't have edit rights to this bug but it still improperly says RDX.

Remaining tasks:
- Learn how to use the simulator to pass relevant safety testing (and what that testing is)
- Testdrive the opendbc version
- Look into the repeated missing vin in the error log (might need a new DBC, will poke around cabana)

[...]

PR #1967, opened 2025-03-10
Acura 2022 RDX support

mvl-boston:
PR for Acura RDX 2022

Replacement for [PR1946](https://github.com/commaai/opendbc/pull/1946)
Resolution for [bug 1647](https://github.com/commaai/opendbc/issues/1637)

Notes:
- car fingerprint named 3G MMR. That is honda-speak for 3rd generation mid-model refresh.   Segregated 2022 fingerprints from 2019-2021 fingerprints.
- RDX uses imperial speed/cruise units on the canbus in the USA, Canadian RDXs probably use metric.  Rewrote the cruise unit to read the metric canbus flag, rather than be based on hardcoded fingerprint names, to automate support of this.
- RDX removes steering when under 70kph with radar on (eg: when longitudinal is off). Identified the canbus flag that identifies the steering lockout, and set a low speed warning when this occurs.   Will also serve as a warning to steering lockout that occurs when a driver adds too much manual torque.

Current state:
- per feedback on Honda/Acura discord, moving this back to draft for the following reasons:
1. Need a route with full rlogs
2. Need to isolate changes specific to the RDX to prevent risk of regression to other cars.

- due to issues with longitudinal, and longitudinal still being an alpha product, I removed longitudinal tuning and will add support later when I can retest it.  Did not want to hold up pr so that car can be added back to compatibility list.
- regression on gas strength found that this model runs 1 m/s per 1000 gas units. Existing code has 1m/s per 800 units for Bosch. Unsure if this model is different (or other Bosch models are tuned wrong) so previously wrote logic that flips to 1000 units only for this fingerprint.  Delayed this logic until longitudinal is working/tested, as non-longitudinal does not use this feature, to minimize complexity of PR.
- have not heard back from community member with 2023 RDX model for fingerprint.   Have a portable mount and will see if local dealership will let me fingerprint their car instead.


Yash0270:
Hey! thanks for this PR, eager waiting for this PR to be merged. I own a 2022 RDX and want to get comma 3x for it.

jyoung8607:
@mvl-boston did you mean to close this? Now that Odyssey is in, this was ready to be rebased, tested, and reviewed.

mvl-boston:
I didn't mean to close.   I see what happened, deleted dead 22-mdx older forks, and accidentally deleted the rdx in the process.  Reopening.

jyoung8607:
@mvl-boston This one seems ready-to-go except for current test data. The dongle ID from the CI test route hasn't driven since January. Do you have any other testers with this car?

mvl-boston:
Jan dongle was mine (since warranty swapped) dongle from my wife's now-offlimits car.

Tester hasn't had a chance to get a route to me yet.   Had a couple family issues past month, let me check back.

mvl-boston:
Still no update from testers.

Second tester seems to have a bad C3 (canTX failing on separate brand cars, multiple harnesses, multiple harness boxes).   Owner is trying some board repair and if it fails will get in line for a C4.

Separately, I have a 23 RDX booked for vacation in late April.  If testers don't get back to me by then, then I'll get new test routes myself.

PR #1965, opened 2025-03-10
Hyundai Safety: Cleanup V2

sunnyhaibin:
- Part of https://github.com/commaai/opendbc/issues/1872

Parent of
- https://github.com/commaai/opendbc/pull/1968
- https://github.com/commaai/opendbc/pull/1964

Issue #1942, opened 2025-03-07
Clean up Hyundai safety and car interface


sshane:
Blocking new model ports until this is done, it's almost impossible to work on the safety without a good chance of introducing bugs.

sshane:
```
  CAN-FD only flags without CANFD_ prefix:
  - CHECKSUM_CRC8 (line 126) — only applies to CAN-FD
  - CHECKSUM_6B (line 127) — only applies to CAN-FD
  - CANFD_NO_RADAR_DISABLE (line 137) — this one does have the prefix

  Flags with CANFD_ prefix:
  - CANFD_LKA_STEER_MSG, CANFD_ALT_BUTTONS, CANFD_ALT_GEARS, CANFD_CAMERA_SCC, CANFD_ALT_GEARS_2, CANFD_LKA_STEER_MSG_ALT, CANFD_ANGLE_STEERING, CANFD_NO_RADAR_DISABLE

  Ambiguous ones:
  - SEND_LFA (line 106) — is this CAN-FD only?
  - USE_FCA (line 107) — same question
  - ALT_LIMITS / ALT_LIMITS_2 — no prefix, but are these CAN-FD only?
  - ```

johnnathanwood:
+1 on this direction — the `ALT_LIMITS`-style flags are a big part of what makes the Hyundai safety hard to reason about (#3026 "we want to move away from ALT_LIMITS-style params", #2072 "this is exactly why we can't continue with alt limit flags").

One question on how the torque path factors in here: the angle and curvature checks already run through the car-independent ISO-11270 envelope in `lateral.h` (`ISO_LATERAL_ACCEL` + jerk), but `steer_torque_cmd_checks` still just caps against the fixed per-car max-torque tier in `hyundai.h` (384/270/170). Is the end-state to bring the legacy CAN cars under that same envelope, so the tier stops being load-bearing and folds into the CAN-only namespace with the rest of these?

Happy to help with the groundwork — I know #1964 didn't land, but I'd gladly pick up a slice of it. (Got some Kona torque telemetry on #1137 if it's ever useful; won't clutter this thread with it.)

PR #1968, opened 2025-03-10
Hyundai CAN FD Safety: Split TX and RX checks assignments

sunnyhaibin:
- Split from https://github.com/commaai/opendbc/pull/1965
- Splits TX and RX safety checks into separate sections
- Replaces `BUILD_SAFETY_CFG` with dedicated `SET_TX_MSGS` and `SET_RX_CHECKS`
- No functional changes
- Part of https://github.com/commaai/opendbc/issues/1942


sshane:
I worry that this is harder to keep track of because now the if statements are duplicated

PR #1972, opened 2025-03-10
Hyundai CAN FD: Universal accelerator pedal signal

sunnyhaibin:
Red diff and simplified `gasPressed` 😄

- [ ] Validate with notebook
- [x] `test_models` validation


sunnyhaibin:
Unfortunately, a few models across the CAN FD fleet do not utilize the signal at 0x175 for the accelerator pedal press.

PR #1973, opened 2025-03-11
Hyundai: dashcam FCEV

sshane:
Dashcamming the car from https://github.com/commaai/opendbc/pull/933 as it's missing an rx check

Bug is in the safety tests in https://github.com/commaai/opendbc/pull/1903. @sunnyhaibin can you fix?


sunnyhaibin:
Should I open a PR for `disable-hyundai-fcev` or `master`?

PR #1974, opened 2025-03-11
Hyundai: Fix FCEV RX check and move out of dashcam

sunnyhaibin:
Fixes https://github.com/commaai/opendbc/pull/1973.

Tested with https://github.com/commaai/opendbc/pull/1903 as base and it passed `test_models` locally.

Split `0x91` into its own macro to ensure RX check is included for FCEV accelerator pedal pressed.


opendbc/safety/safety/safety_hyundai.h
```diff
@@ -298,8 +299,22 @@ static safety_config hyundai_init(uint16_t param) {
...
+      SET_RX_CHECKS(hyundai_fcev_long_rx_checks, ret);
+    } else {
+      SET_RX_CHECKS(hyundai_long_rx_checks, ret);
+    }
+
+    if (hyundai_camera_scc) {
+      SET_TX_MSGS(HYUNDAI_CAMERA_SCC_LONG_TX_MSGS, ret);
+    } else {
+      SET_TX_MSGS(HYUNDAI_LONG_TX_MSGS, ret);
+    }
```
sshane, line 317: @sunnyhaibin even though it's longer, this is easier to read

sshane: and it finds a missing test!

PR #1996, opened 2025-03-14
fix ruff failure

sshane:


sunnyhaibin:
Was pulling my hair out why it started catching C420 all of the sudden with zero change 😅

PR #1998, opened 2025-03-14
safety: start per-message fwd config

sshane:
Split from https://github.com/commaai/opendbc/pull/1954 with only brands that do not have any conflicts and for only currently checked relay msgs first


sshane:
Done in https://github.com/commaai/opendbc/pull/2095

PR #2000, opened 2025-03-14
Ford: radar is invalid when in reverse

sshane:
Radar just sends the last message while in reverse (for everything; tracks, headers, etc.), so we need to detect a stalled scan index and add a fault instead of stop sending liveTracks:

image

### DIFF
--- a/opendbc/car/car.capnp
+++ b/opendbc/car/car.capnp
@@ -307,6 +307,7 @@ struct RadarData @0x888ad6581cf0aacb {
     canError @0;
     fault @1;
     wrongConfig @2;
+    unavailableTemporary @3;  # radar data is temporarily unavailable due to conditions the car sets
   }
 
   # similar to LiveTracks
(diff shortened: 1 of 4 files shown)


opendbc/car/car.capnp
```diff
@@ -307,6 +307,7 @@ struct RadarData @0x888ad6581cf0aacb {
     canError @0;
     fault @1;
     wrongConfig @2;
+    unavailableTemporary @3;  # radar data is temporarily unavailable due to conditions the car sets
```
adeebshihadeh, line 310: let's make these bool fields

adeebshihadeh (changes requested):

opendbc/car/car.capnp
sshane: https://github.com/commaai/opendbc/pull/2001

sunnyhaibin:
This the same for Hyundai when it's not in Drive gear.

sshane:
Want to make an issue?

sunnyhaibin:
> Want to make an issue with a route?

Opened issue.
- https://github.com/commaai/opendbc/issues/2003

PR #2002, opened 2025-03-15
car interfaces: each specify their own dependencies

sshane:
Removes need to define empty radar interface file, one less thing when porting a car. pytest collection time is not sped up with the static imports, so leaving that out until there's a cleaner solution


opendbc/car/car_helpers.py
```diff
@@ -40,6 +41,9 @@ def _get_interface_names() -> dict[str, list[str]]:
 # imports from directory opendbc/car/<name>/
 interface_names = _get_interface_names()
 interfaces = load_interfaces(interface_names)
+INTERFACES = {
+  "toyota": (ToyotaCarInterface, ToyotaCarInterface.CarController, ToyotaCarInterface.CarState, ToyotaCarInterface.RadarInterface)
```
sshane, line 45:
```suggestion
  "toyota": ToyotaCarInterface,
```

PR #2004, opened 2025-03-15
Hyundai Legacy Safety: ignore checksum and counters in rx checks

### LINKED ISSUE #1994: Controls mismatch alert and subsequent disengage
This issue was originally posted under OpenPilot #34677 and moved here per request.
---

Seemingly random controls mismatch alert and subsequent disengage.

I have been unable to notice a rhythm this behavior. Sometimes it happens very quick after engagement, while other times it takes some time.

Sometimes it happens once then calms down while other times it will happen repeatedly in quick succession. 

Operation had been rock solid on older builds. I have not pinned down at which version the issue started.

The system can immediately be re-engaged after alarm.
2015 Hyundai Genesis
3687c15bdfe09295/0000005f--888dec6063
Release 0.9.7

sunnyhaibin:
Fixes https://github.com/commaai/opendbc/issues/1994

After fix:
```
replaying 3687c15bdfe09295/0000005f--888dec6063 with safety mode 23, param 0, alternative experience 1

RX
total rx msgs: 6226045
invalid rx msgs: 0
safety tick rx invalid: False
invalid addrs: set()

TX
total openpilot msgs: 160558
total msgs with controls allowed: 119208
blocked msgs: 10
blocked with controls allowed: 0
blocked addrs: Counter({1265: 9, 832: 1})
```

### DIFF
--- a/opendbc/safety/safety/safety_hyundai.h
+++ b/opendbc/safety/safety/safety_hyundai.h
@@ -330,7 +330,7 @@ static safety_config hyundai_init(uint16_t param) {
 
     static RxCheck hyundai_fcev_rx_checks[] = {
       HYUNDAI_COMMON_RX_CHECKS(false)
-      HYUNDAI_SCC12_ADDR_CHECK(0)
+      HYUNDAI_SCC12_ADDR_CHECK(0, false)
       HYUNDAI_FCEV_GAS_ADDR_CHECK
     };
 
--- a/opendbc/safety/safety/safety_hyundai.h
+++ b/opendbc/safety/safety/safety_hyundai.h
@@ -47,8 +47,8 @@ const LongitudinalLimits HYUNDAI_LONG_LIMITS = {
   {.msg = {{0x251, 0, 8, .ignore_checksum = true, .ignore_counter = true, .frequency = 50U}, { 0 }, { 0 }}},                                              \
   {.msg = {{0x4F1, 0, 4, .ignore_checksum = true, .max_counter = 15U, .frequency = 50U}, { 0 }, { 0 }}},                                                  \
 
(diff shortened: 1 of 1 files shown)


sshane:
Can we add a route to test models? It should have caught this

sunnyhaibin:
> Can we add a route to test models? It should have caught this

@MrXelious is this route okay to be public for tests in openpilot? `3687c15bdfe09295/0000005f--888dec6063`

If not, please get another route with rlogs uploaded and preserved in areas that are public.

MrXelious:
> > Can we add a route to test models? It should have caught this
> 
> @MrXelious is this route okay to be public for tests in openpilot? `3687c15bdfe09295/0000005f--888dec6063`
> 
> If not, please get another route with rlogs uploaded and preserved in areas that are public.

Let me get you a route that doesn't have a specific stopping point. I think I have one saved. 

As for testing the changes, I will do so ASAP. Though, I just dropped my car off at the body shop the other day so it's not in my possession right now.

I will report back.

MrXelious:
> > > Can we add a route to test models? It should have caught this
> > 
> > @MrXelious is this route okay to be public for tests in openpilot? `3687c15bdfe09295/0000005f--888dec6063`
> > 
> > If not, please get another route with rlogs uploaded and preserved in areas that are public.
> 
> Let me get you a route that doesn't have a specific stopping point. I think I have one saved. 
> 
> As for testing the changes, I will do so ASAP. Though, I just dropped my car off at the body shop the other day so it's not in my possession right now.
> 
> I will report back.

If you're able to trim logs before pushing as a test model, snipping the last few minutes of this drive would be adequate. Just getting rid of the destination is all.

sshane:
See issue.

PR #2019, opened 2025-03-20
VW PQ: Rename DBC

EliasKotlyar:


opendbc/can/dbc.cc
```diff
@@ -59,7 +59,7 @@ ChecksumState* get_checksum(const std::string& dbc_name) {
     s = new ChecksumState({16, -1, 0, -1, true, HKG_CAN_FD_CHECKSUM, &hkg_can_fd_checksum});
   } else if (startswith(dbc_name, {"vw_mqb_2010", "vw_mqbevo", "vw_meb"})) {
     s = new ChecksumState({8, 4, 0, 0, true, VOLKSWAGEN_MQB_MEB_CHECKSUM, &volkswagen_mqb_meb_checksum});
-  } else if (startswith(dbc_name, "vw_golf_mk4")) {
+  } else if (startswith(dbc_name, "vw_golf_mk5_pq35")) {
```
jyoung8607, line 62: I'm generally supportive of this, have been meaning to do it for awhile, but let's just go with PQ for the name since we support PQ35 / PQ46 / NMS and more with a single DBC.

```suggestion
  } else if (startswith(dbc_name, "vw_pq")) {
```

EliasKotlyar: @jyoung8607 Many thanks, for the suggestion. I have renamed it again. Please have a look!

PCBPQ35: This DBC looks to be a hybrid of old/new pq stuff and mixed together canbusses, shouldnt it be called "vw_pq_summary" or "vw_pq_general"?

jyoung8607: If we arrive at a need to divide NMS/PQ35/PQ46, and/or separate the various buses, we'll cross that bridge when we come to it (and do so using the generator).

PR #2025, opened 2025-03-22
safety tests: remove steer bit rt limit param

sshane:
Similar to https://github.com/commaai/opendbc/pull/2024

### DIFF
--- a/opendbc/safety/tests/common.py
+++ b/opendbc/safety/tests/common.py
@@ -266,7 +266,11 @@ class SteerRequestCutSafetyTest(TorqueSteeringSafetyTestBase, abc.ABC):
   # Safety around steering request bit mismatch tolerance
   MIN_VALID_STEERING_FRAMES: int
   MAX_INVALID_STEERING_FRAMES: int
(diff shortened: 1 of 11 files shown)


opendbc/safety/tests/common.py
```diff
@@ -266,7 +266,11 @@ def setUpClass(cls):
   # Safety around steering request bit mismatch tolerance
   MIN_VALID_STEERING_FRAMES: int
   MAX_INVALID_STEERING_FRAMES: int
-  MIN_VALID_STEERING_RT_INTERVAL: int
+  STEER_STEP: int = 1
+
+  @property
+  def MIN_VALID_STEERING_RT_INTERVAL(self):
+    return int((self.MIN_VALID_STEERING_FRAMES + 1) * self.STEER_STEP * 10000 * 0.9)
```
sshane, line 273:
```suggestion
    # a ~10% buffer
    return int((self.MIN_VALID_STEERING_FRAMES + 1) * self.STEER_STEP * 10000 * 0.9)
```

PR #2030, opened 2025-03-24
docs: remove comma power from J533 harness

incognitojam:
The J533 harness has constant power from the gateway.


jyoung8607 (approved):
I think we can go ahead and remove it.  It's completely fine for normal supported cars. It's _mostly_ fine for cars without factory LKA; we don't need it for normal driving but we need it for a [one-time config task](https://github.com/commaai/openpilot/blob/master/selfdrive/debug/vw_mqb_config.py).

The folks without factory LKA already know they don't have full official support. We can tell them to manually add it to their order, or to use a diagnostic tool like VCDS or OBDeleven. Just noting because we have data showing they're ~15% of VW users.

Issue #2012, opened 2025-03-18
Intermittent "Dashcam Mode" error on startup


Verylukyguy:
**Issue:**  Intermittent "Dashcam Mode" error on startup, usually resolves if restart again.

**Vehicle:**  2018 GMC Acadia

**Route:**  df51b2195de66a9b/00000002--63c7249888/0

**OpenPilot version:**  0.9.9 nightly e31c87e

sshane:
When your fingerprint is cached, the radar's delayed startup can cause a false positive radarUnavailable.

![Image](https://github.com/user-attachments/assets/a76586b1-4466-4991-843b-0865fe2d0d3d)

PR #2033, opened 2025-03-24
GM: check camera object data for radarUnavailable

sshane:
Closes https://github.com/commaai/opendbc/issues/2012

### DIFF
--- a/opendbc/car/gm/interface.py
+++ b/opendbc/car/gm/interface.py
@@ -7,7 +7,7 @@ from opendbc.car.common.basedir import BASEDIR
 from opendbc.car.common.conversions import Conversions as CV
 from opendbc.car.gm.carcontroller import CarController
 from opendbc.car.gm.carstate import CarState
-from opendbc.car.gm.radar_interface import RadarInterface, RADAR_HEADER_MSG
+from opendbc.car.gm.radar_interface import RadarInterface, RADAR_HEADER_MSG, CAMERA_DATA_HEADER_MSG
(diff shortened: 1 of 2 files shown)


sshane:
@nworb-cire can you provide a few dongles of Volts without a radar that we support in dashcam mode (supported here: https://github.com/commaai/openpilot/pull/27245)?

nworb-cire:
This is mine: `806907cd33ef2647`

I don’t keep records of anyone else’s. Let me know if you need me to provide a route on master; most of my recent drives have been on my fork.

sshane:
Ah I recall us discussing that these radarless cars have no bus 1 traffic, your routes match that: `...1907: 7}, 1: {}, 2: {560:...`

opendbc/car/gm/interface.py
```diff
@@ -127,7 +127,8 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, experime
     else:  # ASCM, OBD-II harness
       ret.openpilotLongitudinalControl = True
       ret.networkLocation = NetworkLocation.gateway
-      ret.radarUnavailable = RADAR_HEADER_MSG not in fingerprint[CanBus.OBSTACLE] and not docs
+      # LRR messages can take up to a few seconds to start sending after ignition, check camera data as well
```
sshane, line 130:
```suggestion
      # LRR messages can take up to a few seconds to start sending after ignition, check camera data as well which starts earlier
```

Issue #1917, opened 2025-03-04
[$500 bounty] Tesla: fuzzy fingerprinting


adeebshihadeh:
See the [Ford bounty](https://github.com/commaai/openpilot/issues/31052) for context. 

Requirements
* unit tests
* works for all supported cars

We also need to answer these questions:
* has Tesla ever changed the APIs we use? 
* are there any variants of our supported cars that won't work?

adeebshihadeh:
@lukasloetkolben interested in this one?

lukasloetkolben:
> [@lukasloetkolben](https://github.com/lukasloetkolben) interested in this one?

I am a bit busy, if it is not done by next week I will do it.

xunwen-art:
I'd like to work on this bounty. 

Plan:
1. Study existing fingerprinting implementation
2. Implement fuzzy matching based on FW versions
3. Add unit tests for all supported Tesla models
4. Test edge cases (API changes, variants)

Questions to answer:
- Check if Tesla has changed APIs
- Verify all variants compatibility

ETA: 3-5 days

xunwen-art:
## Implementation Ready

I've implemented fuzzy fingerprinting for Tesla vehicles. Due to fork issues, posting the code here for review:

### Files Changed

**1. New file: `opendbc/car/tesla/fuzzy_fp.py`**
```python
"""Tesla fuzzy fingerprinting implementation."""
import re
from opendbc.car.fw_query_definitions import LiveFwVersions, OfflineFwVersions
from opendbc.car.structs import CarParams
from opendbc.car.tesla.values import CAR

Ecu = CarParams.Ecu

# Tesla FW version pattern examples:
# TeM3_E014p10_0.0.0 (16),E014.17.00        -> Model 3
# TeMYG4_Main_0.0.0 (77),Y4003.05.4          -> Model Y Gen4
# TeM3_SP_XP002p2_0.0.0 (23),XPR003.6.0      -> Model X

FW_PATTERN = re.compile(
  r'^Te'                           # Tesla prefix
  r'(?P<platform>[A-Z0-9]+?)_'     # Platform: M3, MYG4
  r'(?P<variant>[A-Za-z0-9_]+)_'   # Variant: E014p10, Main, SP
  r'[\d.]+\s*\(\d+\),'             # Version with build
  r'(?P<ver>[A-Z][A-Z0-9.]+)$'     # Final version: E014.17.00, Y4003.05.4
)

[...]

PR #2044, opened 2025-03-25
Add Tesla Fuzzy Fingerprint

ssysm:
Allow Tesla to be fuzzy matched. Using the the `TeM` identifier for make id and model code after `(25),`.

### DIFF
--- a/opendbc/car/tesla/values.py
+++ b/opendbc/car/tesla/values.py
@@ -3,7 +3,7 @@ from enum import Enum, IntFlag
 from opendbc.car import Bus, CarSpecs, DbcDict, PlatformConfig, Platforms, AngleSteeringLimits
 from opendbc.car.structs import CarParams, CarState
 from opendbc.car.docs_definitions import CarDocs, CarFootnote, CarHarness, CarParts, Column
-from opendbc.car.fw_query_definitions import FwQueryConfig, Request, StdQueries
+from opendbc.car.fw_query_definitions import FwQueryConfig, Request, StdQueries, LiveFwVersions, OfflineFwVersions
 
 Ecu = CarParams.Ecu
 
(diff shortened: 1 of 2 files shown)


sshane:
We need to make sure that this can restrict Teslas within a certain model year/capability range. For example, not HW2 (or others), and not Model S/X right now.

opendbc/car/tesla/values.py
```diff
@@ -34,6 +34,36 @@ class TeslaCarDocsHW4(CarDocs):
...
+
+      live_model = decode_tesla_fingerprint_model(live_current_fw)
+      if live_model is None:
+        continue
+      if live_model == current_model:
+        candidates.add(candidate)
+  return candidates
+
+def decode_tesla_fingerprint_model(fingerprint: bytes) -> tuple:
+  if not fingerprint.startswith(b'TeM'): # Not a tesla
```
sshane, line 60:
```suggestion
  if not fingerprint.startswith(b'TeM'):  # Not a tesla
```

two spaces before comments

opendbc/car/tesla/values.py
```diff
@@ -34,6 +34,36 @@ class TeslaCarDocsHW4(CarDocs):
...
+    return None
+  model_split = fingerprint.split(b'),')
+  if len(model_split) != 2:
+    return None
+  model = model_split[1][0]
+  return chr(model)
```
sshane, line 66: two lines between functions

opendbc/car/tesla/values.py
```diff
@@ -34,6 +34,36 @@ class TeslaCarDocsHW4(CarDocs):
...
+        candidates.add(candidate)
+  return candidates
+
+def decode_tesla_fingerprint_model(fingerprint: bytes) -> tuple:
+  if not fingerprint.startswith(b'TeM'): # Not a tesla
+    return None
+  model_split = fingerprint.split(b'),')
+  if len(model_split) != 2:
+    return None
+  model = model_split[1][0]
```
sshane, line 65: this should be robust to garbage input, i.e. no Python exceptions. We use a simple regex pattern for other brands, it ensures the parsing logic is in one place and if there's no match, it's obvious what it will return (None)

sshane:
Also see this comment: https://github.com/commaai/opendbc/issues/1917#issuecomment-2740315033

ssysm:
Will check, thanks for the info

sshane:
We can potentially detect these missing APIs in the interface and set the dashcam flag if needed: https://github.com/commaai/opendbc/blob/1cc799801f03a45180de1e64193981b102aec5c4/opendbc/car/ford/interface.py#L70-L79

ssysm:
Changed the firmware decode to a regex, should be able to extract the model code from the firmware. Added a VIN check to restrict only 2018+ cars [Math](https://www.findmyelectric.com/tesla-vin-decoder/). 

As of HW2.5.. I'm not really sure, https://github.com/commaai/opendbc/issues/1917#issuecomment-2740315033. As far I'm seeing in from the route that lukas posted, the `autoPark` signal stayed at `unavailable` the whole trip in `DI_State` and `DAS_Control`  message disappeared after few seconds. Maybe this can give a clue about if ACC is equipped on HW2.5?

opendbc/car/tesla/values.py
```diff
@@ -36,34 +37,50 @@ class TeslaPlatformConfig(PlatformConfig):
 
 def match_fw_to_car_fuzzy(live_fw_versions: LiveFwVersions, vin: str, offline_fw_versions: OfflineFwVersions) -> set[str]:
   candidates: set[str] = set()
+
+  # Check year is greater than 2018, but only if VIN is valid
+  if ord(vin[9]) < 74 and vin != '00000000000000000':
```
sshane: You can keep the vin constants

opendbc/car/tesla/values.py
```diff
@@ -110,3 +159,14 @@ class TeslaFlags(IntFlag):
 DBC = CAR.create_dbc_map()
 
 STEER_THRESHOLD = 0.5
+
+AP_VERSIONS = {'HW3':'3', 'HW4':'YG4'}
```
lukasloetkolben, line 163: YG4 is not an identifier for HW4

sshane:
We should only ever use the VIN or the fingerprint to make this easier to reason about

PR #2016, opened 2025-03-19
Rivian: increase low speed torque

sshane:
2 things noticed from plotting torque points in 3D from torqued on this route: `2d2da5dea63b5c5e/00000011--fc863c57a2/q`

- Lateral acceleration response to torque looks linear (as it should)
- Lateral acceleration achieved across 15-30 m/s^2 looks to be the same (as it should)

But this goes against what we found in the lateral injection test. Perhaps higher torques become non-linear, need to plot more data.

https://github.com/user-attachments/assets/990400dc-b3f4-4651-841a-1fa0cfda38da

PR #2047, opened 2025-03-26
safety: support dynamic max torque from speed

sshane:
For https://github.com/commaai/opendbc/pull/2016


devtekve:
Excellent! I've been doing something similar for my ioniq5 PE. Although on hyundai angle cnafd the torque does not seem to be linear to speed  as you saw for rivian (maybe the torque limit ceiling is?), I believe it's related to curvature info. Nonetheless, the use of speed feels the right way

Issue #1999, opened 2025-03-14
Rivian safety: check 2nd speed source


sshane:
Will allow us to boost torque at low speed

sshane:
Done! https://github.com/commaai/opendbc/pull/2049

PR #2048, opened 2025-03-26
Rivian: add all validity checks for speed message

sshane:
For https://github.com/commaai/opendbc/pull/2016. After this we need https://github.com/commaai/opendbc/issues/1999


sshane:
Passes test_models

PR #2050, opened 2025-03-26
Honda: fix radarless braking

mvl-boston:
2025/2026 Civic Hybrid and 2025 Acura ADX do not brake at full requested strength.

Initial fix attempts (now abandoned) focused on radarless camera.

Instead, recently found that the idlestop message is on all the time while stock ACC does braking.   Replicating this (setting bit whenever braking) fixed braking strength on ADX and Civic Hybrid.

- Before: ad9840558640c31d/00000023--67187e578c
- After: ad9840558640c31d/0000002b--8d05b8be4c

Todo:
- [x] Test Hybrid Civic - fceae97165094dc2/00000002--2f81703767
- [x] Regression test ICE Civic - a2152effbad22b70/000003e3--0f13f16c7b


sshane:
What does this fix exactly?

mvl-boston:
> What does this fix exactly?

User in honda discord has logs showing that braking is maxing out at -0.5m/s2, ignoring requests up to -3.5m/s2.

Not sure this works, but the goal is to imitate a new brake channel that ACC sends on hybrids to see if it fixes it.   Testdrive should be imminent.

The fix is similar to one I had do to for the MDX Hybrid PR.

(I set as draft PR because I like the CI code review stuff)

mvl-boston:
Revised approach to redefine a brake message, to align with how stock ACC does braking.   Code was force-pushed to clean/new working PR.

Tests were successful.   This is ready for review.

opendbc/car/honda/hondacan.py
```diff
@@ -85,20 +85,19 @@ def create_acc_commands(packer, CAN, enabled, active, accel, gas, stopping_count
   acc_control_values = {
     'ACCEL_COMMAND': accel_command,
     'STANDSTILL': standstill,
+    'BRAKE_REQUEST': braking,
   }
 
   if car_fingerprint in HONDA_BOSCH_RADARLESS:
     acc_control_values.update({
       "CONTROL_ON": enabled,
-      "IDLESTOP_ALLOW": stopping_counter > 200,  # allow idle stop after 4 seconds (50 Hz)
```
sshane, line 92: was the signal defined incorrectly? does it not do stop/start anymore with this?

sshane:
Huh, looks like this was added for the initial radarless Civics we supported: https://github.com/commaai/openpilot/pull/27722

sshane:
Is there a stock route I can look at?

MarcoTheDingo:
> Is there a stock route I can look at?

Here's a stock route from a couple of weeks ago on my 22 Civic ICE
a2152effbad22b70/000004a5--2c90123137

opendbc/car/honda/hondacan.py
mvl-boston: My guess is that it is defined incorrectly.   I don't see how Honda would have allowed delaying of idlestop for 4 seconds, that's not something that occurs on stock.   Stock goes into idlestop immediately upon standstill on all Honda/Acura cars I've driven.  It probably was a braking message all along.

With this bit set on OP, the car goes into idlestop immediately upon standstill, just like stock.

As discussed above in the initial comments, the Civic Hybrid and ADX ICE will not brake properly without this message set.   They are limited to approximately -0.5m/s2 without this bit, which might be just wind resistance.

sshane:
> > Is there a stock route I can look at?
> 
> Here's a stock route from a couple of weeks ago on my 22 Civic ICE a2152effbad22b70/000004a5--2c90123137

Strange, it's always 1. I suppose it doesn't break anything, even if it differs from @royjr's specific Civic which we don't have a route for. Maybe he put it in a special mode/turned it off?

sshane:
@MarcoTheDingo Oh your route doesn't enable ACC from what I see. If you have one with stock ACC I can check out, we can merge

MarcoTheDingo:
@sshane here's a few examples, each of them bookmarked after "Stopped" showed on the dashboard.

- a2152effbad22b70/0000002a--686c35edd5/38 06:01:50 EST
- a2152effbad22b70/0000002b--79cf7696b7/36 15:59:30 EST
- a2152effbad22b70/0000002b--79cf7696b7/38 16:02:25 EST
- a2152effbad22b70/0000002b--79cf7696b7/48 16:12:00 EST

opendbc/car/honda/hondacan.py
MarcoTheDingo: @mvl-boston on the 22 Civic ICE the car does not go into idle stop for 4 seconds which is why we set it this way. Before the original PR that made this change, as soon as openpilot stopped, it would attempt to turn the engine off, which we don't want for quick stops.

royjr:
not sure if it helps, but I had installed an idlestopper which simulates the idle stop button press on car start to disable it so I never drove with op with idle stop enabled

davkg:
Hi, giving a bump here. This fixes braking on alpha long for 2025+ Civic hybrids.

Just want to emphasize this is a critical/functional fix, not a quality of life one. Currently with alpha long enabled, the Civic Hybrid will not brake at all, only accelerate or coast.

jrgiacone:
Following up on this one, I have instances on stock openpilot without this fix where comma is requesting more braking but the car is not responding. 2026 civic hatcbhack hybrid. Car will rear end without these fixes.

mhayden123:
@sshane Can we get an update on this please? Having this fix merged into stock openpilot would be HUGE

sshane:
It sounds like the behavior differs depending on model year. Can we find a bit from the camera or PCM that describes the difference?

mvl-boston:
I believe this is due to being needed with electronic braking assist.   Aligns with models that have Bosch electric brake boosters, or hybrids.    I'll revise to only trigger on those models and publish to my Honda forks which are widely used, and solicit feedback of any issues/regressions.

jrgiacone:
I can report, I pulled this fix onto stock and IQ pilot, and this fix allows for full braking authority at speeds on 26 civic hybrid, for now honda users need to be on a fork with the fix to command full braking, if this is incorporated into stock opendbc, then honda civic hybrid users will be able to switch to base openpilot if they choose.

mvl-boston:
> It sounds like the behavior differs depending on model year. Can we find a bit from the camera or PCM that describes the difference?

It has been about 10 days in my dev forks and the flag-based implementation has caused no issues for ICE or Hybrid drivers.  Seems good to go for me.

I also rebaselined to get the prior logic out of commit history.

And as a reminder: this is low risk as it only impacts dev branches since it is gated behind an alphalong toggle.

jrgiacone:
Any thing else need to be done on this to get it merged?

davkg:
@mvl-boston I think you forgot to update the PR, the diff is showing the old change

mvl-boston:
> @mvl-boston I think you forgot to update the PR, the diff is showing the old change

Not following your concern, just re-merged latest master but didn't see issues before nor after.  What update?

davkg:
> Not following your concern, just re-merged latest master but didn't see issues before nor after. What update?

Ah sorry disregard! The diff is updated, I misread the diff. I was referring to the hybrid/alt_brake flag, which is there 👍

PR #2060, opened 2025-03-28
Subaru Outback: adjust lateral accel factor

sshane:
Need to find the average across the fleet first

@fredyshox can you try this?

image

image


fredyshox:
Steer actuator delay should be 0.25, which will affect factor/offset too. I'm going to rerun torqued to verify this

PR #2067, opened 2025-03-30
Enable "Blind Spot Monitoring" for Mazda

FrogAi:
The blind spot signals are mapped here:

https://github.com/commaai/opendbc/blob/2ed580d9f58bfe40d73ab5def33ce628e3a85ca3/opendbc/car/mazda/carstate.py#L51

but they're not enabled. Not sure if there's a reason behind that or if it's merely an oversight.

Credit to @bigboigahoy.

### DIFF
--- a/opendbc/car/mazda/interface.py
+++ b/opendbc/car/mazda/interface.py
@@ -29,4 +29,6 @@ class CarInterface(CarInterfaceBase):
 
     ret.centerToFront = ret.wheelbase * 0.41
 
+    ret.enableBsm = True
+
     return ret


adeebshihadeh:
Why do we even have this? I think we should just remove it and not set the carState fields. @jyoung8607  @sshane any thoughts?

sshane:
I think it's for detection

adeebshihadeh:
I see that now, but it's pretty confusing having not seen any of this code in a while. For example, should I gate reading `carState.leftBlindspot` on `enableBsm` in openpilot?

It's similar to the `cruiseState.enabled` problem where that used to be read for openpilot's engagement state. The scope is really jumbled around with some of these fields, and we need to at least figure out a way to make this more obvious. `enableBsm` for example probably doesn't even need to leave the car port or even make it to carParams.

jyoung8607:
> but they're not enabled. Not sure if there's a reason behind that or if it's merely an oversight.

Not sure of that myself, and I was looking through the original PRs that added it earlier. The GEN1 flag is used in an odd way that makes it look optional, but all Mazda cars carry that flag, so we happen to always get though current Mazda carstate updates with no other condition check around BSM. Perhaps it's preparation for upstreaming another Mazda platform.

Think the right thing is to go ahead set enableBsm, as proposed, for consistency and clarity if nothing else. I don't know if all current supportable Mazda have it, but we sure act like they do, so nothing else technically needed at this time.

sshane:
This does nothing

PR #2069, opened 2025-04-01
Tesla: fingerprint Model 3 2025 (HW4)

lukasloetkolben:
6321083713becd53/00000003--b6c8cf2fd2


lukasloetkolben:
> If you're sure it's a MY 2025, go ahead and extend model-year support in CarDocs. The VIN query doesn't seem to work on Tesla so I can't check here.

Yes its a Tesla Model 3 2025. I updated the CarDocs

PR #1137, opened 2024-08-20
Hyundai: Car Port for Kona 2022

sunnyhaibin:
**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with stock system: `3bd70ed1a93771dc|2023-09-19--19-16-45`
- [x] route with openpilot: `0099bdb24d82951b/00000005--c38d940b04`
- [x] harness Type: [Hyundai O](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)
- [x] architecture: CAN

**Link to all physical parts during installation**
  - [comma 3X](https://comma.ai/shop/comma-3x)
  - [car harness kit](https://comma.ai/shop/car-harness)
    - [Hyundai O](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

Thanks to the community 2022 Hyundai Kona owner big_bull_ (Discord) and johnnathanwood.

**Requires**
- https://github.com/commaai/panda/pull/2038

Supersedes https://github.com/commaai/openpilot/pull/30025

PR #2072, opened 2025-04-01
Kona 2022 alt limit flag

johnnathanwood:
The 2022 Hyundai Kona requires more conservative steering torque limits compared to standard Hyundai models. This PR applies the ALT_LIMITS_2 flag 

dongle id: [0099bdb24d82951b]


sunnyhaibin:
This change was somehow left out during the port we merged at COMMA_HACK_5. https://github.com/commaai/opendbc/pull/1137

@adeebshihadeh @jyoung8607 this car is behind dashcam mode, but still needs this flag to actually limit the torque by half if anyone decides to remove the dashcam mode.

sshane:
This is exactly why we can't continue with alt limit flags

johnnathanwood:
> This is exactly why we can't continue with alt limit flags


Hey @sshane, I understand the concerns with alt limit flags. Out of curiosity, is there a way we could make torque limits dynamic? For example, applying more conservative torque at highway speeds but allowing higher torque on surface streets where tighter turns are more common? Just thinking out loud

sshane:
@adeebshihadeh these cars are not dashcammed due to the line below...

```
ret.dashcamOnly = candidate in {CAR.KIA_OPTIMA_H, }
```

PR #2075, opened 2025-04-03
docs: add setup note

incognitojam:
Add an example setup note to be shown on the website setup guide

It looks like we don't use `shop_footnote` anywhere anymore, so I re-used it for this (I might have missed somewhere?)
image

In the future if we want to support more possibilities we could use the column attr to define a SetupColumn and attach footnotes to different parts of the setup guide


incognitojam:
Probably need to make sure this footnote doesn't appear in the openpilot docs

PR #2076, opened 2025-04-03
No-op changes to make Panda happy

aubsw:
No-op change for https://github.com/commaai/panda/issues/2171

- Moves some functions and declarations around so that they can be correctly included in Panda headers.

## Details

In [panda](https://github.com/commaai/panda), we currently need some defintions from [safety.h](https://github.com/commaai/opendbc/blob/master/opendbc/safety/safety.h). But we can't include safety.h directly because it contains function implementatinos (we eventually should link the implementations at compile time). 

So in this PR, we're just safely moving code that will allow us to include headers from opendbc without compiler errors.

- Panda needs access to `SAFETY_ALLOUTPUT`. But we can't directly include safety.h anywhere in panda because safety.h contains implementations (can't link multiple objects files that contain definitions for the same functions). 
  - Hence, we move the definition of `SAFETY_ALLOUTPUT` to safety_declarations.h, which we can then include in panda source.
- Move forward declarations of `generic_rx_checks` and `stock_ecu_check` from safety_declarations.h to safety.h (since can't include forward-declared static functions without their corresponding implementation)
- make `get_ts_elapsed` static inline! This way we can safely include utils.h without compiler errors for multiple definitions.

Further discussion: https://github.com/commaai/panda/issues/2171#issuecomment-2784918381

### DIFF
--- a/opendbc/safety/board/drivers/can_common.h
+++ b/opendbc/safety/board/drivers/can_common.h
@@ -1,3 +1,4 @@
+#pragma once
 #include "can_common_declarations.h"
 
 uint8_t calculate_checksum(const uint8_t *dat, uint32_t len) {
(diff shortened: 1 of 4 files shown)


robbederks:
LGTM, @sshane any comments?

PR #2080, opened 2025-04-07
Honda: Cleanup Nidec steering dbc

vanillagorillaa:
Made a separate PR from my other cleanup since this one is more critical to openpilot. 

Nidec uses two steering messages so made two common dbc files to import. 

The `honda_odyssey_extreme_edition_2018_china_can` dbc has the `STEER_TORQUE` signal as 16 bits, when it is 12 on the rest. Not sure if this is just by accident or not. Don't have a way to validate so left it as is.


adeebshihadeh:
Nice cleanup!

PR #2085, opened 2025-04-09
FingerPrint F-150

Yeek2:
Adding F-150 ECU (2022)
Route: 790ca9ce86fae049/00000002--53a2aafecf


incognitojam (approved):
LGTM

sshane:
Why didn't this fuzzy fingerprint?

incognitojam:
The `NL34` prefix was outside the known range for the F-150

incognitojam:
The fuzzy fingerprinting can be even fuzzier if we are just trying to detect the model (and we probably could be querying the hardware part no instead of software). At the moment it will exclude software versions newer or older than the newest and oldest in the table.

PR #2086, opened 2025-04-09
safety: replace fwd hooks with relay addr check config

sshane:
We're in a tricky spot with some safety modes, such as Tesla which need these to be dynamic, but we stopped giving the safety mode these messages if they're not in the rx checks anymore


sshane:
Done in https://github.com/commaai/opendbc/pull/2095

PR #2101, opened 2025-04-10
Move static forward declaration

aubsw:
For https://github.com/commaai/panda/issues/2171.

We are in an unfortunate situation right now where we are trying to migrate the panda repo to source/header style. This is fundamentally incompatible with forward-declared static functions that were implemented in  https://github.com/commaai/opendbc/pull/2088 (see https://github.com/commaai/opendbc/pull/2088/files#diff-99ed3acebc587c8e2e9a0ffd69d9c8e87c62183fcdf7e6bac0286acc5e30eeb8)

This PR is a temporary hack to help us land https://github.com/commaai/panda/issues/2171 without major changes to opendbc.

This should be a no-op.

### DIFF
--- a/opendbc/safety/safety.h
+++ b/opendbc/safety/safety.h
@@ -114,6 +114,9 @@ uint16_t current_safety_param = 0;
 static const safety_hooks *current_hooks = &nooutput_hooks;
 safety_config current_safety_config;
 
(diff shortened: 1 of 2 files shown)


aubsw:
implemented in https://github.com/commaai/opendbc/pull/2076

PR #2103, opened 2025-04-10
safety: move Honda to msg block config

sshane:
TODO:
- [x] verify it's okay to forward camera messages to radar with Bosch long (since we're sending on bus 1, the forwarding logic doesn't block any of these from reaching bus 0/the radar but it should be okay since it's disabled) - https://connect.comma.ai/84ec13cd18bc76ca/000000ff--0eb28a38b6


sshane:
scared

PR #2104, opened 2025-04-10
Honda Bosch: Steer to zero

vanillagorillaa:
Verification - 178b5f3e0f647798/00000000--a34caccddf


sshane:
Nice, how does this work?

vanillagorillaa:
Just have verbal confirmation from test user that this works. Waiting for a route and logs for confirmation.

sshane:
I mean how does the bit work?

vanillagorillaa:
@sshane Just added a route. This works by setting the "STEER_TO_ZERO" bit previously found by greg after dumping the EPS firmware (https://github.com/commaai/opendbc/pull/317). The EPS `STEER_STATUS` message enters into a state of `1`, which is a previously unseen, unknown state. The EPS seems to still accept commands in this state.

PR #2115, opened 2025-04-12
Tesla: raised lateral limits

sshane:
Right now, without vehicle model's curvature factor that assumes we get less curvature than expected at higher speeds, we get 60% less lateral accel than we could at 80 m/s. This graph converts 3 m/s^2 accel to curvature without curvature factor, then uses the vehicle model to convert to lateral accel again with curvature factor:

image


opendbc/car/tesla/carcontroller.py
```diff
@@ -1,9 +1,57 @@
...
+from opendbc.car import ACCELERATION_DUE_TO_GRAVITY, Bus, AngleSteeringLimits, DT_CTRL, rate_limit
+from opendbc.car.interfaces import CarControllerBase, ISO_LATERAL_ACCEL
 from opendbc.car.tesla.teslacan import TeslaCAN
 from opendbc.car.tesla.values import CarControllerParams
+from opendbc.car.vehicle_model import VehicleModel
+
+MAX_ANGLE_RATE = 10  # deg/20ms frame, EPS faults at 12 at a standstill
+
+# Add extra tolerance for average banked road since safety doesn't have the roll
+AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
```
greatgitsby, line 13: here it says `higher actual roll lowers lateral acceleration` but ford comment says `higher actual roll raises lateral acceleration`

sshane: Yes that's correct curvature and angle behave differently

riderx:
@sshane 3 m/s^2 is what Tesla has implemented in EU.
It's not great; Tesla cars do give up in many normal EU corners.
I was thinking of getting a coma device to bypass this stupid "comfort norm" 
Currently my car disengages in turn like here: https://maps.app.goo.gl/35J159g5sGKFecPN8
On the highway, the limit of the road is 110km/h who everyone takes this turn around 100/110.
Often my car on FSD(EU version) tries to slow down to 60km/h .who is already dangerous on its own, and disengages midcorner.
I do understand following convention can be great but this one should not or be allowed to be disabled, as it's a "comfort norm"

PR #2117, opened 2025-04-12
Honda: CANFD CRV

vanillagorillaa:


vanillagorillaa:
Previous PR with verification route - https://github.com/commaai/openpilot/pull/32806

vanillagorillaa:
Rebased off of master after Pilot merge

opendbc/car/honda/carstate.py
```diff
@@ -92,6 +92,8 @@ def __init__(self, CP):
       self.gearbox_msg = "GEARBOX_ALT"
     elif CP.transmissionType == TransmissionType.manual:
       self.gearbox_msg = "GEARBOX_ALT_2"
+    elif CP.carFingerprint == CAR.HONDA_CRV_6G:
+      self.gearbox_msg = "GEARBOX_15T"
```
sshane, line 96: Move up with Accord check?

vanillagorillaa: We could, but then we would also need to add the check in interface to assign the CVT transmission type

vanillagorillaa:
Will rebase after https://github.com/commaai/opendbc/pull/2141

PR #2121, opened 2025-04-13
Honda: abstract out RADAR_HUD msg definition

vanillagorillaa:
Found that `RADAR_HUD` had some inconsistencies. 

`zeros_boh` overlaps into `apply_brakes_for_canc`

`lead_distance` is actually 1 byte, and `lead_state` was not a real signal. 

Verification 5cfa1ad8f35779d0|2023-07-03--09-17-39


adeebshihadeh (changes requested):
Can we abstract this out and import it?

PR #2126, opened 2025-04-14
Rivian: alternative Ignition Signal

lukasloetkolben:


sshane:
Know the conditions for this one? Same as the EPAS one?

lukasloetkolben:
> Know the conditions for this one? Same as the EPAS one?

Yes, the signal is matching the `VDM_EpasPowerMode` we are already using.

Issue #1100, opened 2023-11-30
Honda: check regen paddle


plygrnd:
## TL;DR

Honda Insight's brake regen adjustment paddles cause OP to throw a CruiseDisabled error instead of canceling cruise control.

## More Info

The Honda Insight has adjustable regenerative braking. This is controlled by paddles on the steering wheel. Hitting either paddle when cruise is engaged will disengage cruise. 

Disengaging cruise control in this fashion causes OP to trigger a "cruise not engaged" message and sound the alarm, instead of gracefully exiting cruise control via the "Cancel" button. This is governed in the Honda interface, [L332](https://github.com/commaai/openpilot/blob/a2bb41e0ec5fd96763e9a01a7116efda059de6e6/selfdrive/car/honda/interface.py#L332).  

It appears that the car sends a generic "Cruise Control disabled" message; there's a lot of junk in Cabana for this trip which isn't mapped to any signals, but one of those could be the issue.

### Desired Outcome

When using the steering wheel paddles to disable cruise control, OP should treat this as a standard disablement instead of an emergency.

### Which car does this affect?

Honda Insight EX 2019

### Provide a route where the issue occurs

8958942b212e8cd9|2023-11-30--08-18-06--10

### openpilot version

On `release3`, latest release.

vanillagorillaa:
Hi @plygrnd I think I have resolved this issue. If you are on the comma discord do you mind sending me a DM? I have the same username

jyoung8607:
For benefit of anyone that's experiencing this bug, looks like @vanillagorillaa has a good start on the issue in #2229 but we're blocked looking for an Insight, Clarity, or other test user who has regen paddles that are currently causing problems with openpilot disengagement. If you're experiencing this issue, please share a link to a route in this issue.

vanillagorillaa:
FWIW, i'm no longer working/contributing on/to this issue. Good luck to however decides to pick it up

jyoung8607:
Potentially connected, the Insight DBC has a weird difference in its STANDSTILL message which I suspect is a bug. If it's a bug, it might falsely trigger `accFaulted` under openpilot longitudinal. This could explain difficulties fixing the regen paddle.

jyoung8607:
Notes:

* The stock ACC case will be resolved by #2624.
* The openpilot longitudinal case will require some safety code refactoring.
* Back out commaai/openpilot#35929 before closing.

mvl-boston:
[...]

PR #2129, opened 2025-04-15
Honda: Acura MDX 2025 (canfd)

mvl-boston:
PR to support 2025 Acura MDX

- Supports OP steering/lateral control at all speeds
    -Limited torque as typical for Honda models
- Uses stock Acura long control, works at all speeds, includes stop/go support
- OP long control is in separate [PR2162](https://github.com/commaai/opendbc/pull/2162)
- renamed pilot 2023 dbc to honda_canfd_common since all canfd cars so far are identical
- changes consistent with canfd CRV https://github.com/commaai/opendbc/pull/2475

Code variable definitions:
- honda_canfd - new generation of radar using Honda Bosch C with common code changes required across multiple in-progress Honda/Acura cars.
- 4G MMR - standard Honda-speak for 4th generation (2022+ MDX) mid-model-refresh (Changed to CANFD in 2025)

Test Hardware: Comma 3x (comma power not used)
Test Harness: Comma Honda Bosch C
Test Route: ad9840558640c31d/0000001a--d6cd4871c2

Dashcam mode for steer calibration: 1a7007d141712ce6/00000002--46a5706ebc
Newer testroute if preferred: 1a7007d141712ce6/00000000--c7613dc482

Using Comma standard mount, calibration is 4.7 degrees down.   Since the middle of the acceptable range is 2 degrees down, I do not feel this needs the 8-degree mount like the Honda Pilot did.
Calibration route: 1a7007d141712ce6/00000002--c4eac90703 thanks to gregg

Key events in test route:
7:16am - stop go
7:18am - steer max reached
7:21am - successful door open test

Adding the following logs at sshane's request in discord:
- Holding the 3x against the windshield" log: ad9840558640c31d/00000008--84c6baca80
- New route for the PR after sshane's edits: "ad9840558640c31d/0000000a--505c644d4b".   Stop and go works @ 2:26..

Huge thanks to vanillagorillaa for support and for the base dbc work on the 2023 honda pilot.


opendbc/car/honda/values.py
```diff
@@ -216,6 +216,12 @@ class CAR(Platforms):
     {Bus.pt: 'honda_pilot_2023_can_generated'},
     flags=HondaFlags.BOSCH_CANFD | HondaFlags.BOSCH_ALT_BRAKE,
   )
+  ACURA_MDX_4G_MMR = HondaBoschPlatformConfig(
+    [HondaCarDocs("Honda Pilot 2023", "All")],
```
sshane, line 220: can you fix the docs?

sshane:
What does MMR mean?

opendbc/car/honda/fingerprints.py
```diff
@@ -494,6 +494,14 @@
       b'36161-TDK-J530\x00\x00',
     ],
   },
+  CAR.ACURA_MDX_4G_MMR: {
```
sshane, line 497:
```suggestion
  # mid model refresh
  CAR.ACURA_MDX_4G_MMR: {
```

sshane:
FYI whatever mount you're using does not look good. It has significant play compared to the hood from what I can see in your test route.

opendbc/car/honda/values.py
sshane, line 220:
```suggestion
    [HondaCarDocs("Acura MDX 2025", "All")],
```

mvl-boston:
Merged latest master.   Website merge corrupted things with safety folder rename.   Force-pushed to revert and did a command line merge instead.

mvl-boston:
cleaned up to align with changes in https://github.com/commaai/opendbc/pull/2475

opendbc/safety/tests/test_honda.py
adeebshihadeh, line 1: this mess is the next thing on the chopping block

mvl-boston: Let me know how I can help.  Canfd had different tx signals/bus numbers which was the need for this breakout.

adeebshihadeh: nah, this is a more general refactor that we have to do. it's not particular to your port

mvl-boston: OK, am going to use this version to baseline a canfd long control pr that jyoung will roadtest next week.   I'll keep an eye on it and modify my pr along whatever cleanup themes you refactor into baseline.

jyoung8607:
- [x] Get a dashcam route with strong factory LKA actuation to find known good actuation range
- [x] Convert to lateral accel torque controller
- [x] Refine params with test drives
- [x] Update CI test routes

mvl-boston:
I think this test route already has what you need.   Can you advise if sufficient?

> * [ ]  Get a dashcam route with strong factory LKA actuation to find known good actuation range
> 
- per original comments, max LKA actuation is observed at 7:18am of existing test route

>     * [ ]  Convert to lateral accel torque controller
> 
- done using Accord as a base

>     * [ ]  Refine params with test drives
> 
- already done by shane based on existing test route.
- 4096 software steer control was fully responsive, so car likely could have responded to more.   If I can find a tester to prove higher I will update.
Only question is steering ratio, I used manufacturer published spec instead of observed from log (and shane didn't modify).  You revised Accord to change to observed from log, should I do the same here?

>     * [ ]  Update CI test routes
- is that still needed or was existing sufficient?
-

jyoung8607:
> I think this test route already has what you need. Can you advise if sufficient?

`1a7007d141712ce6/00000002--c4eac90703` is indeed a dashcam route, but rlogs aren't uploaded so I don't really have CAN traffic to look at.

> * done using Accord as a base - only question is max of steer actuation based on 7:18am.    I can download steer control vs lataccel into excel and manually calc max, but is there already an existing plotjuggler graph I can use to detect more easily?

That's really two questions:

For the factory LKA control/actuator limit, just use cabana (or PlotJuggler with the --can option) and look for an obvious plateau when you're maxing it out. From the Accord:

image

For the corresponding lateral accel factor, I just let `torqued` figure it out while driving. It's always listening and learning, we just weren't using its learned value before.  We want to allow it to learn online anyway, so we should get close to the value it's going to learn, so driving behavior is consistent from first startup.

The issue here, during development, is trusting it for values beyond what the camera sends. The entire Honda port is full of `# TODO: determine if there is a dead zone at the top end` and there do, in fact, seem to be deadzones at the top. 

`torqued` does its job by looking at what percentage of available actuator you needed to reach a given lat accel. Using the Accord as an example, it thinks your lat accel factor is a lot higher for a given turn when we're sending 500 out of range 0-4096 than we're sending 500 out of range 0-2560. It has no idea if there's a deadzone at the top.

> * already done by shane based on existing test route.   Only question is steering ratio, I used manufacturer published spec instead of observed from log (and shane didn't modify).  You revised Accord to change to observed from log, should I do the same here?

Yes, we'll want to update the steering ratio with something close to the online learned values. We'll also want to check the liveDelay estimate and use that (minus 0.2) to set the steerActuatorDelay. I've been removing the artificially low tire stiffness parameters. They were a [workaround for older steering racks](https://github.com/commaai/opendbc/blob/53fdac763f9880631b809b7b80176d5660fba58b/opendbc/car/honda/interface.py#L82-L84) with top-of-rack steering angle sensors.

Once we identify and match the factory camera maximum, the existing lateral accel params are probably a good enough guess to start driving and learning. It's going to wander around a bit depending on road conditions, what speed you're driving, etc. More history and more varied driving conditions over time will improve its estimate, and then we can update the params YAML with a better idea of how the car performs.

None of this is super exact, we just need to be close enough to give people a good fresh-install first-drive experience.

> * is that still needed or was existing sufficient?

It's acceptable if we have to, but if convenient, I'd rather have a fresh one with actuator limits and lateral control method matching what we merge with, and with `lagd` online actuator delay estimates that openpilot recently began supporting. Make sure the owner knows the test route will be public, so it's a good idea to start and stop the route somewhere other than one's home or workplace or anything else privacy-relevant.

mvl-boston:
I was referring to the test route in the software: ad9840558640c31d/0000001a--d6cd4871c2.

Not sure where your dashcam route came from?

I was able to figure out and pull/update all numbers from the software test route.   Only issue is that steer_torque_sensor and angle_rate responded at the max steer_control of 4096 (no deadzone) so it would be nice if one of the testdrivers who still own this car could try higher steer_control.   Both donuts and your dashcam suggestion should work for this.

Beyond possibly increasing the 4096 max I think this PR is ready.   Let me know if anything I missed.

jyoung8607:
> I was referring to the test route in the software: ad9840558640c31d/0000001a--d6cd4871c2.

That route has openpilot doing the driving. It's sending actuation commands up to 4096 because that's what we told it to do. We need one in dashcam mode to see what the factory camera would do instead.

> Not sure where your dashcam route came from?

It was the "calibration" route right above the time references you talked about, I guess I misunderstood. Either way, we do still need an uploaded dashcam route with some strong LKA interventions.

> I was able to figure out and pull/update all numbers from the software test route.

Those values aren't very trusthworthy yet due to reasons I outlined in my prior comment.

jyoung8607:
BTW, those TODO items were really for me. I'd certainly be grateful if you feel like doing the legwork, but if not, I'm happy to do it. All I need is a cooperative tester that can produce a dashcam test route and then install a testing branch.

jyoung8607:
> * [x]  Get a dashcam route with strong factory LKA actuation to find known good actuation range

This car also peaks at 2560: `1a7007d141712ce6/00000002--46a5706ebc`

image

mvl-boston:
> BTW, those TODO items were really for me. I'd certainly be grateful if you feel like doing the legwork, but if not, I'm happy to do it. All I need is a cooperative tester that can produce a dashcam test route and then install a testing branch.

Thanks, I misread.  Yes I have a tester doing dashcam LKAS this morning and will pass on the log once received.


> > * [x]   Get a dashcam route with strong factory LKA actuation to find known good actuation range
> 
> This car also peaks at 2560: `1a7007d141712ce6/00000002--46a5706ebc`

Strange - I was using angle m/s as my benchmark and it kept moving all the way to 4096 (in .ad9840558640c31d/0000001a--d6cd4871c2).  Is steer_torque trimmed or maybe missing a bit in the DBC?

Edit: if above was from a dashcam, I'm suspecting that factory LKAS isn't commanding as high as EPS allows.  Pilot tester said that dashcam route didn't steer as well my community fork route, and Pilot/MDX are Honda/Acura twins.

Honda markets LKAS as a nudge aide, but expects you to actively steer the entire time.   So it would be consistent with LKAS shutting off before EPS max.

mvl-boston:
Dashcam mode for steer calibration: 1a7007d141712ce6/00000002--46a5706ebc
Newer testroute if preferred: 1a7007d141712ce6/00000000--c7613dc482

jyoung8607:
LGTM. Notes for merge:

* There is truly only one model-year of support, MY2025
* The fingerprint list was recreated from two known dongles, notes in commit history/git blame
  * Would have preferred a bit more data here, worth another glance once there's more testers
  * #2632
  * ~~There's a hybrid variant, so #1100~~

opendbc/car/honda/fingerprints.py
jyoung8607: @sshane this gets discarded by tools/car_porting/auto_fingerprint, moved it to cardocs in #2654.

mvl-boston:
> * There is truly only one model-year of support, MY2025
2026 is on sale and likely a clone of 2025.   Just waiting an owner to give it a try to test/fingerprint.



>   * There's a hybrid variant, so #1100
There is no Hybrid for the canfd MDX.  Acura paused MDX hybrids after 2020.

Issue #1834, opened 2025-02-24
Move CAN ignition hooks and add test


adeebshihadeh:
CAN ignition lives in [board/drivers/can_common.h](https://github.com/commaai/panda/blob/master/board/drivers/can_common.h). This is the last car-specific code in panda and should be moved here.

It should also be possible to fully validate these signals in CI. Assert that the hook doesn't unintentionally trip on all the other cars, etc.

sshane:
with proper rx checks (counter, checksum, etc.) to reduce chances of overlap between cars

adeebshihadeh:
> with proper rx checks (counter, checksum, etc.) to reduce chances of overlap between cars

Let's do this as a separate issue once this is in.

sshane:
@ccdunder want to tackle this one?

aubsw:
@sshane I humbly propose https://github.com/commaai/opendbc/pull/2131

Lemme know what you think!

aubsw:
I've been [experimenting](https://github.com/commaai/openpilot/pull/35082) with adding coverage to test_models.py

We actually have some decisions we need to make here:

My first thought was to check every rx from every log to ensure that only [gm|mazda|rivian|tesla] see CAN-ignition. i.e.:

```diff
diff --git a/selfdrive/car/tests/test_models.py b/selfdrive/car/tests/test_models.py
index 53f2bb586..48b33b42b 100644
--- a/selfdrive/car/tests/test_models.py
+++ b/selfdrive/car/tests/test_models.py
@@ -243,6 +243,11 @@ class TestCarModelBase(unittest.TestCase):
         if self.safety.safety_rx_hook(to_send) != 1:
           failed_addrs[hex(msg.address)] += 1

+        if self.CP.brand in ("gm", "mazda", "rivian", "tesla"):
+          self.assertTrue(self.safety.get_ignition_can())
+        else:
+          self.assertFalse(self.safety.get_ignition_can())
+
       # ensure all msgs defined in the addr checks are valid
       self.safety.safety_tick_current_safety_config()
       if t > 1e6:
```

https://github.com/commaai/openpilot/blob/c9c6eaa725ed9f2a3c0dd98d844979ccdf358fea/selfdrive/car/tests/test_models.py#L237-L249

But there are two problems with this:

1. `assertFalse` on every rx blows up the test run time from 6 mins to 15 mins
2. Not every single message from the [gm|mazda|rivian|tesla] logs have ignition_can == True 
[...]

PR #2131, opened 2025-04-15
Port CAN ignition detection from panda

aubsw:
For https://github.com/commaai/opendbc/issues/1834

- Introduce a `ignition_can` safety hook
- Port CAN ignition detection logic from https://github.com/commaai/panda/blob/master/board/drivers/can_common.h#L163-L212 for tesla, rivian, mazda, and GM
- Add unit tests to ensure that CAN message sequences don't trigger unexpected behaviors


opendbc/safety/safety.h
```diff
@@ -209,6 +212,10 @@ bool safety_rx_hook(const CANPacket_t *to_push) {
     heartbeat_engaged_mismatches = 0;
   }
 
+  if (current_hooks->ignition_can != NULL && GET_BUS(to_push) == 0) {
+    current_hooks->ignition_can(to_push);
+  }
+
```
sshane, line 218: I thought about this, but the problem is that the current hooks are undefined until we set a safety mode (after we go onroad and fingerprint). You'll need to iterate through each safety mode's ignition hook

aubsw: Kk, looking into it. 

Overall, are you happy with the approach?

sshane: I think we're fine with single function as we used to use, just moving it is appropriate for the bounty.

aubsw: Got it, will revert to single function.

> the problem is that the current hooks are undefined until we set a safety mode

Got it. 

Are there any particular types of state you want me to check for in the tests aside from ensuring `ignition_can` is set to `true` when the right can messages get sent?

I was originally assuming we **do not** want to set `ignition_can` to `true` unless we're in the corresponding safety mode. But clearly that was not the old behavior, anyway.

aubsw:
Moving to draft for now. Working on sister PR in openpilot.

PR #2138, opened 2025-04-19
Nissan: Add multiple messages and multiple signals to existing messages, add L/R blind spot to carState

gncnpk:


adeebshihadeh:
Please do not use any kind of LLM or auto-summarizer. Write a good description or just leave it blank.

gncnpk:
> Please do not use any kind of LLM or auto-summarizer. Write a good description or just leave it blank.

Sounds good, I'll leave it blank!

PR #2139, opened 2025-04-20
Mazda: Add FW for 2025 Mazda CX-5

valdrox:
fingerprint for Mazda CX-5 2025. 

Tested, works as expected.

Two of the values looked a bit out of place. Please let me know if I need them as well!


adeebshihadeh:
Got a dongle ID / route?

valdrox:
dongle ID : 1a095eb4bc56e261

jyoung8607:
Looks like just one real bit of missing firmware, I think the rest came from other-platform queries before the car was properly identified. Merging after cleanup.

valdrox:
@jyoung8607 thanks i'll try and give the master branch a go this weekend. :)

PR #2140, opened 2025-04-20
Honda: Clean up carstate

vanillagorillaa:
I'm not even finding where this code is used. And git blame seems to go back to very early days of openpilot and seems to have been added without explanation?

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -62,11 +62,6 @@ def get_can_messages(CP, gearbox_msg):
         ("ACC_HUD", 10),
         ("ACC_CONTROL", 50),
       ]
-  else:  # Nidec signals
-    if CP.carFingerprint == CAR.HONDA_ODYSSEY_CHN:
-      messages.append(("CRUISE_PARAMS", 10))
-    else:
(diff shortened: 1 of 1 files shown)


mvl-boston:
Agree, use was removed in [pr 22181](https://github.com/commaai/openpilot/pull/22181).   And they forgot to remove the signal.

opendbc/car/honda/carstate.py
```diff
@@ -62,11 +62,6 @@ def get_can_messages(CP, gearbox_msg):
         ("ACC_HUD", 10),
         ("ACC_CONTROL", 50),
       ]
-  else:  # Nidec signals
-    if CP.carFingerprint == CAR.HONDA_ODYSSEY_CHN:
-      messages.append(("CRUISE_PARAMS", 10))
-    else:
-      messages.append(("CRUISE_PARAMS", 50))
```
sshane, line 69: Sometimes signals are added just to do CAN validity checks on them, but I'm not sure if this is important

adeebshihadeh: it was used for a "cruise speed offset" (found in commaai/one)

PR #2142, opened 2025-04-22
Bolt 70kW Regen Signal

firestar5683:
When paired with EBCMRegenPaddle, State "5" unlocks the full 70kW regen available to the chevy bolt (State 7 is for gen1 bolts, state 5 is gen2). Static bytes must also be defined and sent for car to accept the signal. 

I'm using it to get access to full regen with a pedal interceptor, but it may also be an applicable method for regen blending on GM acc enabled vehicles. Still testing. 
Screenshot 2025-03-29 at 9 20 44 PM


firestar5683:
To clarify "all bytes must be defined" -
When spoofing the message to the car, this must be used. 
Screenshot 2025-04-21 at 10 46 46 PM

Since I haven't proven it enables blended regen on acc enabled vehicles yet, I wasn't sure if you would want the placeholder bytes. We're successfully in final tuning for non-acc vehicles

PR #2155, opened 2025-04-24
Perodua Myvi AV 2022+

arishwin:


adeebshihadeh:
Cool! Got a harness schematic? Any YouTube videos?

PR #2164, opened 2025-04-29
Honda: add directional LDW for Bosch

mvl-boston:
- Adding directional LDW when driver is steering outside of lane (eg: with OP off) - reported by Greggg
- Directional LDW means there is a separate warning for left late LDW vs right lane LDW
- Provide baseline for future PR to add steer warning and directional LDW when Honda steermax is reached, rather than when about to leave lane

Todo:
- initial drive testing

Routes:
- confirming left/right DBC: ad9840558640c31d/00000009--7b5b808ebf
- confirming functionality test: route TBD


mvl-boston:
Closing.   Will reopen when I have a test car for this.

PR #2165, opened 2025-04-29
Honda Bosch - longitudinal tuning

mvl-boston:
Implemented missing aspects of Nidec Honda tuning into Bosch:
- Longitudinal integral
- Wind brake

Now that TR works much better with radarless cars (radar is diabled when present for Honda Bosch, this lack of tuning is more noticeable.  Massive reduction in disengagements.

Todo: need to upload before/after routes:


jyoung8607:
Parking this for similar reasons as #2347. Wind resistance will be part of the drivetrain torque calculation. However, we can go ahead and bring in the pitch compensation piece since that offsets acceleration.

PR #2168, opened 2025-04-30
Honda: Cleanup carstate doors

mvl-boston:
Change is logically equivalent to prior door status code.
- Prior code separately enumerated every single Bosch car except HONDA_CRV_5G
- Revised code explicitly says "Bosch except HONDA_CRV_5G"
- HONDA_CRV_5G exclusion was deliberate, with a separate dbc with separate doors signals

This change prevents need for new Bosch A car ports to add to a growing list of individually enumerated cars, including.
-[PR1967](https://github.com/commaai/opendbc/pull/1967) for 2022 Acura RDX 
-[PR2170](https://github.com/commaai/opendbc/pull/2170) for 2021-2025 Honda Odyssey
-I am actively soliciting testers for missing model years for Bosch A, including 6th gen TLX pre-MMR, and 4th gen MDX pre-MMR.  This code simplification would help those years too.


opendbc/car/honda/carstate.py
```diff
@@ -68,9 +67,7 @@ def get_can_messages(CP, gearbox_msg):
     else:
       messages.append(("CRUISE_PARAMS", 50))
 
-  if CP.carFingerprint not in (CAR.HONDA_ACCORD, CAR.HONDA_CIVIC_BOSCH, CAR.HONDA_CIVIC_BOSCH_DIESEL, CAR.HONDA_CRV_HYBRID, CAR.HONDA_INSIGHT,
-                               CAR.ACURA_RDX_3G, CAR.HONDA_E, CAR.HONDA_ODYSSEY_CHN, CAR.HONDA_FREED, CAR.HONDA_HRV, *HONDA_BOSCH_RADARLESS,
-                               *HONDA_BOSCH_CANFD):
+  if CP.carFingerprint == CAR.HONDA_CRV_5G or CP.carFingerprint not in (CAR.HONDA_ODYSSEY_CHN, CAR.HONDA_FREED, CAR.HONDA_HRV, *HONDA_BOSCH):
```
vanillagorillaa, line 70: I'm not sure this is correct. HONDA_CRV_5G is a HONDA_BOSCH Car, so wouldn't this not make sense?

opendbc/car/honda/carstate.py
```diff
@@ -136,8 +133,7 @@ def update(self, can_parsers) -> structs.CarState:
     # panda checks if the signal is non-zero
     ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 1e-5
     # TODO: find a common signal across all cars
-    if self.CP.carFingerprint in (CAR.HONDA_ACCORD, CAR.HONDA_CIVIC_BOSCH, CAR.HONDA_CIVIC_BOSCH_DIESEL, CAR.HONDA_CRV_HYBRID, CAR.HONDA_INSIGHT,
-                                  CAR.ACURA_RDX_3G, CAR.HONDA_E, *HONDA_BOSCH_RADARLESS, *HONDA_BOSCH_CANFD):
+    if (self.CP.carFingerprint in HONDA_BOSCH) and not (self.CP.carFingerprint == CAR.HONDA_CRV_5G):
```
vanillagorillaa, line 136: Same as above

opendbc/car/honda/carstate.py
mvl-boston: They have a special dbc that has non-Bosch door signals.  Don't know what Honda was thinking.

opendbc/car/honda/carstate.py
mvl-boston: See above reply

adeebshihadeh:
This is still a huge mess that's hard to review. Can we find some new signals that are ideally common to all Honda or even just Honda Bosch? 

Our CAN dataset should make this easy: https://huggingface.co/datasets/commaai/commaCarSegments.

mvl-boston:
Honda Bosch is 100% clean except for 1 car.   I'll check huggingface for the 1 exception car, but it will be unlikely to have an open door in a route to validate.

PR #2175, opened 2025-05-02
Tesla: move steering disengage to generic param

sshane:


opendbc/car/tesla/carcontroller.py
```diff
@@ -17,33 +17,27 @@ def update(self, CC, CS, now_nanos):
     actuators = CC.actuators
     can_sends = []
 
-    # Disengage and allow for user override on high torque inputs
-    # TODO: move this to a generic disengageRequested carState field and set CC.cruiseControl.cancel based on it
-    hands_on_fault = CS.hands_on_level >= 3
-    cruise_cancel = CC.cruiseControl.cancel or hands_on_fault
-    lat_active = CC.latActive and not hands_on_fault
-
```
sshane, line 25: This change makes since as now reaching a level 3 override is a latching disengage. Previously, if the car somehow ignored our cancel message and the user was super quick, we could stay enabled.

PR #2179, opened 2025-05-03
draft Honda: Bosch gasonly PID

mvl-boston:
(moving to draft, per [openpilot PR 35375](https://github.com/commaai/openpilot/pull/35375) this needs to be moved to a carcontroller PID)

Honda Bosch has a built in PID tuner for brake, but does not have one for gas.

Even after adjusting gas for detectable conditions (incline, speed-based drag), these calculations are always bit off for undetectable conditions like tailwinds, weather/air density, etc.   Thus a gas PID tuner is helpful for Honda Bosch.

However, the current openpilot PID tuner is only on overall accel.   Any PID adjustment on gas then applies to brake, doubling up with the builtin Honda brake PID, and causing undesired braking strength until openpilot's PID adjustment winds down.   Once it winds down, then you get undesired gas strength until it winds up again.

This pr allows for the PID tuner to only impact gas on Honda Bosch models.

Requires companion [openpilot PR 35375](https://github.com/commaai/openpilot/pull/35375) to send the pre-PID target acceleration in Actuators.

Test route (all tests also included [pr 2165](https://github.com/commaai/opendbc/pull/2165) to enable Bosch PID):
- Bosch C (2023 Honda Accord Hybrid): 15bd03c45fb5f401/0000000a--1805a6553f  (test included [PR 2191](https://github.com/commaai/opendbc/pull/2191))
- Bosch C (2023 Honda Pilot): 9662a1dab70c9dc3/00000032--d0dac7d95b
- Bosch C (2025 Acura MDX): ad9840558640c31d/00000013--893d36d514
- more coming from other cars in discord

Discord feedback:
- I drove for five hours, and it was an incredible experience."
- "worked well on the way to work today"
- "Worked way better than the 0.9.8 branch I had last been using. Actually useable this time around. "
- "once we can fix the rapid acceleration issue I think I’d daily op long" (retest in process after accel fix in [pr2185](https://github.com/commaai/opendbc/pull/2185))
- "Acceleration and braking were exactly as they should be. I drove over to a construction zone to see how it would handle sudden starts/stops, and it was perfect." (test included [PR 2191](https://github.com/commaai/opendbc/pull/2191))

Positive side effect:
- Bosch cars now stop at most stoplights in experimental mode.


opendbc/car/honda/carcontroller.py
```diff
@@ -206,11 +208,13 @@ def update(self, CC, CS, now_nanos):
...
         if self.CP.carFingerprint in HONDA_BOSCH:
-          self.accel = float(np.clip(accel, self.params.BOSCH_ACCEL_MIN, self.params.BOSCH_ACCEL_MAX))
-          self.gas = float(np.interp(accel, self.params.BOSCH_GAS_LOOKUP_BP, self.params.BOSCH_GAS_LOOKUP_V))
-
           stopping = actuators.longControlState == LongCtrlState.stopping
           self.stopping_counter = self.stopping_counter + 1 if stopping else 0
+
+          stoppingDecelAmount = max ( self.CP.stopAccel, self.stopping_counter * -self.CP.stoppingDecelRate / 50 ) # CC frame rate 50x speed of longplanner
+          self.accel = float(np.clip(aTarget + stoppingDecelAmount, self.params.BOSCH_ACCEL_MIN, self.params.BOSCH_ACCEL_MAX))
+          self.gas = float(np.interp(accel, self.params.BOSCH_GAS_LOOKUP_BP, self.params.BOSCH_GAS_LOOKUP_V))
```
sshane, line 216: This seems like it may be possible for both gas and brake to actuate at the same time.

mvl-boston: This overlap was intentional.   This allows for a smooth handover between gas and the Honda PID brake controller.   It won't actuate brake unless needed to maintain the commanded absolute accel.

sshane: We probably don't ever want overlap

mvl-boston:
Closing, replaced by:
https://github.com/commaai/opendbc/pull/2347

PR #2180, opened 2025-05-03
fix(macos): macOS 15 and greater

greatgitsby:
resolves: https://github.com/commaai/openpilot/issues/35093

special handling was added for mac to resolve gcc (internally, clang) to the actual GNU compiler (resolved as gcc-<ver> with brew): https://github.com/commaai/panda/pull/1241

with a recent update to MacOS (some time near 15.4), `gcc` was broken.

upon checking, clang17 supports `-std=gnu11`, which was the original complaint. as well, builds just fine. but there are still issues with building on less than 15 with clang.

using gcc on less than 15.

image

```
-- with .app --

➜ clang --version
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

-- with default CommandLineTools version, also supported --

➜ clang --version
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,7 @@ jobs:
         include:
           - os: ${{ github.repository == 'commaai/opendbc' && 'namespace-profile-amd64-8x16' || 'ubuntu-24.04' }}
           - os: macos-latest
+          - os: macos-15
     steps:
     - uses: commaai/timeout@v1
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


greatgitsby:
build works for macos-15 now but not 14 :/

andiradulescu:
Been there https://github.com/commaai/panda/issues/1957 :/

PR #2182, opened 2025-05-04
Tesla: fix disengagement without stalk

lukasloetkolben:
On Tesla models without a cruise control stalk, the steering wheel scroll wheel is used to engage and disengage. 
Unlike the stalk, pressing the scroll wheel button doesn't change the DI_state. This meant that pressing the scroll wheel button currently does not disengage openpilot, whenever openpilot's longitudinal control is enabled

When the scroll wheel button is pressed to cancel ACC, DAS_accState becomes 13. 
We have to listen to DAS_accState sent by the autopilot computer and add it to the cancel condition, to ensure that openpilot correctly disengages on a scroll wheel button press.

The screenshot shows several button is presses, but the DI_State does not change.
Bildschirmfoto 2025-05-04 um 12 39 51

route: f15cb91e37913f92/00000002--ee7e5b52c1

**We need to make sure that ACC_CANCEL_GENERIC_SILENT (state 13) is not sent by the autopilot computer in other scenarios where we don't want to disengage.**

### DIFF
--- a/opendbc/car/tesla/carcontroller.py
+++ b/opendbc/car/tesla/carcontroller.py
@@ -30,7 +30,9 @@ class CarController(CarControllerBase):
     # Longitudinal control
     if self.CP.openpilotLongitudinalControl:
       if self.frame % 4 == 0:
-        state = 13 if CC.cruiseControl.cancel else 4  # 4=ACC_ON, 13=ACC_CANCEL_GENERIC_SILENT
+        # Handle ACC cancel request from autopilot computer for cars without a stalk
+        cancel = CC.cruiseControl.cancel or CS.das_control["DAS_accState"] == 13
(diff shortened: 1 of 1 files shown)


sshane:
FSD always sets this value to 13 BTW, so we need to check that it doesn't ever set it to 13 for other reasons as you suspected.

sshane:
This is very confusing, can you get a stock route with Autopilot so we can compare with angle control being active? I don't understand when it's enabled vs. disabled here since it never falls

image

lukasloetkolben:
This is a stock route with Autopilot: f15cb91e37913f92/00000006--3fc2538b96

adeebshihadeh:
Is this still an issue? If so, I'd like to make sure it goes in 0.9.10 @sshane.

PR #2185, opened 2025-05-05
draft - HONDA: Correct gas strength for Bosch A Civic

mvl-boston:
Longitudinal tuning found that CAR.HONDA_CIVIC_BOSCH had 200% mistuned gas power.

- Longitudinal maneuvers route (before fix): d42bd5940e7d94dd/00000004--7bc3c111ee
- Longitudinal maneuvers route (after fix): in progress

Todo: Longitudinal maneuvers route

### DIFF
--- a/opendbc/car/honda/interface.py
+++ b/opendbc/car/honda/interface.py
@@ -86,7 +86,6 @@ class CarInterface(CarInterfaceBase):
 
     if candidate in HONDA_BOSCH:
       ret.longitudinalActuatorDelay = 0.5 # s
-      ret.boschMaxGas = 1600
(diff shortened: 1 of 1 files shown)


mvl-boston:
Found a different driver to retest the fix and included log.   Let me know if you need a long maneuver route rerun.

jyoung8607:
For future longitudinal tuning PRs, we ideally need to see apples-to-apples tests before and after, with only a single change at a time. That said, the "before" gas mapping is clearly very bad, and it seems decent in the "after" route.

I know Honda Bosch longitudinal has major tuning issues, and it's all gated behind the alpha longitudinal toggle, so we'll go ahead and merge this and then iterate on the rest.

PR #2198, opened 2025-05-07
Better tesla engage

sshane:
Brings back some of https://github.com/commaai/opendbc/pull/2175 so that you can enable while in tight curves. On master, handsOnLevel==3 is a noEntry, but Tesla's disengage tone is much too delayed so you think you've enabled when you actually haven't. Which is not great since you enable and let go of the steering wheel only to realize it isn't steering.

By waiting for handsOnLevel to fall (reducing user steering force), this should provide a smoother and more predictable engage in these curves.

Scenarios:

engaged:
- rising edge of steeringDisengage: both panda and openpilot disable via setting state=False, controls_allowed=False

disengaged:
- handsOnLevel==3 no longer is a noEntry. we enable ACC, but only start steering once user relaxes steering wheel.

### DIFF
--- a/opendbc/car/tesla/carcontroller.py
+++ b/opendbc/car/tesla/carcontroller.py
@@ -17,12 +17,17 @@ class CarController(CarControllerBase):
     actuators = CC.actuators
     can_sends = []
 
+    # Tesla EPS enforces disabling steering on heavy lateral override force.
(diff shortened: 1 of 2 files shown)


sshane:
Works as expected: https://connect.comma.ai/2c912ca5de3b1ee9/00000288--750fa23e86

PR #2186, opened 2025-05-05
Ford Q4 setup video, long_odbc_cable for Q4

alan-polk:
Added Cable.long_odbc_cable for Q4 harness and link to Q4 harness install video.

PR #2199, opened 2025-05-07
Revert "Ford Q4 setup video, long_odbc_cable for Q4"

sshane:
Reverts commaai/opendbc#2186


sshane:
@alan-polk we already add the long USB C cable

PR #2201, opened 2025-05-08
3071 Impreza / Crosstrek

ClockeNessMnstr:
upstream the "correct" values for crosstrek/subaru torque limit. 

This is for Global pre-2020 Impreza/Crosstrek as well as the hybrid although that is dashcam only upstream. 

The limit on these cars for steering is 3071 not 2047. 
Torque is similar but lower than the 2019 Forester which uses the 2047 limit and rate limits. 

stock Slew rate limit is right around 60. (For the forester it is around 30)


ClockeNessMnstr:
include a route or your device' dongle ID if relevant

ClockeNessMnstr:
63c983aefcc56c48/00000001--31309048c1/0

PR #2202, opened 2025-05-08
Tesla safety: check 2nd speed source

sshane:
Closes https://github.com/commaai/opendbc/issues/2200, matching how https://github.com/commaai/opendbc/pull/2049 added 2nd speed check, which was based on Ford

It looks like the signal in `DI_speed` uses the motor speeds to calculate its speed (calibration is likely [mentioned here](https://service.tesla.com/docs/ModelY/ServiceManual/en-us/GUID-5B457CEE-3E8B-4EBA-920E-5F9A2402F131.html), look for "Inverter Resolver Error Learning"), and the signal in `ESP_B` uses the wheel speed sensors ([mentioned here](https://service.tesla.com/docs/Model3/ServiceManual/en-us/GUID-E4D1FACE-A01E-4645-8DA0-0D9296C5AB4C.html)).

The closest Tesla error code I found was [DIR_a075_motorSpeedMismatch](https://tesware.net/alerts/model3/DIR_a075_motorSpeedMismatch), so it's very likely it also checks for mismatches between ESP and DI for Autopilot and would disengage.

An example of ESP_vehicleSpeed showing some noise correlating with wheel speeds:

<details><Summary>View</summary>

image
</details>

### DIFF
--- a/opendbc/dbc/tesla_model3_party.dbc
+++ b/opendbc/dbc/tesla_model3_party.dbc
@@ -111,6 +111,14 @@ BO_ 341 ESP_B: 8 PARTY
  SG_ ESP_wheelPulseCountFrR : 8|8@1+ (1,0) [0|254] "1"  app
  SG_ ESP_wheelPulseCountFrL : 0|8@1+ (1,0) [0|254] "1"  app
 
+BO_ 373 ESP_wheelSpeeds: 8 CH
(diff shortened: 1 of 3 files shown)


sshane:
When aEgo was > 5 m/s^2 here (AWD craze), the max delta between the two signals was 1.3 m/s:

image

PR #2204, opened 2025-05-09
Tesla safety: check rx counters

sshane:


sshane:
test_models' fuzzy test was catching valid mismatches, so that's why adding counter to CANParser is required for this

opendbc/can/common.h
```diff
@@ -28,6 +33,9 @@ unsigned int xor_checksum(uint32_t address, const Signal &sig, const std::vector
 unsigned int hkg_can_fd_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d);
 unsigned int fca_giorgio_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d);
 unsigned int pedal_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d);
+unsigned int tesla_checksum(uint32_t address, const Signal &sig, const std::vector<uint8_t> &d);
+
+bool tesla_is_counter(const Signal &sig);
```
sshane, line 38: @jyoung8607 does this kind of pattern enable any kind of refactors/unblock anything you've been wanting to do?

sshane: Actually it would be better to have one post init/setup hook for each DBC that can set things and run checks:

https://github.com/commaai/opendbc/blob/295be61b97b5bf92c3a3b656530a12b1562a0523/opendbc/can/dbc.cc#L94-L101

What do you think>

jyoung8607: A long time ago I proposed [Vector compliant message attributes](https://github.com/commaai/opendbc/pull/633#discussion_r910897364) that could be *added* to a DBC, without changing any of the existing naming or other content.

That said, now that car ports are part of opendbc, just a post init/setup hook where we can encapsulate all the manufacturer specific stuff would be easier, and wouldn't require any DBC changes at all.

I think I like your solution better.

opendbc/safety/safety/safety_tesla.h
```diff
@@ -16,6 +16,30 @@ static bool tesla_stock_lkas_prev = false;
...
+  if (addr == 0x2b9) {
+    // Signal: DAS_controlCounter
+    cnt = GET_BYTE(to_push, 6) >> 5;
+  } else if (addr == 0x488) {
+    // Signal: DAS_steeringControlCounter
+    cnt = GET_BYTE(to_push, 2) & 0x0FU;
+  } else if ((addr == 0x257) || (addr == 0x118) || (addr == 0x39d) || (addr == 0x286) || (addr == 0x311)) {
+    // Signal: DI_speedCounter, DI_systemStatusCounter, IBST_statusCounter, DI_locStatusCounter, UI_warningCounter
+    cnt = GET_BYTE(to_push, 1) & 0x0FU;
+  } else if ((addr == 0x155)) {
```
sshane, line 32:
```suggestion
  } else if (addr == 0x155) {
```

PR #2205, opened 2025-05-09
Tesla: fix cruise enabled test_models Autopark mismatch

sshane:


sshane:
@maxime-desroches another example of low MAX_EXAMPLES in CI not catching very much at all

maxime-desroches:
Setting up a cronjob on your local machine at night with a massive MAX_EXAMPLES would be great and the simplest thing to do right now

PR #2209, opened 2025-05-09
Ranger install video

alan-polk:
Ranger Install Video


opendbc/car/ford/values.py
```diff
@@ -82,6 +82,9 @@ def init_make(self, CP: CarParams):
     if harness == CarHarness.ford_q4:
       self.setup_video_link = "https://www.youtube.com/watch?v=uUGkH6C_EQU"
 
+    if CP.carFingerprint in (CAR.FORD_RANGER_MK2):
```
sshane, line 85: you're saying if the string is in the string

alan-polk: [CAR.FORD_RANGER_MK2]: instead of (CAR.FORD_RANGER_MK2):

?

PR #2217, opened 2025-05-10
Subaru: 2023 Ascent & LKAS Angle

bravochar:
This is a car port for the 2023 Subaru Ascent, including safety models for the Subaru Angle-based LKAS

2023 Subaru Ascent

- [ ] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/opcar/docs.py` to generate new docs
- [ ] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: "3d0a2ec61b4035d3|0000021e--146a1ecffd"
- [x] route with stock system: "99437cef6d5ff2ee|2023-03-13--21-21-38" (It was like that when I got here)
- [x] car harness used (if comma doesn't sell it, put N/A): Subaru D

### DIFF
--- a/opendbc/safety/tests/common.py
+++ b/opendbc/safety/tests/common.py
@@ -694,7 +694,8 @@ class AngleSteeringSafetyTest(VehicleSpeedSafetyTest):
   def test_angle_cmd_when_enabled(self):
     # when controls are allowed, angle cmd rate limit is enforced
     speeds = [0., 1., 5., 10., 15., 50.]
-    angles = np.concatenate((np.arange(-self.STEER_ANGLE_MAX * 2, self.STEER_ANGLE_MAX * 2, 5), [0]))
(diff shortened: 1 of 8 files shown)


opendbc/safety/tests/common.py
bravochar, line 1: LKAS_ANGLE is a signed, 17-bit integer, so angle values larger than ~655 degrees will over/underflow and this test will erroneously fail.

bravochar:
The move to `safety/modes` made git very sad, so I squashed everything to a single patch, re-applied it, and then force-pushed the new branch here. Bonus: cleaner history!

bravochar:
I know I'm getting a little stale, but I'm still interested in getting this merged. I should have some time this weekend to resolve those conflicts and get everything else up-to-date.

PR #2214, opened 2025-05-10
Tesla: let CANPacker set counters

sshane:
Still need to manually set `DAS_control` since we try to match the stock counter when canceling with stock long

```python
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
000000
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
000100
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
000200
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
000300
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
000400
print(packer.make_can_msg('APS_eacMonitor', 0, {})[1].hex())
```

PR #2220, opened 2025-05-12
Reapply "Tesla: let CANPacker set counters (#2214)"

sshane:
This reverts commit b9225178021be2b606dab89bf941440562f4cc6c.


sshane:
I see now. We pack the messages twice to calculate the checksums. We should move those into the CANPacker!

PR #2222, opened 2025-05-12
dbc: fix incorrect endwith check

sshane:
Two failure modes with current function:

- Suffix one character longer than string: false positive (npos is -1, so is subtraction)
- String starts with suffix and ends with suffix: false negative (gets first occurrence)

```c++
endswith("DI_gearDI", "DI") -> false
endswith("DI_gear", "Checksum") -> true
```

### DIFF
--- a/opendbc/can/common.h
+++ b/opendbc/can/common.h
@@ -44,7 +44,7 @@ unsigned int pedal_checksum(uint32_t address, const Signal &sig, const std::vect
   } while (false)
 
 inline bool endswith(const std::string& str, const char* suffix) {
-  return str.find(suffix, 0) == (str.length() - strlen(suffix));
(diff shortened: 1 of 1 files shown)


opendbc/can/common.h
```diff
@@ -44,7 +44,7 @@ unsigned int pedal_checksum(uint32_t address, const Signal &sig, const std::vect
   } while (false)
 
 inline bool endswith(const std::string& str, const char* suffix) {
-  return str.find(suffix, 0) == (str.length() - strlen(suffix));
+  return str.find(suffix, str.length() - strlen(suffix)) != std::string::npos;
```
sshane, line 47: https://cplusplus.com/reference/string/string/find/ says that a pos input bigger than the string is defined to return no match. Negative values are wrapped around to large positive values, so no undefined behavior here

PR #2229, opened 2025-05-13
WIP - Honda Regen support

vanillagorillaa:
https://github.com/commaai/opendbc/issues/1100

Still needs to be tested. Maybe check if these messages are found on other Honda cars. Maybe get a list of which Honda Cars have regen support (just hybrids?)


opendbc/safety/tests/test_honda.py
```diff
@@ -565,6 +565,15 @@ def _accel_msg(self, accel):
   def test_spam_cancel_safety_check(self):
     pass
 
+# ********************* Honda Regen **********************
+
+
+class TestHondaRegenSafetyBase(HondaBase):
+
+  # existence of _user_regen_msg adds regen tests
```
sshane, line 573: add to base class?

sshane:
What's left to do here?

vanillagorillaa:
I've only been able to get a test user on a Clarity, and seeing some issues there still (reporting pedal pressed or LKAS Fault when trying to engage). Haven't had a chance to debug further. Have not also found a user with an upstreamed car to test, however, the new CANFD cards have paddles, so once those are merged I will try again.

PR #2231, opened 2025-05-13
Allow configurable maximum angle for AngleSteeringSafetyTest

bravochar:
The Subaru LKAS_ANGLE message is limited to a signed, 17-bit integer, which is more than enough for it's maximum steering of around 485 degrees (at least for my Ascent). However, the `test_angle_cmd_when_enabled()` defaults to twice the maximum steering angle, which will overflow this value and result in erroneously failed tests.

I have added a class variable, and logic that will default the test angle to twice the max angle value when it is not specifically configured. This keeps current behavior for existing tests, but will allow me to constraint he maximum test angle to the limits of my CAN bus messages. It's quite possible that other vehicles will be affected as more angle-based steering systems are supported, too.

### DIFF
--- a/opendbc/safety/tests/common.py
+++ b/opendbc/safety/tests/common.py
@@ -683,6 +683,7 @@ class AngleSteeringSafetyTest(VehicleSpeedSafetyTest):
   def test_angle_cmd_when_enabled(self):
     # when controls are allowed, angle cmd rate limit is enforced
     speeds = [0., 1., 5., 10., 15., 50.]
+    # TODO: what should CANPacker do here? we should also have good coverage checks on this
     if self.STEER_ANGLE_TEST_MAX is None:
         self.STEER_ANGLE_TEST_MAX = self.STEER_ANGLE_MAX * 2
     angles = np.concatenate((np.arange(-self.STEER_ANGLE_TEST_MAX, self.STEER_ANGLE_TEST_MAX, 5), [0]))
--- a/opendbc/safety/tests/common.py
+++ b/opendbc/safety/tests/common.py
@@ -644,6 +644,7 @@ class VehicleSpeedSafetyTest(PandaSafetyTestBase):
 class AngleSteeringSafetyTest(VehicleSpeedSafetyTest):
 
   STEER_ANGLE_MAX: float = 300
+  STEER_ANGLE_TEST_MAX: float = None
   DEG_TO_CAN: float
   ANGLE_RATE_BP: list[float]
   ANGLE_RATE_UP: list[float]  # windup limit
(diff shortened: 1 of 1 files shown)


bravochar:
I need this for https://github.com/commaai/opendbc/pull/2217

adeebshihadeh:
It's not clear this is the right fix... perhaps we should clip to the max in the CANPacker? That still doesn't fail explicitly though.

bravochar:
While it's possible that there is a better fix, I don't have the knowledge to identify or execute it. There are many other instances in these Python test where the code for the vehicles communicates limits and restrictions back to the base tests in `common.py` - `STEER_ANGLE_MAX` in this same class for example.

I definitely see the wisdom in there being a mechanism for the packer to throw some sort of exception or error, but I think that even that would just result in the fine tuning of test parameters for those cases where the CAN packer failed, like this one.

Another option might be to initiate that range of angles to something smaller then `+/- 2*STEER_ANGLE_MAX`. Something like the maximum steering angle plus twice the maximum adjustment up would allow for two tests beyond the safety boundary and not risk overflowing small values?

I'm really interested in getting something merged so I can remove my temporary fix from #2217

jacobwaller:
@adeebshihadeh @bravochar is there a path forward here?

opendbc/safety/tests/common.py
```diff
@@ -682,7 +683,9 @@ def test_steering_angle_measurements(self):
   def test_angle_cmd_when_enabled(self):
     # when controls are allowed, angle cmd rate limit is enforced
     speeds = [0., 1., 5., 10., 15., 50.]
-    angles = np.concatenate((np.arange(-self.STEER_ANGLE_MAX * 2, self.STEER_ANGLE_MAX * 2, 5), [0]))
+    if self.STEER_ANGLE_TEST_MAX is None:
+        self.STEER_ANGLE_TEST_MAX = self.STEER_ANGLE_MAX * 2
```
adeebshihadeh, line 688:
```suggestion
      self.STEER_ANGLE_TEST_MAX = self.STEER_ANGLE_MAX * 2
```

opendbc/safety/tests/common.py
```diff
@@ -682,7 +683,9 @@ def test_steering_angle_measurements(self):
   def test_angle_cmd_when_enabled(self):
     # when controls are allowed, angle cmd rate limit is enforced
     speeds = [0., 1., 5., 10., 15., 50.]
-    angles = np.concatenate((np.arange(-self.STEER_ANGLE_MAX * 2, self.STEER_ANGLE_MAX * 2, 5), [0]))
+    if self.STEER_ANGLE_TEST_MAX is None:
```
adeebshihadeh, line 687:
```suggestion
    # TODO: what should CANPacker do here? we should also have good coverage checks on this
    if self.STEER_ANGLE_TEST_MAX is None:
```

adeebshihadeh (approved):

Issue #2237, opened 2025-05-14
Enable `cpplint`


adeebshihadeh:
Step 1 is to enable cpplint in the pre-commit config: https://github.com/commaai/opendbc/blob/57be4f17cde3c58a5c9bc9afcb0bfb7a6172350a/.pre-commit-config.yaml#L32.

Then run `./test.sh` and fix all the reported errors until `./test.sh` passes.

PR #2238, opened 2025-05-14
enable cpplint

aubsw:
For https://github.com/commaai/opendbc/issues/2237


opendbc/safety/safety/safety_defaults.h
```diff
@@ -13,7 +13,7 @@ void default_rx_hook(const CANPacket_t *to_push) {
 
 static safety_config nooutput_init(uint16_t param) {
   UNUSED(param);
-  return (safety_config){NULL, 0, NULL, 0, true};
+  return (safety_config){NULL, 0, NULL, 0, true}; // NOLINT(readability/braces)
```
aubsw, line 16:
```
opendbc/safety/safety/safety_defaults.h:16:  You don't need a ; after a }  [readability/braces] [4]
```

cpplint seems to be wrong here...

Issue #2156, opened 2025-04-24
Hyundai 2016 Genesis: OP does not engage after pressing "Cruise" and then "Set"    takes multiple presses of the "Set" button to engage.


stevegf3675:
Hyundai Genesis 2016 AWD / No SMDPS

Comma3

0.9.8-release / releasec3 / 5a84ebb / April 16

installed 2025/04/24 on my Comma3

After pressing the "Cruise" button and then pressing "set," the SCC does not engage and requires multiple presses to engage.

Preserved routes with this behavior :

27a580d124055206/00000000--b81083aa2e
27a580d124055206/00000001--3399a5981f
27a580d124055206/00000002--3748f9144e

sshane:
Does it say anything when you press SET?

stevegf3675:
There are no messages. The only icon visible on the dashboard is the SCC icon, which flashes each time you press the “set button.” Most of the time, it requires a quick double-press to engage OP.

sshane:
This will be fixed in https://github.com/commaai/openpilot/pull/35207

PR #2262, opened 2025-05-18
Hyundai: fix recent enable button press regression

sshane:
We went from checking every 50 Hz CLU11 button message to checking at 100 Hz, so we halve the time we sample for user button presses. I guess this wasn't a problem for most cars, but the issue below shows the PCM enabling a little bit out of the new halved sample window which is now fixed again.

Broken in https://github.com/commaai/openpilot/pull/33710

Closes https://github.com/commaai/opendbc/issues/2156

Matching old reference carstate and interface: https://github.com/commaai/openpilot/blob/6a15aa3f8c58e40e5e47c7ce4c906f5368ec243a/selfdrive/car/hyundai/interface.py


sshane:
image

PR #2265, opened 2025-05-18
Hyundai: move low speed steering alert to car state

sshane:


sshane:
@jyoung8607 wanna open a PR for VW?

PR #2267, opened 2025-05-18
Split safety.h checks

adeebshihadeh:


adeebshihadeh:
Worth noting that Codex didn't pick up the MISRA failure since `./test.sh` doesn't cover it!

https://github.com/commaai/opendbc/issues/2268

PR #2269, opened 2025-05-18
Update docs generation

adeebshihadeh:


docs/CARS.md
```diff
@@ -2,20 +2,18 @@
 
 # Support Information for 361 Known Cars
 
+# Upstream Cars
+
 |Make|Model|Package|Support Level|
 |---|---|---|:---:|
```
sshane, line 8: remove support level here?

PR #2275, opened 2025-05-20
run safety tests in test.sh

adeebshihadeh:


.github/workflows/tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        flags: ['', '--ubsan']
+        flags: ['--ubsan']
```
greatgitsby, line 35: you can delete the matrix strategy now

adeebshihadeh: leaving since we'll want to run with other sanitizers soon

PR #2279, opened 2025-05-20
try moving pytest to lefthook

adeebshihadeh:


adeebshihadeh:
I don't like the test.sh -> lefthook.yml -> pytest/test_misra.sh/etc indirection, but it is a decent speedup.

PR #2284, opened 2025-05-22
Rivian: update VDM_AdasSts

lukasloetkolben:


sshane:
What does this do?

lukasloetkolben:
It does not change any logic. It's just a minor correction.

PR #2291, opened 2025-05-24
safety cleanups

sshane:


opendbc/safety/safety_declarations.h
```diff
@@ -64,7 +64,8 @@ extern const int MAX_WRONG_COUNTERS;
 // used to represent floating point vehicle speed in a sample_t
 #define VEHICLE_SPEED_FACTOR 1000.0
 #define MAX_TORQUE_RT_INTERVAL 250000U
-
+//#define MS_TO_KPH 3.6
+#define KPH_TO_MS(kph) ((kph) / 3.6)
```
adeebshihadeh, line 68: functions are preferable when possible

sshane: a constant seems more reasonable to me here

adeebshihadeh: re: constant, i have a slight preference for the function, but up to you

PR #2298, opened 2025-05-24
Tesla: generic angle cmd limit check

sshane:


opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
 }
 
+static float get_curvature_factor(const float speed, const AngleSteeringParams params) {
+  return 1. / (1. - (params.slip_factor * (speed * speed))) / params.wheelbase;
+}
+
+bool steer_angle_cmd_checks_vm(int desired_angle, bool steer_control_enabled, const AngleSteeringLimits limits,
+                               const AngleSteeringParams params) {
+  // This check uses a simple vehicle model to allow for true lateral acceleration and jerk limits.
+  // TODO: remove the inaccurate breakpoint angle limiting function above and always use this one
```
adeebshihadeh, line 834: what does true mean? be more specific like "constant jerk"

opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+
+  static const float RAD_TO_DEG = 57.29577951308232;
+  static const float ISO_LATERAL_ACCEL = 3.0;  // m/s^2
```
adeebshihadeh, line 837: why is max jerk also not defined?

opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+    violation |= max_limit_check(desired_angle, max_angle_can, -max_angle_can);
+  }
+  desired_angle_last = desired_angle;
+
+  // Angle should either be 0 or same as current angle while not steering
+  if (!steer_control_enabled) {
+    const int max_inactive_angle = CLAMP(angle_meas.max, -limits.max_angle, limits.max_angle) + 1;
+    const int min_inactive_angle = CLAMP(angle_meas.min, -limits.max_angle, limits.max_angle) - 1;
+    violation |= (limits.inactive_angle_is_zero ? (desired_angle != 0) :
+                  max_limit_check(desired_angle, max_inactive_angle, min_inactive_angle));
```
adeebshihadeh, line 882: how does this pass the safety coverage tests? we should only hit one branch

opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+
+  if (controls_allowed && steer_control_enabled) {
+    // *** ISO lateral jerk limit ***
+    // calculate maximum angle rate per second
+    const float speed = MAX(fudged_speed, 1.0);
+    const float max_curvature_rate_sec = MAX_LATERAL_JERK / (speed * speed);
+    const float max_angle_rate_sec = max_curvature_rate_sec * params.steer_ratio / curvature_factor * RAD_TO_DEG;
+
+    // finally get max angle delta per frame
+    const float max_angle_delta = max_angle_rate_sec * (0.01f * 2.0f);  // 50 Hz
```
adeebshihadeh, line 859: 50Hz should be pulled out, this is Tesla-specific

opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
...
+
+  const float fudged_speed = (vehicle_speed.min / VEHICLE_SPEED_FACTOR) - 1.;
+  const float curvature_factor = get_curvature_factor(fudged_speed, params);
+
+  bool violation = false;
+
+  if (controls_allowed && steer_control_enabled) {
+    // *** ISO lateral jerk limit ***
+    // calculate maximum angle rate per second
+    const float speed = MAX(fudged_speed, 1.0);
```
adeebshihadeh, line 854: `fudged_speed` reads like it's downstream of `speed`, but we go from vehicle_speed.min -> fudged_speed -> speed. I'd just collapse it into this
```
const float speed = MAX((vehicle_speed.min / VEHICLE_SPEED_FACTOR) - 1., 1.0);
```

opendbc/safety/safety.h
```diff
@@ -824,6 +824,70 @@ bool steer_angle_cmd_checks(int desired_angle, bool steer_control_enabled, const
   return violation;
 }
 
+static float get_curvature_factor(const float speed, const AngleSteeringParams params) {
+  return 1. / (1. - (params.slip_factor * (speed * speed))) / params.wheelbase;
+}
+
+bool steer_angle_cmd_checks_vm(int desired_angle, bool steer_control_enabled, const AngleSteeringLimits limits,
```
adeebshihadeh, line 831: seems this doesn't have any real time checks. we absolutely need these, e.g. what happens if I send the steer command at 100Hz?

opendbc/safety/safety.h
sshane: previous angle limiting was in angle space, so not true jerk as it varied across the speed range. constant is more clear

opendbc/safety/safety.h
sshane: I think it's by line, so without an if statement it always hits the ternary. Will switch it to if

sshane: You also removed coverage testing when you removed the safety job

opendbc/safety/safety.h
sshane: we need to use the non-fudged and fudged for a reason to match OP, let me check if I can simplify

opendbc/safety/safety.h
sshane: We never had these for angle limiting, but I thought about adding a generic frequency param for relevant tx messages to check for this, and removes torque's duplicate limits * 1.2 as well

opendbc/safety/safety.h
sshane: defined like what? 3 isn't ISO jerk

opendbc/safety/safety.h
sshane: https://github.com/commaai/opendbc/pull/2328

opendbc/safety/safety.h
sshane: https://github.com/commaai/opendbc/pull/2329

opendbc/safety/safety.h
sshane: https://github.com/commaai/opendbc/pull/2334

PR #2303, opened 2025-05-25
safety: default quality flag check to true

sshane:


opendbc/safety/modes/body.h
```diff
@@ -31,7 +31,7 @@ static bool body_tx_hook(const CANPacket_t *to_send) {
 
 static safety_config body_init(uint16_t param) {
   static RxCheck body_rx_checks[] = {
-    {.msg = {{0x201, 0, 8, .ignore_checksum = true, .ignore_counter = true, .frequency = 100U}, { 0 }, { 0 }}},
+    {.msg = {{0x201, 0, 8, .ignore_checksum = true, .ignore_counter = true, .ignore_quality_flag=true, .frequency = 100U}, { 0 }, { 0 }}},
```
adeebshihadeh: this should be more like “has quality flag”

adeebshihadeh: or should we call them validity flags? did “quality” come from ford?

sshane: it seems to be common among the brands. Tesla and Ford call it qf https://github.com/commaai/opendbc/pull/2227#issue-3058732911

adeebshihadeh: sure, but "ignore quality flag" sounds like it's there but we're choosing not to respect it

sshane: I read it as "ignore quality flag check." Agree it's not the most clear, but it matches the other counter & checksum skip flags for now.

There are both cases where a message has counter but we want to ignore it due to some issue and the message having no counter. One name will be less ideal for one case.

PR #2314, opened 2025-05-28
draft Honda Canfd stopgo test - will delete

mvl-boston:
ci checks on local repositories are timing out for some reason.   Will delete once ci is done.


mvl-boston:
Deleting - ci tests passed

PR #2315, opened 2025-05-28
ci: no timeout for forks

maxime-desroches:


mvl-boston:
Thanks!  very helpful.

still having one test timing out, is this controlled somewhere else?

https://github.com/mvl-boston/opendbc/actions/runs/15291490451/job/43011656401?pr=112

PR #2317, opened 2025-05-28
HKG Adding new camera signals and updating existing

devtekve:
I came across some dbc signals that are interesting, these are official signals from an official DBC I found on an open source repo from what seems to be a 3rd party outsourced from hyundai. I have a many more but I am trying to make sure they have some meaning before pushing them.


devtekve:
I will cleanup a bit, because the diff is too large

devtekve:
@sshane reminder that this is still here :D

devtekve (commented):
Missed a property and some of the valus mapped

sshane:
Not sure what any of these do. Effectively a well described message name was made more obfuscated. I know it's more correct now, but that along with no information about these other messages in the description makes this hard to evaluate priority for. I assume this is needed for a port of some kind?

We would have merged this sooner if it was clearly laid out why this is needed.

sshane:
If the usage is not clear in the port PR, can you add DBC message comments for these in a new PR?

PR #2319, opened 2025-05-28
cache mull-17

sshane:


sshane:
probably not needed with https://github.com/commaai/opendbc/commit/fc84c6602f90ad6f64d750fe2d6aaed838eda92b for now. should replace mull since we can't easily cache it fully (setup script can be slow)

PR #2326, opened 2025-05-29
Restore safety coverage testing

sshane:


test.sh
```diff
@@ -6,12 +6,31 @@ cd $DIR
 
 source ./setup.sh
 
+# reset safety coverage data and generate gcc note file
+rm -f ./libsafety/*.gcda
+
 # *** build ***
-scons -j8 "$@"
+scons -j8 --coverage "$@"
```
adeebshihadeh, line 13: does this need to be a flag? why not make it the default?

test.sh
```diff
@@ -6,12 +6,31 @@ cd $DIR
...
 
 # *** lint + test ***
 lefthook run test
 
+# generate and open report
+if [ "$1" == "--report" ]; then
+  mkdir -p coverage-out
+  gcovr -r ../ --html-nested coverage-out/index.html
+  sensible-browser coverage-out/index.html
+fi
```
adeebshihadeh, line 23: can this go in pytest?

test.sh
sshane: don't notice any extra test time https://github.com/commaai/opendbc/pull/2327

adeebshihadeh: red diff!

sshane:
Revert this first: https://github.com/commaai/opendbc/pull/2449

sshane:
already back

PR #2327, opened 2025-05-29
safety: always build with coverage information

sshane:


devtekve:
This breaks building openpiot on my mac though 

```
cc -o opendbc_repo/opendbc/safety/tests/libsafety/libsafety.so -dynamiclib opendbc_repo/opendbc/safety/tests/libsafety/safety.os -lgcov
ld: library 'gcov' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
scons: *** [opendbc_repo/opendbc/safety/tests/libsafety/libsafety.so] Error 1
```

PR #2330, opened 2025-05-30
safety: check tx message frequency

sshane:
This PR is to add a generic max frequency param that is checked for any message that specifies it, such as lat/long actuation messages. It will replace the rt logic:

https://github.com/commaai/opendbc/blob/65b880d13b416fd662d2d2961c89b9eb66a4a123/opendbc/safety/safety.h#L665-L673


sshane:
See https://github.com/commaai/opendbc/pull/2332#issuecomment-2921142337

PR #2335, opened 2025-05-30
Tesla: realtime angle limit checks

sshane:
https://github.com/commaai/opendbc/pull/2333 but works


sshane:
Added support and enabled for Tesla in https://github.com/commaai/opendbc/pull/2343

PR #2339, opened 2025-05-30
angle safety: implement realtime checks

sshane:
- every 250ms window, store previous angle ~and speed~
- to calculate the max delta for this current 250ms window, use old angle and new speed
  - if speed decreases, old angle and new speed will allow correct higher non-realtime angle delta
  - if speed increases, old angle and new speed will block higher than allowed angle delta


sshane:
Added support and enabled for Tesla in https://github.com/commaai/opendbc/pull/2343

PR #2342, opened 2025-05-31
Honda: Bosch stopsign/stoplight fixes

mvl-boston:
Prior Honda Bosch code released brakes above -0.2m/s at all speeds.

That prior logic improved smoothness at driving speeds, but caused experimental mode to release brakes prior to completing a full stop at stoplights/stopsigns.

This PR changes logic at low speeds to remove the -0.2 buffer so the car can execute a full stop.

9662a1dab70c9dc3/0000006c--903762454e


opendbc/car/honda/hondacan.py
```diff
@@ -74,9 +75,9 @@ def create_brake_command(packer, CAN, apply_brake, pump_on, pcm_override, pcm_ca
   return packer.make_can_msg("BRAKE_COMMAND", CAN.pt, values)
 
 
-def create_acc_commands(packer, CAN, enabled, active, accel, gas, stopping_counter, car_fingerprint):
+def create_acc_commands(packer, CAN, enabled, active, accel, gas, stopping_counter, car_fingerprint, vEgo):
   commands = []
-  min_gas_accel = CarControllerParams.BOSCH_GAS_LOOKUP_BP[0]
+  min_gas_accel = float (np.interp(vEgo, [5.0, 10.0], [0.01, CarControllerParams.BOSCH_GAS_LOOKUP_BP[0]]))
```
sshane, line 80: I don't think you want to change this without changing other places it's referenced (carcontroller).

Sure, now you can command brakes from 0 m/s^2 instead of -0.2 m/s^2, but gas looks like it will still command more than it should until it brakes. For example, the point it switches from gas to brake now is not smooth, it will snap from some non-zero value to braking.

```
>>> np.interp(0, [-0.2, 2.0], [0, 1600])
np.float64(145.45454545454547)
```

I agree that if we're controlling gas/brake directly in an ICE car with creep acceleration, we need some kind of compensation for optimal control

opendbc/car/honda/hondacan.py
```diff
@@ -74,9 +75,9 @@ def create_brake_command(packer, CAN, apply_brake, pump_on, pcm_override, pcm_ca
   return packer.make_can_msg("BRAKE_COMMAND", CAN.pt, values)
 
 
-def create_acc_commands(packer, CAN, enabled, active, accel, gas, stopping_counter, car_fingerprint):
+def create_acc_commands(packer, CAN, enabled, active, accel, gas, stopping_counter, car_fingerprint, vEgo):
   commands = []
-  min_gas_accel = CarControllerParams.BOSCH_GAS_LOOKUP_BP[0]
```
sshane, line 75: `CarControllerParams.BOSCH_GAS_LOOKUP_BP[0]`

turn this into a function that takes speed?

opendbc/car/honda/hondacan.py
mvl-boston: The goal was just to fix stoplights and stopsigns.

The gas only pid controller in https://github.com/commaai/opendbc/pull/2347 should clean up any snapping and creep issues on cutover.

Not 100% sure why the -0.2 was put there, another reason may have been to leave some room for wind drag to preserve brake pads when holding setspeed, or following a lead with slightly varying speeds.

opendbc/car/honda/hondacan.py
mvl-boston: Confirming this comment: you prefer moving the function from hondacan to values.py where it is declared?

Wouldn't it be better to put the function in carcontroller since it already has access to vEgo, and just pass the result to hondacan?

jyoung8607:
Marking draft for the same reasons described in #2347.

PR #2343, opened 2025-05-31
Tesla: simple angle frequency check

sshane:
Allows for 1.2x frequency buffer in any 250ms window, just like torque.

If 100Hz, and if a lag of x message causes x+1 messages sent the next frame, openpilot can lag for up to 5 messages/50ms with no false positive blocks.

50Hz is also 50ms, but 2.5 messages.


opendbc/safety/tests/safety_replay/helpers.py
```diff
@@ -84,6 +84,15 @@ def init_segment(safety, msgs, mode, param):
...
 
+  cnt = 0
+  for msg in msgs:
+    if msg.which() == 'can':
+      cnt += 1
+      if cnt > 100:
+        break
+      for can in msg.can:
+        safety.safety_rx_hook(package_can_msg(can))
+
```
sshane:
```suggestion
```

opendbc/safety/modes/ford.h
```diff
@@ -110,6 +110,7 @@ static bool ford_get_quality_flag_valid(const CANPacket_t *to_push) {
                                                                                                 \
   /* no blending at low speed due to lack of torque wind-up and inaccurate current curvature */ \
   .angle_error_min_speed = 10.0,    /* m/s */                                                   \
+  .frequency = 20U,  /* Hz */                                                                   \
```
adeebshihadeh: woah, didn't realize ford steering was so low freq!

opendbc/safety/modes/nissan.h
```diff
@@ -67,6 +67,7 @@ static bool nissan_tx_hook(const CANPacket_t *to_send) {
       {0., 5., 15.},
       {5., 3.5, .4}
     },
+    .frequency = 100U,  // 100 Hz
```
adeebshihadeh: don't duplicate the value in the comment. and ideally put the unit in the variable name if it's unclear, but in this case, i think it's clear enough

opendbc/safety/tests/test_tesla.py
```diff
@@ -42,6 +42,10 @@ class TestTeslaSafetyBase(common.PandaCarSafetyTest, common.AngleSteeringSafetyT
   ANGLE_RATE_UP = None
   ANGLE_RATE_DOWN = None
 
+  # Real time limits
+  RT_INTERVAL = 250000
```
adeebshihadeh: we should really standardize on seconds. anything else is way too hard to read

adeebshihadeh: also don't we use the same interval everywhere? why don't we just define it once?

opendbc/safety/tests/test_tesla.py
```diff
@@ -344,6 +348,23 @@ def test_lateral_jerk_limit(self):
...
+      self.assertEqual(should_tx, self._tx(self._angle_cmd_msg(0, True)))
+
+    # Test recovery after sending too many messages
+    for rt_ts in np.arange(0, self.RT_INTERVAL * 2, 10000):
+      should_tx = rt_ts > self.RT_INTERVAL
+      self.safety.set_timer(rt_ts)
+      self.assertEqual(should_tx, self._tx(self._angle_cmd_msg(0, True)))
+      if should_tx:
+        # Break before we block again
+        break
```
adeebshihadeh: this test is pretty hard to read, especially the early termination condition

opendbc/safety/tests/test_tesla.py
```diff
@@ -344,6 +348,23 @@ def test_lateral_jerk_limit(self):
         # Recover
         self.assertTrue(self._tx(self._angle_cmd_msg(0, True)))
 
+  def test_rt_limits(self):
```
adeebshihadeh: this should be a common test!

adeebshihadeh (approved):
LGTM aside from comments

opendbc/safety/tests/test_tesla.py
sshane: was going to move it once I made this a common test

opendbc/safety/modes/ford.h
```diff
@@ -110,6 +110,7 @@ static bool ford_get_quality_flag_valid(const CANPacket_t *to_push) {
                                                                                                 \
   /* no blending at low speed due to lack of torque wind-up and inaccurate current curvature */ \
   .angle_error_min_speed = 10.0,    /* m/s */                                                   \
+  .frequency = 20U,                                                                             \
```
sshane:
```suggestion
```

opendbc/safety/modes/nissan.h
```diff
@@ -67,6 +67,7 @@ static bool nissan_tx_hook(const CANPacket_t *to_send) {
       {0., 5., 15.},
       {5., 3.5, .4}
     },
+    .frequency = 100U,
```
sshane:
```suggestion
```

opendbc/safety/modes/toyota.h
```diff
@@ -190,6 +190,7 @@ static bool toyota_tx_hook(const CANPacket_t *to_send) {
       {5., 25., 25.},
       {0.36, 0.26, 0.26}
     },
+    .frequency = 100U,
```
sshane:
```suggestion
```

PR #2191, opened 2025-05-06
Honda: Fix gas strength for Bosch C Accord

mvl-boston:
Longitudinal tuning found that CAR.HONDA_ACCORD_11G had 50% mistuned gas power.

- Longitudinal maneuvers route (before fix): 626e73503debb76c/0000000e--4129252dce
- Longitudinal maneuvers route (after fix): in progress
- Testdrive after fix: 15bd03c45fb5f401/0000000a--1805a6553f

Notes:
- This code is dependent on https://github.com/commaai/opendbc/pull/2120 (merged in as baseline for this PR)
- Before test also included https://github.com/commaai/opendbc/pull/2162
- After test will include https://github.com/commaai/opendbc/pull/2179, https://github.com/commaai/opendbc/pull/2165, and https://github.com/commaai/openpilot/pull/35109, and https://github.com/commaai/opendbc/pull/2162.

Todo: Longitudinal maneuvers route after fix

PR #2347, opened 2025-06-01
Draft - Honda: Bosch gasonly pid

mvl-boston:
(replacement of [opendbc pr2179](https://github.com/commaai/opendbc/pull/2179) and [opendbc pr35375](https://github.com/commaai/openpilot/pull/35375), moved to a carcontroller PID at Shane's request.  This is intended to be identical behavior.)

Honda Bosch has a built in PID tuner for brake, but does not have one for gas.

Even after adjusting gas for detectable conditions (incline, speed-based drag), these calculations are always bit off for undetectable conditions like tailwinds, weather/air density, etc. Thus a gas PID tuner is helpful for Honda Bosch.

However, the current openpilot PID tuner is only on overall accel. Any PID adjustment on gas then applies to brake, doubling up with the builtin Honda brake PID, and causing undesired braking strength until openpilot's PID adjustment winds down. Once it winds down, then you get undesired gas strength until it winds up again.

This pr allows for the PID tuner to only impact gas on Honda Bosch models.
Todo: Test route on new carcontroller-only logic

Prior test routes from [opendbc pr2179](https://github.com/commaai/opendbc/pull/2179), that should have identical behavior:

    Bosch C (2023 Honda Accord Hybrid): 15bd03c45fb5f401/0000000a--1805a6553f (test included https://github.com/commaai/opendbc/pull/2191)
    Bosch C (2023 Honda Pilot): 9662a1dab70c9dc3/00000032--d0dac7d95b
    Bosch C (2025 Acura MDX): ad9840558640c31d/00000013--893d36d514
    more coming from other cars in discord

Discord feedback:

    I drove for five hours, and it was an incredible experience."
    "worked well on the way to work today"
    "Worked way better than the 0.9.8 branch I had last been using. Actually useable this time around. "
    "once we can fix the rapid acceleration issue I think I’d daily op long" (retest in process after accel fix in https://github.com/commaai/opendbc/pull/2185)
    "Acceleration and braking were exactly as they should be. I drove over to a construction zone to see how it would handle sudden starts/stops, and it was perfect." (test included https://github.com/commaai/opendbc/pull/2191)

### DIFF
--- a/opendbc/car/honda/carcontroller.py
+++ b/opendbc/car/honda/carcontroller.py
@@ -4,8 +4,10 @@ from collections import namedtuple
 from opendbc.can.packer import CANPacker
 from opendbc.car import Bus, DT_CTRL, rate_limit, make_tester_present_msg, structs
 from opendbc.car.honda import hondacan
-from opendbc.car.honda.values import CruiseButtons, VISUAL_HUD, HONDA_BOSCH, HONDA_BOSCH_RADARLESS, HONDA_NIDEC_ALT_PCM_ACCEL, CarControllerParams
+from opendbc.car.honda.values import CruiseButtons, VISUAL_HUD, HONDA_BOSCH, HONDA_BOSCH_RADARLESS, HONDA_NIDEC_ALT_PCM_ACCEL, \
+                                     CarControllerParams, GasOnlyTuning
 from opendbc.car.interfaces import CarControllerBase
+from opendbc.car.common.pid import PIDController
 
 VisualAlert = structs.CarControl.HUDControl.VisualAlert
 LongCtrlState = structs.CarControl.Actuators.LongControlState
@@ -115,6 +117,9 @@ class CarController(CarControllerBase):
     self.gas = 0.0
     self.brake = 0.0
     self.last_torque = 0.0
+    self.gasonly_pid = PIDController (k_p=([0,], [0,]),
+                                      k_i=(GasOnlyTuning.kiBP, GasOnlyTuning.kiV),
+                                      k_f=1, rate=2 / DT_CTRL )
 
   def update(self, CC, CS, now_nanos):
     actuators = CC.actuators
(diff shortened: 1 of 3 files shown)


opendbc/car/honda/interface.py
```diff
@@ -86,10 +87,13 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
 
     if candidate in HONDA_BOSCH:
       ret.longitudinalActuatorDelay = 0.5 # s
+      # default longitudinal gas-only tuning for all Bosch hondas
+      GasOnlyTuning.kiBP = [0., 5., 35.]
+      GasOnlyTuning.kiV = [1.2, 0.8, 0.5]
```
sshane: Modifying a global variable, much less an uninstantiated class is not a good practice.

opendbc/car/honda/carcontroller.py
```diff
@@ -207,7 +212,11 @@ def update(self, CC, CS, now_nanos):
 
         if self.CP.carFingerprint in HONDA_BOSCH:
           self.accel = float(np.clip(accel, self.params.BOSCH_ACCEL_MIN, self.params.BOSCH_ACCEL_MAX))
-          self.gas = float(np.interp(accel, self.params.BOSCH_GAS_LOOKUP_BP, self.params.BOSCH_GAS_LOOKUP_V))
+
+          # perform a gas-only pid
+          gas_error = self.accel - CS.out.aEgo
+          gas_pedal_force = self.gasonly_pid.update(gas_error, speed=CS.out.vEgo, feedforward=self.accel)
+          self.gas = float(np.interp(gas_pedal_force, self.params.BOSCH_GAS_LOOKUP_BP, self.params.BOSCH_GAS_LOOKUP_V))
```
sshane, line 225: make sure we don't apply gas while applying brake

mvl-boston: Yes, this happens in hondacan:
82:   gas_command = gas if active and gas_force > min_gas_accel else -30000
84:   braking = 1 if active and gas_force < min_gas_accel else 0

82 and 84 disable gas and brakes, respectively, based on exact opposites of the same condition clause.

opendbc/car/honda/interface.py
mvl-boston: Done.   Moved to carcontroller.

jyoung8607:
Per our discussion, I'd like to try PIF control with drivetrain torque as the feedforward rather than acceleration.

Keeping this open but marking as draft, since it's now Plan B.

PR #2355, opened 2025-06-02
Hyundai CAN FD: add supported configuration checking

sshane:
Would show dashcam mode for a user in Zendesk today instead of various CAN bus disconnect errors (non-US car with HDA2 topology but marketed as HDA1, so got wrong harness)

### DIFF
--- a/opendbc/car/hyundai/interface.py
+++ b/opendbc/car/hyundai/interface.py
@@ -157,8 +157,6 @@ class CarInterface(CarInterfaceBase):
     if candidate in (CAR.KIA_OPTIMA_H,):
       ret.dashcamOnly = True
 
-    print('ret.flags:', ret.flags)
-    # raise Exception
     return ret
 
   @staticmethod
--- a/opendbc/car/hyundai/interface.py
+++ b/opendbc/car/hyundai/interface.py
@@ -33,6 +33,10 @@ class CarInterface(CarInterfaceBase):
     cam_can = CanBus(None, fingerprint).CAM
     lka_steering = 0x50 in fingerprint[cam_can] or 0x110 in fingerprint[cam_can]
     CAN = CanBus(None, fingerprint, lka_steering)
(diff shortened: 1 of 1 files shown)


opendbc/car/hyundai/interface.py
```diff
@@ -33,6 +33,10 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
     cam_can = CanBus(None, fingerprint).CAM
     lka_steering = 0x50 in fingerprint[cam_can] or 0x110 in fingerprint[cam_can]
     CAN = CanBus(None, fingerprint, lka_steering)
+    print(cam_can, lka_steering, CAN)
+
+    # assert 0x50 not in fingerprint[CAN.ACAN] and 0x110 not in fingerprint[CAN.ACAN]
+    # assert 0x50 not in fingerprint[CAN.ECAN] and 0x110 not in fingerprint[CAN.ECAN]
```
sshane: can't do this, some cars have garbage LKAS_ALT on ECAN (not real LKAS_ALT I'm pretty sure)

opendbc/car/hyundai/interface.py
```diff
@@ -58,6 +62,8 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
           ret.flags |= HyundaiFlags.CANFD_ALT_BUTTONS.value
         if not ret.flags & HyundaiFlags.RADAR_SCC:
           ret.flags |= HyundaiFlags.CANFD_CAMERA_SCC.value
+          # sanity check SCC_CONTROL isn't on E-CAN (powertrain bus)
+          ret.dashcamOnly = 0x1a0 in fingerprint[CAN.ECAN]
```
sshane: but checking if SCC_CONTROL isn't on the bus we expect is safe

Issue #2357, opened 2025-06-02
car interfaces: add dashcam reason


sshane:
Useful to show to the user all of this information: https://github.com/commaai/opendbc/blob/aefff8bed44e36d5e719a912a590428460770735/opendbc/car/ford/interface.py#L64-L65

PR #2359, opened 2025-06-02
log

sshane:
Closes https://github.com/commaai/opendbc/issues/2357

something like this


sshane:
@adeebshihadeh what do you think about using this field on the permanent dashcam alert's second line?

image

PR #2362, opened 2025-06-03
Tesla: Port HW4 Model X

ssysm:
**Checklist**

- [X] added entry to CAR in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [X] route with openpilot: [bdda168c0c35fad7/00000001--5c5a36ec06](https://connect.comma.ai/bdda168c0c35fad7/00000001--5c5a36ec06)
- [X] route with stock system: [bdda168c0c35fad7/00000002--6df4ad4fab](https://connect.comma.ai/bdda168c0c35fad7/00000002--6df4ad4fab)
- [X] car harness used (if comma doesn't sell it, put N/A): ~~Modified~~ Tesla B (CAN 1 can be ~~RADAR CAN or~~ VEH CAN)

### DIFF
--- a/opendbc/car/tesla/carstate.py
+++ b/opendbc/car/tesla/carstate.py
@@ -114,7 +114,6 @@ class CarState(CarStateBase):
     if self.CP.carFingerprint in (CAR.TESLA_MODEL_3, CAR.TESLA_MODEL_Y):
       ret.invalidLkasSetting = cp_ap_party.vl["DAS_settings"]["DAS_autosteerEnabled"] != 0
     else:
-      #TODO: Find S/X equivalent
       pass
     # Buttons # ToDo: add Gap adjust button
 
--- a/opendbc/car/tesla/interface.py
+++ b/opendbc/car/tesla/interface.py
@@ -2,7 +2,7 @@ from opendbc.car import get_safety_config, structs
 from opendbc.car.interfaces import CarInterfaceBase
 from opendbc.car.tesla.carcontroller import CarController
 from opendbc.car.tesla.carstate import CarState
-from opendbc.car.tesla.values import TeslaSafetyFlags
+from opendbc.car.tesla.values import TeslaSafetyFlags, CAR
 
 
 class CarInterface(CarInterfaceBase):
@@ -30,5 +30,7 @@ class CarInterface(CarInterfaceBase):
       ret.vEgoStopping = 0.1
       ret.vEgoStarting = 0.1
       ret.stoppingDecelRate = 0.3
+
+    ret.dashcamOnly = candidate in (CAR.TESLA_MODEL_X) # dashcam only, pending find invalidLkasSetting signal
 
     return ret


opendbc/car/tesla/carstate.py
```diff
@@ -136,10 +142,12 @@ def get_can_parsers(CP):
       ("DAS_control", 25),
       ("DAS_steeringControl", 50),
       ("DAS_status", 2),
-      ("DAS_settings", 2),
       ("SCCM_steeringAngleSensor", 100),
     ]
 
+    if is3Y:
+      ap_party_messages.append(("DAS_settings", 2))
+
```
sshane:
```suggestion
    if CP.carFingerprint in (CAR.TESLA_MODEL_3, CAR.TESLA_MODEL_Y):
      ap_party_messages.append(("DAS_settings", 2))

```

no need to abstract into a variable for something this simple

sshane: need to import CAR from values at the top

opendbc/car/tesla/values.py
```diff
@@ -55,6 +55,12 @@ class CAR(Platforms):
      ],
     CarSpecs(mass=2072., wheelbase=2.890, steerRatio=12.0),
   )
+  TESLA_MODEL_X = TeslaPlatformConfig(
+    [
+      TeslaCarDocsHW4("Tesla Model X (with HW4) 2024"),
+    ],
```
sshane:
```suggestion
    [TeslaCarDocsHW4("Tesla Model X (with HW4) 2024")],
```

opendbc/car/tesla/values.py
```diff
@@ -55,6 +55,12 @@ class CAR(Platforms):
...
+    CarSpecs(mass=2495., wheelbase=2.960, steerRatio=12.0),
```
sshane, line 60: TODO for me: check lateral accel differences using Model Y vehicle model

sshane:
Nice! This is a very small PR to add the X!

Do you mind getting new routes? The ones you added are a few months old and probably don't have logs to upload

ssysm:
I'll try to get them in the next few days!

opendbc/car/tesla/carstate.py
```diff
@@ -22,6 +22,8 @@ def __init__(self, CP):
     self.hands_on_level = 0
     self.das_control = None
 
+    self.is3Y = CP.carFingerprint in (CAR.TESLA_MODEL_3, CAR.TESLA_MODEL_Y)
+
```
sshane:
```suggestion
```

same

adeebshihadeh:
> * [x]  car harness used (if comma doesn't sell it, put N/A): ~Modified~ Tesla B (CAN 1 can be ~RADAR CAN or~ VEH CAN)

So it's a stock Tesla B or you had to swap the buses around?

ssysm:
Stock Tesla B with CAN1 on vehicle bus.

ssysm:
Updated routes

sshane:
CAN 1 should already be on vehicle bus (and I don't believe we use it currently). Did you modify the physical harness or were there no modifications?

ssysm:
I don't have the official Tesla B that comma sells (made my own harness) so I'm not sure. But the distance from the VEH bus (diagnostics port) from the AP computer is quite far so there might be a need to modify for a longer cable.

sshane:
That bus is no longer needed, you can disconnect it and see if everything still works

ssysm:
Yes it will work.

lukasloetkolben:
I can confirm that it works with the Tesla B harness on a new HW4 Model S.

opendbc/car/tesla/carstate.py
```diff
@@ -111,8 +111,11 @@ def update(self, can_parsers) -> structs.CarState:
     ret.stockLkas = cp_ap_party.vl["DAS_steeringControl"]["DAS_steeringControlType"] == 2  # LANE_KEEP_ASSIST
 
     # Stock Autosteer should be off (includes FSD)
-    ret.invalidLkasSetting = cp_ap_party.vl["DAS_settings"]["DAS_autosteerEnabled"] != 0
-
+    if self.CP.carFingerprint in (CAR.TESLA_MODEL_3, CAR.TESLA_MODEL_Y):
+      ret.invalidLkasSetting = cp_ap_party.vl["DAS_settings"]["DAS_autosteerEnabled"] != 0
+    else:
+      #TODO: Find S/X equivalent
+      pass
```
adeebshihadeh: we need to find this before merging

adeebshihadeh (commented):
LGTM aside from that comment. @sshane might have a few more comments though

ssysm:
Re the invalidLKASsetting flag: there's a very good chance that S/X's DAS settings signal is not on the party bus and I still haven't seen anything like it on Cabana when I switch it from the car's UI. I'll try to dig a bit more but is very likely not there.

adeebshihadeh:
> Re the invalidLKASsetting flag: there's a very good chance that S/X's DAS settings signal is not on the party bus and I still haven't seen anything like it on Cabana when I switch it from the car's UI. I'll try to dig a bit more but is very likely not there.

@sshane got any ideas here?

opendbc/car/tesla/carstate.py
adeebshihadeh: @ssysm let's set the `dashcamOnly` flag True for Model X and leave a comment explaining it's pending this. once that's done, I think we can merge this

ssysm: Added

adeebshihadeh:
Thanks!

sshane:
@ssysm try this: https://github.com/commaai/opendbc/issues/2854#issuecomment-3814831301

ssysm:
Thanks for the advice! I think I found an interesting bit on Addr 0x431. I'll double check tmr to see if it actually matches the TACC signal.

Issue #2369, opened 2025-06-05
macOS Apple Silicon: Cannot compile - "ld: library 'gcov' not found"


sunnyhaibin:
Introduced in https://github.com/commaai/opendbc/pull/2327

Environment: Apple Silicon Sequoia 15.1

```
gcc -o opendbc_repo/opendbc/safety/tests/libsafety/libsafety.so -dynamiclib opendbc_repo/opendbc/safety/tests/libsafety/safety.os -lgcov
gcc -o panda/tests/libpanda/panda.os -c -nostdlib -fno-builtin -std=gnu11 -Wfatal-errors -Wno-pointer-to-int-cast -fPIC -Ipanda/tests/libpanda -Ipanda/board -Iopendbc_repo panda/tests/libpanda/panda.c
ld: library 'gcov' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
scons: *** [opendbc_repo/opendbc/safety/tests/libsafety/libsafety.so] Error 1
scons: building terminated because of errors.
```

sunnyhaibin:
https://github.com/commaai/opendbc/pull/2327#issuecomment-2939169565

tddutctit:
seems need to check/update the env for gcov lib.

adeebshihadeh:
Any idea why this passes in CI? Do the CI machines have some package pre-installed?

devtekve:
> Any idea why this passes in CI? Do the CI machines have some package pre-installed?

It may be due to different osx version (?) sunny has 15.1 I have 15.5. (I have the same problem). I think I may have found a solution

devtekve:
cross posting this as a fix https://github.com/commaai/opendbc/pull/2373

andiradulescu:
> Any idea why this passes in CI? Do the CI machines have some package pre-installed?

Should be because both CI runners are running macos-14 (sonoma) - namespace 8x14 and github macos-14-arm64 - thus making it revert to using `gcc` on macos.

This should fix it https://github.com/commaai/opendbc/pull/2384

PR #2373, opened 2025-06-06
Fix macos 15.x build after adding gcov

devtekve:
This is currently working on my local setup with sequoia 15.5 

see https://github.com/commaai/opendbc/issues/2369

# Highlighting the change because git somehow don't think it's relevant to show it.
image


devtekve:
@adeebshihadeh @sshane @greatgitsby can you please look and validate this fix? I can confirm the build works as expected now on my device

PR #2374, opened 2025-06-06
fingerprint: Hyundai Tucson 2022

orsonjiang:
Validation
* Dongle ID: c972e3029b8f5d54
* Route: c972e3029b8f5d54/0000000d--a2dba036fb


jyoung8607:
FYI this isn't being ignored, but much like #2650 there's a systemic conflict between Tucson and Santa Cruz that prevents us from merging it right away. We'll keep this around until it's fixed.

PR #2375, opened 2025-06-06
Volkswagen PQ: add HCA7 as a valid HCA state for EPS (lateral control)

tealtwo:
Validation
* Dongle ID: 0f53129ed44f6920
* Route: 0f53129ed44f6920/000001f2--e55a4f4367

added HCA7 for VW PQ as a valid steering state for lateral control

### DIFF
--- a/opendbc/dbc/vw_pq.dbc
+++ b/opendbc/dbc/vw_pq.dbc
@@ -1637,7 +1637,7 @@ VAL_ 872 ACS_Fehler 1 "Fehlerspeichereintrag" 0 "kein_Fehlerspeichereintrag" ;
 VAL_ 872 ACS_zul_Regelabw 254 "ADR_nicht_aktiv" 255 "Fehler" ;
 VAL_ 872 ACS_max_AendGrad 254 "Neutralwert" 0 "Neutralwert" 255 "Fehler" ;
 
-VAL_ 978 LH2_Sta_HCA 0 "disabled" 1 "initializing" 2 "fault" 3 "ready" 4 "rejected" 5 "active";
+VAL_ 978 LH2_Sta_HCA 0 "disabled" 1 "initializing" 2 "fault" 3 "ready" 4 "rejected" 5 "active" 7 "active";
 VAL_ 1088 Waehlhebelposition__Getriebe_1_ 8 "P" 7 "R" 6 "N" 5 "D" 9 "U" 12 "S" 14 "T" 10 "T" 11 "T";
 
 VAL_ 1386 ACA_StaACC 6 "ACC_rev_aus" 0 "Hauptschalter_aus" 4 "ACC_im_Hintergrund" 3 "ACC_aktiv" 1 "Reserve" 2 "ACC_passiv" 7 "ACC_irrev_aus" 5 "frei" ;


sshane:
What does this state describe? How do you get into this state?

tealtwo:
HCA is the steering signal that openpilot is sending to steer, on MQB only status 5 is a valid HCA steering state but PQ Volkswagens have HCA7 in addition to HCA5. HCA7 has a substantial increase in torque compared to HCA5 and just needs a simple status change when being sent to the EPS (instead of 5, we send 7). A lot of the PQ Volkswagen community (almost all actually) use HCA7 and it still has not been added to the upstream Volkswagen PQ DBC, which only has 5 as a valid state for HCA. This PR is simply adding 7 as a valid HCA steering state for PQ. Here is an example of one such fork that uses HCA7: (https://github.com/tealtwo/openpilot/blob/1926a3940b82760e4ae1f182c9eca4abd470562b/selfdrive/car/volkswagen/pqcan.py#L5)

tealtwo:
> What does this state describe? How do you get into this state?

HCA is the steering signal that openpilot is sending to steer, on MQB only status 5 is a valid HCA steering state but PQ Volkswagens have HCA7 in addition to HCA5. HCA7 has a substantial increase in torque compared to HCA5 and just needs a simple status change when being sent to the EPS (instead of 5, we send 7). A lot of the PQ Volkswagen community (almost all actually) use HCA7 and it still has not been added to the upstream Volkswagen PQ DBC, which only has 5 as a valid state for HCA. This PR is simply adding 7 as a valid HCA steering state for PQ. Here is an example of one such fork that uses HCA7: (https://github.com/tealtwo/openpilot/blob/1926a3940b82760e4ae1f182c9eca4abd470562b/selfdrive/car/volkswagen/pqcan.py#L5)

sshane:
This seems intentional then. @jyoung8607 should we wait to add this until we use this new mode?

tealtwo:
> This seems intentional then. @jyoung8607 should we wait to add this until we use this new mode?

I don't see a reason to wait to add this to upstream, its already widely used and well known.

dparring:
This is the best explanation that I've seen

https://discord.com/channels/469524606043160576/534359517836607488/785545653521612811
"we have information that calls them HCA modes 1 vs 2, and that the second mode ignores or bypasses self-centering"

dparring:
Also noting that this is a functional mode on the MLB platform as well

tealtwo:
> This is the best explanation that I've seen
> 
> https://discord.com/channels/469524606043160576/534359517836607488/785545653521612811 "we have information that calls them HCA modes 1 vs 2, and that the second mode ignores or bypasses self-centering"

The removal of the self centering is what allows the massive increase in torque output:)

opendbc/dbc/vw_pq.dbc
```diff
@@ -1637,7 +1637,7 @@ VAL_ 872 ACS_Fehler 1 "Fehlerspeichereintrag" 0 "kein_Fehlerspeichereintrag" ;
 VAL_ 872 ACS_zul_Regelabw 254 "ADR_nicht_aktiv" 255 "Fehler" ;
 VAL_ 872 ACS_max_AendGrad 254 "Neutralwert" 0 "Neutralwert" 255 "Fehler" ;
 
-VAL_ 978 LH2_Sta_HCA 0 "disabled" 1 "initializing" 2 "fault" 3 "ready" 4 "rejected" 5 "active";
+VAL_ 978 LH2_Sta_HCA 0 "disabled" 1 "initializing" 2 "fault" 3 "ready" 4 "rejected" 5 "active" 7 "active";
```
jyoung8607, line 1640: We know that both 5 and 7 are valid for the command channel, `HCA_1.HCA_Status`. Do we know for certain the EPS rack reflects that value in `Lenkhilfe_2.LH2_Sta_HCA` when in use? It's been a long time since I've looked at data for this mode.

jyoung8607:
> This seems intentional then. @jyoung8607 should we wait to add this until we use this new mode?

Pending confirmation of my review comment, we can add it to the DBC. I don't expect to switch openpilot to use it unless and until we have radical improvements in our ability to control it. It's not dangerous, it's just been difficult to tune.

opendbc/dbc/vw_pq.dbc
dparring: Does this answer?  
image
Route: e8dac44d3a3458a5/00000094--cc9fdeb8fb

tealtwo: 7 is a valid HCA Status for `Lenkhilfe_2.LH2_Sta_HCA `, it’s also shown in the route linked in the initial PR. Openpilot should remain on HCA5 as you said, 7 is still extremely difficult to tune.

PR #2381, opened 2025-06-08
Honda: Car Port for Thai HR-V 2023 (3G)

AlexandreSato:
Validation
* Dongle ID: 1e4baee1aa2687a0
* Route: 
* User: somkanae#5442

reference [PR #27915](https://github.com/commaai/openpilot/pull/27915/commits)

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -85,6 +85,8 @@ class CarState(CarStateBase):
       self.gearbox_msg = "GEARBOX_15T"
     elif CP.carFingerprint == CAR.HONDA_CIVIC_2022 and CP.transmissionType == TransmissionType.cvt:
       self.gearbox_msg = "GEARBOX_ALT"
+    elif CP.carFingerprint == CAR.HONDA_HRV_3G and CP.flags & HondaFlags.BOSCH_ALT_GEARS:
+      self.gearbox_msg = "GEARBOX_ALT"
     elif CP.transmissionType == TransmissionType.manual:
       self.gearbox_msg = "GEARBOX_ALT_2"
 
--- a/opendbc/car/honda/fingerprints.py
+++ b/opendbc/car/honda/fingerprints.py
@@ -794,17 +794,21 @@ FW_VERSIONS = {
     (Ecu.eps, 0x18da30f1, None): [
       b'39990-3M0-G110\x00\x00',
       b'39990-3W0-A030\x00\x00',
+      b'39990-3M0-J030\x00\x00',
     ],
     (Ecu.gateway, 0x18daeff1, None): [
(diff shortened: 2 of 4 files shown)


vanillagorillaa:
I think there is an easier way to do this. You can get rid of setting the safety flag in values.py, change interface to put the checks with the rest of the gearbox message checks (lines 69-77), set this car transmission as a cvt, then you can just include this vehicle with the Civic gearbox check in carstate.py

sshane:
Is this ready for review? If the under braking stopping behavior still happens, can you leave a simple comment about this model of HR-V?

AlexandreSato:
> Is this ready for review? If the under braking stopping behavior still happens, can you leave a simple comment about this model of HR-V?

Last changes suggested by VanilaGorila not tested yet. That braking behavior only happens in the ACC stock of the Brazilian model. Op long doesn't happen. 
What sonkanae was getting was HDC errors (descent control) but from what I understood from what he wrote in discord is that with a new official harness the HDC errors stopped
IMG_2606

somkanae:
> > Is this ready for review? If the under braking stopping behavior still happens, can you leave a simple comment about this model of HR-V?
> 
> Last changes suggested by VanilaGorila not tested yet. That braking behavior only happens in the ACC stock of the Brazilian model. Op long doesn't happen. What sonkanae was getting was HDC errors (descent control) but from what I understood from what he wrote in discord is that with a new official harness the HDC errors stopped IMG_2606

Three years ago, I bought a comma 3. The Honda Bosch B harness wasn't officially out yet, so I re-pinned a comma harness myself. It worked, but I kept getting a hill descent error. Once comma released the official Honda Bosch B harness, I bought it, and I haven't had any issues since.

opendbc/car/honda/carstate.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, CP):
     self.gearbox_msg = "GEARBOX"
     if CP.carFingerprint == CAR.HONDA_ACCORD and CP.transmissionType == TransmissionType.cvt:
       self.gearbox_msg = "GEARBOX_15T"
-    elif CP.carFingerprint == CAR.HONDA_CIVIC_2022 and CP.transmissionType == TransmissionType.cvt:
+    elif CP.carFingerprint in (CAR.HONDA_CIVIC_2022, CAR.HONDA_HRV_3G) and CP.transmissionType == TransmissionType.cvt:
```
sshane, line 86: just check if radarless?

AlexandreSato: The American one (north and south) is radarless too but uses a different gear box signal.

jyoung8607: I want to overhaul all the signal selection and detection stuff anyway.

PR #2384, opened 2025-06-09
safety: fix coverage on macos

andiradulescu:
Fixes https://github.com/commaai/opendbc/issues/2369

Tested on macos-15, with both `clang` and `gcc` and on Ubuntu.

`-fprofile-arcs` and `-ftest-coverage` are supported in `clang` just that you need to pass the flags also to the linker. same for `gcc`, the flags need to be passed to the linker too.

as linker flags, `-fprofile-arcs` seems to solve the linking problem for `gcc` and `-ftest-coverage` for `clang`.

Some reading:
- gcc - https://gcc.gnu.org/onlinedocs/gcc-13.4.0/gcc/Instrumentation-Options.html
- clang - https://clang.llvm.org/docs/SourceBasedCodeCoverage.html


devtekve:
@andiradulescu i had a fix already for this https://github.com/commaai/opendbc/pull/2373

andiradulescu:
@devtekve I know about your fix, but it defaults “all” macos versions to `gcc`. also, when macos-16 is released, it will still not work since “gcov” is not available in clang on macos, so you will need to default again to `gcc`.

devtekve:
> @devtekve I know about your fix, but it defaults “all” macos versions to `gcc`. also, when macos-16 is released, it will still not work since “gcov” is not available in clang on macos, so you will need to default again to `gcc`.

Wouldn't have hurt mentioning it there lmao

andiradulescu:
> Wouldn't have hurt mentioning it there lmao

I’ll mention it next time. Seemed simpler to fix it, than explaining.

PR #2389, opened 2025-06-11
Car docs: fix year test

sshane:
We've been weighting what to fail during runtime and what to fail during tests: Causing an exception during run time is too much for a misformatted docs string, better to keep in the test for now


sshane:
site borken now fix:
image

PR #2393, opened 2025-06-12
fingerprint: VW Golf MK7 2014

mickrideout:
Validation
* Dongle ID:  a3c1d7ae126d02aa
* Route: a3c1d7ae126d02aa/00000001--dcdca33466


adeebshihadeh:
@sshane @jyoung8607 don't we use the VIN for VW? why are we getting FP PRs still?

sshane:
There is no fwdCamera ECU to get a VIN from, @jyoung8607 I forgot how we chose to proceed for these cars? These need some aftermarket programming, correct?

Did we think it's fine if they don't automatically fingerprint?

jyoung8607:
> @sshane @jyoung8607 don't we use the VIN for VW?

Yes. VIN plus any (across all MQB models) known radar firmware as a sanity check.

> why are we getting FP PRs still?

Generally we don't.

I checked his device. All of his uploaded routes are getting good UDS firmware results but returning no VIN, so no fuzzy VIN match. I've not seen this before, it's puzzling.

Let's keep this open pending more data. @mickrideout is there anything unusual about your car or your openpilot setup we should know? Also, can you upload rlogs for one of your routes, and let us know your VIN? Ultimately we'll accept your PR, but you shouldn't have had to make one, and I want to take a closer look.

jyoung8607:
> There is no fwdCamera ECU to get a VIN from, @jyoung8607 I forgot how we chose to proceed for these cars? These need some aftermarket programming, correct?

I don't know yet if this is a retrofit situation, but even if that's the case, those should still fuzzy match just fine.

mickrideout:
This car isnt a retofit. It is a 2014 Golf with no forward camera. The OP is a fork from yesterday:  
- https://github.com/mickrideout/openpilot
- https://github.com/mickrideout/opendbc  

The VIN for the car is WVWZZZAUZEW411115  
The rlogs for the last drive are here - https://drive.google.com/file/d/1C7QNj77EFq6j-Rwv_hlWzYh2N1_zF8OM/view?usp=drive_link  

I had previously owned a 2013 VW Golf in the UK with ACC and no LKAS but it did have a forward camera for the "Front Assist" system for forward collision warning, it didnt require fingerprinting on OP v0.8.13. Im now in Australia and this car has ACC, no LKAS and no forward camera. This is the only difference I can think of between the two cars.

sshane:
That tracks with the lack of VIN, as we use the camera to get it without comma power

PR #2395, opened 2025-06-15
Car Port: Honda City (Brazil) 

vanillagorillaa:
Validation
* Dongle ID: 56b2cf1dacdcd033
* Route: TODO


AlexandreSato:
@vanillagorillaa We found the handbrake signal on #2401, however in the Brazilian market we have both the version with electronic parkbrake (and cruise available up to speed 0) and this one which is old manual and the cruise only starts working from a certain speed.

vanillagorillaa:
Added in handbrake signal found in https://github.com/commaai/opendbc/pull/2401

PR #2397, opened 2025-06-16
Toyota: Add FW for 2025 Lexus LC (Taiwan)

nelsonjchen:
Validation
* Dongle ID: 3feba96d337138a6
* Route: 3feba96d337138a6/00000000--df1c721dcd


nelsonjchen:
Route added

nelsonjchen:
Need to edit the docs, hah, that was the point.

jyoung8607:
Merging with a caveat: the lat accel factor isn't a good match. Since there's a good chance it's wrong for all TSS2 Lexus LC, I've opened #2426 to come back and fix it later.

PR #2401, opened 2025-06-16
Honda City 2023 Handbrake signal

AlexandreSato:
Prerequisite for https://github.com/commaai/opendbc/pull/2395

In Brazil we have Honda City with and without electronic handbrake, this signal was validated for the manual version

* Route: 56b2cf1dacdcd033/0000002a--803651215c/3
image


AlexandreSato:
We found the handbrake signal on #2401, however in the Brazilian market we have both the version with electronic parkbrake (and cruise available up to speed 0) and this one which is old manual and the cruise only starts working from a certain speed.

PR #2406, opened 2025-06-18
Ford: Make CarDocs obvious

ugtthis:
- Making the implicit explicit
- All other brands make models explicit but then you get to `Ford` and models get added during runtime
- Removed `hybrid` / `plug_in_hybrid` flag since it was only used for doc purposes unlike how `Hyundai` uses the `hybrid` flags


adeebshihadeh:
I'm supportive of this. Agree @sshane?

PR #2410, opened 2025-06-25
Honda: Acura MDX 2022 - 2024

mvl-boston:
Support for 2022 through 2024 Acura MDX.

As Honda's flagship vehicle, it is about 75% Bosch A code, but has introduced some early components of Bosch C cars. 

- Has different lowspeed lockout steer status value
- Removed alt-radar flag LKAS logic, caused flashing LKAS alerts on HUD
- Included torque controller
- Unique 8 byte LKAS_HUD in Bosch A required a safety change
- Max steer torque at 3840 per dashcam 619b464263ab23f2/0000000a--842d4a200a
- ALT_BRAKE variants use the canFD cruise fault message - confirmed 1be0211e5e78d925/00000004--fc96beb9b0/7

Additional test routes:
619b464263ab23f2/00000026--9cfb042d6a
619b464263ab23f2/00000028--c8c780d406

Todo:
- [x] Test that driving logic works
- [x] get test route and adjust torque parameters
- [x] Get dashcam route to confirm standard LKAS steer max
- [x] Force a cruise fault on iBooster to prove message bit works same as canfd cars


mvl-boston:
Max stock steering route added - this should be ready to go.

opendbc/car/honda/carstate.py
```diff
@@ -122,7 +124,10 @@ def update(self, can_parsers) -> structs.CarState:
       ret.accFaulted = bool(cp.vl["CRUISE_FAULT_STATUS"]["CRUISE_FAULT"])
     else:
       if self.CP.openpilotLongitudinalControl:
-        ret.accFaulted = bool(cp.vl[self.brake_error_msg]["BRAKE_ERROR_1"] or cp.vl[self.brake_error_msg]["BRAKE_ERROR_2"])
+        if (self.CP.carFingerprint in HONDA_BOSCH_TJA_CONTROL) and (self.CP.flags & HondaFlags.BOSCH_ALT_BRAKE):
+          ret.accFaulted = bool(cp.vl["BRAKE_MODULE"]["CRUISE_FAULT"])
+        else:
+          ret.accFaulted = bool(cp.vl[self.brake_error_msg]["BRAKE_ERROR_1"] or cp.vl[self.brake_error_msg]["BRAKE_ERROR_2"])
```
jyoung8607: I'm unclear why we're gating the BOSCH_ALT_BRAKE source of truth for accFaulted behind the TJA flag, can you elaborate?

opendbc/dbc/generator/honda/acura_mdx_2022_can.dbc
```diff
@@ -0,0 +1,7 @@
+CM_ "IMPORT _honda_common.dbc";
+CM_ "IMPORT _bosch_2018.dbc";
+CM_ "IMPORT _bosch_radar_acc.dbc";
+CM_ "IMPORT _lkas_hud_8byte.dbc";
+CM_ "IMPORT _bosch_standstill.dbc";
+CM_ "IMPORT _steering_sensors_a.dbc";
+CM_ "IMPORT _gearbox_common.dbc";
```
jyoung8607, line 7: This new generator template is identical to honda_common_canfd, let's not duplicate... we can rename the common_canfd DBC if we need to.

opendbc/car/honda/carstate.py
```diff
@@ -97,7 +98,8 @@ def update(self, can_parsers) -> structs.CarState:
     ret.seatbeltUnlatched = bool(cp.vl["SEATBELT_STATUS"]["SEATBELT_DRIVER_LAMP"] or not cp.vl["SEATBELT_STATUS"]["SEATBELT_DRIVER_LATCHED"])
 
     steer_status = self.steer_status_values[cp.vl["STEER_STATUS"]["STEER_STATUS"]]
-    ret.steerFaultPermanent = steer_status not in ("NORMAL", "NO_TORQUE_ALERT_1", "NO_TORQUE_ALERT_2", "LOW_SPEED_LOCKOUT", "TMP_FAULT")
+    ret.steerFaultPermanent = steer_status not in ("NORMAL", "NO_TORQUE_ALERT_1", "NO_TORQUE_ALERT_2", "LOW_SPEED_LOCKOUT", "TJA_LOW_SPEED_LOCKOUT",
+                                                   "TMP_FAULT")
```
jyoung8607: Let's do this in a separate PR, it has the potential to change behavior for other cars, and I'd like to scan it against commaCarSegments.

opendbc/car/honda/carstate.py
```diff
@@ -107,7 +109,7 @@ def update(self, can_parsers) -> structs.CarState:
       # LOW_SPEED_LOCKOUT is not worth a warning
       # NO_TORQUE_ALERT_2 can be caused by bump or steering nudge from driver
       # FIXME: the stock camera stops steering on NO_TORQUE_ALERT_1
-      ret.steerFaultTemporary = steer_status not in ("NORMAL", "LOW_SPEED_LOCKOUT", "NO_TORQUE_ALERT_2")
+      ret.steerFaultTemporary = steer_status not in ("NORMAL", "LOW_SPEED_LOCKOUT", "TJA_LOW_SPEED_LOCKOUT", "NO_TORQUE_ALERT_2")
```
jyoung8607: Same comment as steerFaultPermanent

opendbc/car/honda/interface.py
```diff
@@ -206,7 +206,8 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
     # These cars use alternate user brake msg (0x1BE)
-    if 0x1BE in fingerprint[CAN.pt] and candidate in (CAR.HONDA_ACCORD, CAR.HONDA_HRV_3G, CAR.ACURA_RDX_3G, *HONDA_BOSCH_CANFD):
+    if 0x1BE in fingerprint[CAN.pt] and candidate in (CAR.HONDA_ACCORD, CAR.HONDA_HRV_3G, CAR.ACURA_RDX_3G, CAR.ACURA_MDX_4G,
+                                                      *HONDA_BOSCH_CANFD):
```
jyoung8607, line 209: Let's scan commaCarSegments and see if we can just take away the candidate list entirely

opendbc/car/honda/values.py
```diff
@@ -116,6 +116,9 @@ class Footnote(Enum):
   CIVIC_DIESEL = CarFootnote(
     "2019 Honda Civic 1.6L Diesel Sedan does not have ALC below 12mph.",
     Column.FSR_STEERING)
+  TRAFFIC_JAM_ASSIST = CarFootnote(
+    "ALC is supported below 45mph only when following a lead car.",
+    Column.FSR_STEERING)
```
jyoung8607: What exactly are the behavior differences if the driver dips below 45mph without a lead car? Does steering just silently stop working, or is there intended to be a warning, or.. ?

jyoung8607 (changes requested):

opendbc/car/honda/carstate.py
mvl-boston: This triggers often per driver logs so car is unsteerable after most stops if this is not in steerfaultpermanent.

Is commacarsegments internal and bigger than public huggingface?  Was planning to look at juniper notebooks on the flight.

I could gate exemption to this car if research will take a while.

opendbc/car/honda/values.py
mvl-boston: Stock stops steering per owners manual, likely lanelines change color from green to white to indicate lack of steer, as that is the normal stock 8-bit-LKAS-HUD experience below factory speed.  Drivers stated experience is that steering does not occur once a leadcar leaves your lane.

Need to see first hand when I land and get the car.

opendbc/car/honda/carstate.py
mvl-boston: Other alt_brake cars were untested so the gating was to prevent regressions.  Can't pull from carsegments as you need to deliberately fault each car to test.

This car behaves similar to CANFD long altbrake (not upstreamed) where the brake_error_msg bits are routinely set in non-error situations so cannot be used.  Instead, you need to use the brake_module message.  Same gating was used in CanFD long PR for same reason.

opendbc/dbc/generator/honda/acura_mdx_2022_can.dbc
mvl-boston: Done

opendbc/car/honda/values.py
mvl-boston: Confirmed: Stock goes from solid green lkas lanes to hollow white lkas lanes when the leadcar leaves the lane or is too far to be , and consistent with other Hondas stock stops steering.

On stock, there is a configurable beep for both losing lkas and losing leadcar, car settings can turn either on or off.

Will test current PR once it finishes installing.

opendbc/car/honda/hondacan.py
```diff
@@ -191,7 +191,7 @@ def create_lkas_hud(packer, bus, CP, hud_control, lat_active, steering_available
 
   # New HUD concept for selected Bosch cars, overwrites some of the above
   # TODO: make global across all Honda if feedback is favorable
-  if CP.carFingerprint in HONDA_BOSCH_ALT_RADAR:
+  if CP.carFingerprint == CAR.HONDA_ODYSSEY:
     lkas_hud_values['DASHED_LANES'] = steering_available
     lkas_hud_values['SOLID_LANES'] = lat_active
     lkas_hud_values['LKAS_PROBLEM'] = lat_active and reduced_steering
```
jyoung8607: See what kind of experience you have if you just remove the LKAS_PROBLEM line. That's the part that was kind of questionable and it's not very important.

jeraldhu-yuan:
I'm evaluating Traffic Jam Assist for `ACURA_MDX_4G` (2022–24). Commit `550fa584` in this PR removed `BOSCH_TJA_CONTROL`; MVL's current Honda release and development branches enable it for this platform.

Was the MDX-specific removal prompted by a functional problem, the lead-vehicle dependency / warning behavior, or incomplete validation? I found the earlier broad Honda rollback in #2240 and the per-platform restoration in #2818, but those don't explain this MDX-specific decision.

For the current MDX implementation, are there known requirements or failure cases involving stock ACC versus lateral-only operation, lead loss/reacquisition, or particular EPS versions? Any existing test evidence or discussion would help. The candidate has only been tested offline; it has not been enabled on the vehicle.

mvl-boston:
> I'm evaluating Traffic Jam Assist for `ACURA_MDX_4G` (2022–24). Commit `550fa584` in this PR removed `BOSCH_TJA_CONTROL`; MVL's current Honda release and development branches enable it for this platform.
> 
> Was the MDX-specific removal prompted by a functional problem, the lead-vehicle dependency / warning behavior, or incomplete validation? I found the earlier broad Honda rollback in #2240 and the per-platform restoration in #2818, but those don't explain this MDX-specific decision.
> 
> For the current MDX implementation, are there known requirements or failure cases involving stock ACC versus lateral-only operation, lead loss/reacquisition, or particular EPS versions? Any existing test evidence or discussion would help. The candidate has only been tested offline; it has not been enabled on the vehicle.

I was hoping to find a way to always get it to steer under 45mph but wasn't able to figure it out when I had the rental.   If you can get a few logs of under 45mph steer I can try to build a few test forks remotely, Then we can submit PR for it.

PR #2411, opened 2025-06-25
Mdx 2022 debug steer - for CI

mvl-boston:
Validation
* Dongle ID: 
* Route:


mvl-boston:
mistake - opened in wrong repo

PR #2413, opened 2025-06-26
Fix CI tests

devtekve:
Fixes 

image


opendbc/car/body/fingerprints.py
```diff
@@ -1,4 +1,3 @@
-# ruff: noqa: E501
```
sshane, line 1: these files are autoformatted, so this will come back next time the FP bot runs. you need to edit format_fingerprints.py

PR #2421, opened 2025-06-28
port: (2025 Ioniq5) HKG Angle Steering LFA2

devtekve:
> [!NOTE]  
> **Depends on** 
> * https://github.com/commaai/opendbc/pull/2420
> * https://github.com/commaai/opendbc/pull/2422
> * https://github.com/commaai/opendbc/pull/2503

This PR introduces the **minimum required changes to safely support angle-based steering control on HKG vehicles**, starting with the 2025 Hyundai Ioniq 5, which I personally own. This vehicle serves as the initial reference platform for the “official” port.

Although the current implementation works across all HKG angle steering vehicles **_we've tested so far_**, additional safety-related data needs to be gathered on a per-model and per-region basis. Merging support for this vehicle now enables faster iteration and testing for the remaining variants.

---

### ✅ What’s Included

- Lateral control via angle steering for HKG platform.
- Safety-critical enforcement of steering angle rate limits, maximum lateral acceleration, and jerk.

---

### ❓ Remaining Unknowns

We do **not yet fully understand how the `Torque Reduction Gain` setting behaves** across different HKG vehicles. This parameter, used by the EPS controller, appears to differ by model and region (so far US vs. Rest of the World). While it does not affect core safety constraints, it may influence **driver comfort and perceived steering responsiveness**.

- The **safety code assumes that requested steering angles are applied immediately**, regardless of the vehicle's internal smoothing or delays. Therefore, “Torque Reduction Gain” does **not** factor into the safety model for enforcing lateral acceleration or jerk limits.
- Despite this, the parameter may still influence how the EPS modulates assist during continuous control or how it blends driver and system input.

We could implement additional checks to ensure the driver can always override the system, but this might be redundant: **the EPS controller already disables assist when the driver applies more than approximately 10 Nm of torque**, ensuring manual control takes precedence when needed, although requiring significant override force. 

---

### 🔍 Observations on Torque Reduction Gain

This parameter has been observed to fluctuate during live drives with Lane Following Assist (LFA) enabled. It tends to settle to a steady value when driving on straight roads, but:

- For **EU Ioniq 5 models**, the steady-state value has consistently been **0.384**.
- For **US Ioniq 5 models**, we’ve observed it stabilizing around **0.92**.
- The maximum observed value is **1.0**.

These differences suggest regional or model-specific EPS tuning.

# Testing
* Route: `e1107f9d04dfb1e2/000003a6--950f126991`

## Plotjuggler 
image
image

## Video
https://github.com/user-attachments/assets/014f71bb-bdf7-45d5-a935-8bdeadae221e

### DIFF
--- a/opendbc/safety/tests/test_hyundai_canfd.py
+++ b/opendbc/safety/tests/test_hyundai_canfd.py
@@ -37,7 +37,7 @@ def round_angle(angle_deg: float, can_offset=0):
 class TestHyundaiCanfdBase(HyundaiButtonBase, common.CarSafetyTest, common.DriverTorqueSteeringSafetyTest, common.SteerRequestCutSafetyTest):
 
   TX_MSGS = [[0x50, 0], [0x1CF, 1], [0x2A4, 0]]
-  STANDSTILL_THRESHOLD = 12 # 0.375 kph
+  STANDSTILL_THRESHOLD = 12  # 0.375 kph
   FWD_BLACKLISTED_ADDRS = {2: [0x50, 0x2a4]}
 
   MAX_RATE_UP = 2


opendbc/car/torque_data/override.toml
```diff
@@ -86,3 +86,6 @@ legend = ["LAT_ACCEL_FACTOR", "MAX_LAT_ACCEL_MEASURED", "FRICTION"]
 "HONDA_CIVIC_2022" = [2.5, 1.2, 0.15]
 "HONDA_HRV_3G" = [2.5, 1.2, 0.2]
 "HONDA_PILOT_4G" = [1.0, 1.0, 0.2]
+
+# Hyundai/Kia/Genesis angle control (do we really need this?)
```
adeebshihadeh: yes, needs to be defined for all cars regardless of control method

opendbc/car/torque_data/override.toml
```diff
@@ -86,3 +86,6 @@ legend = ["LAT_ACCEL_FACTOR", "MAX_LAT_ACCEL_MEASURED", "FRICTION"]
...
+"HYUNDAI_IONIQ_5_PE" = [3.172929, 3.5, 0.096019]
```
adeebshihadeh: are these numbers real? they seem way too high

adeebshihadeh: be super careful with these new hyundais. their stock systems will exceed the ISO11270 limits, and we definitely don't want to do that in openpilot

opendbc/car/hyundai/carcontroller.py
```diff
@@ -16,6 +19,21 @@
 MAX_ANGLE_FRAMES = 89
 MAX_ANGLE_CONSECUTIVE_FRAMES = 2
 
+MAX_ANGLE_RATE = 5
+# Add extra tolerance for average banked road since safety doesn't have the roll
+AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
+MAX_LATERAL_ACCEL = ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)  # ~3.6 m/s^2
+MAX_LATERAL_JERK = 3.0 + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)  # ~3.6 m/s^3
+
```
adeebshihadeh: given what we've seen the new Hyundai EPS's do, I'd rather limit the worst case to the ISO limit. the Tesla EPS actually limits exactly to the ISO limits itself

opendbc/car/hyundai/carcontroller.py
```diff
@@ -16,6 +19,21 @@
...
+
+def get_max_angle_delta(v_ego_raw: float, VM: VehicleModel, freq=100.):
+  max_curvature_rate_sec = MAX_LATERAL_JERK / (v_ego_raw ** 2)  # (1/m)/s
+  max_angle_rate_sec = math.degrees(VM.get_steer_from_curvature(max_curvature_rate_sec, v_ego_raw, 0))  # deg/s
+  return max_angle_rate_sec / float(freq) # hz
```
adeebshihadeh: let's start by pulling out anything that's common with the Tesla and other angle cars. this PR is too big to read through as-is

adeebshihadeh (changes requested):

opendbc/car/torque_data/override.toml
devtekve: The numbers were taken from params.toml for `ioniq_5` except the `MAX_LAT_ACCEL_MEASURED` which I got by seeing the max lat the stock system was doing but let's revisit this, I see all angle vehicles have NaN here so i wonder if I should even put values here or not

opendbc/car/hyundai/carcontroller.py
devtekve: Sounds good, will do it and adjust

opendbc/car/hyundai/carcontroller.py
devtekve: * https://github.com/commaai/opendbc/pull/2503

opendbc/car/hyundai/carcontroller.py
devtekve: @adeebshihadeh I updated this to be 80% of that calculation given that the calculation is pretty much taken from here https://github.com/commaai/opendbc/blob/master/opendbc/safety/lateral.h#L298-L300 ; is this good enough? 

(for clarity, below is the change)
```py
MAX_LATERAL_ACCEL = (ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)) * 0.8  # ~2,88 m/s^2; limit to 80% of pada safety.
MAX_LATERAL_JERK = (3.0 + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)) * 0.8  # ~2,88 m/s^3; limit to 80% of pada safety.
```

devtekve:
_Tests currently failing because the limits are lower in py than they are in panda safety, and we test that we either blocked if above the limits, but the limits are not reached because python has them lower. This will be addressed better once https://github.com/commaai/opendbc/pull/2503 is merged and we can provide an arbitrary limit to the method and assess panda safety (AKA: Testing if there would be a bug or a miscalculation that would request a higher than accepted by ISO steering command)_

opendbc/safety/tests/test_hyundai_canfd.py
```diff
@@ -82,7 +94,203 @@ def _button_msg(self, buttons, main_button=0, bus=None):
...
+  # Angle control limits
+  STEER_ANGLE_MAX = 180  # deg
+  DEG_TO_CAN = 10
+  MAX_LATERAL_ACCEL=(ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL))
+  MAX_LATERAL_JERK=(3.0 + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL))
+
+  # Hyundai uses get_max_angle_delta and get_max_angle for real lateral accel and jerk limits
+  # TODO: integrate this into AngleSteeringSafetyTest
+  ANGLE_RATE_BP = None
+  ANGLE_RATE_UP = None
```
sshane, line 142: This to do should be completed before we merge any more hacks

opendbc/safety/modes/hyundai_canfd.h
```diff
@@ -153,16 +162,41 @@ static bool hyundai_canfd_tx_hook(const CANPacket_t *msg) {
...
+    .frequency = 100U,
+  };
+
+  // IONIQ 5 PE values.
+  const AngleSteeringParams HYUNDAI_STEERING_PARAMS = {
+    .slip_factor = -0.0008688329819908074,  // calc_slip_factor(VM)
+    .steer_ratio = 14.26,
+    .wheelbase = 2.97,
+  };
+
```
sshane, line 177: The problem with expanding this to other brands is now it's possible for two angle Hyundais to fingerprint as each other with different steering ratios. Meanwhile all Teslas we've seen have nearly the same vehicle params. This can lead to more or less max lateral acceleration than intended.

This is the exact same problem as the alt torque limits. We need some way to detect steering ratio over CAN, or some unique identifier.

devtekve: For this specifically I have a stop gap solution, unit tests that check if a newly Hyundai flagged as angle steering would be more permissive than the current baseline model. 

This way at least every new car added is checked and we lower the limits for all. This is something I have on SP since there I have support for more than just ioniq 5.

I feel this is the “best” we can do without implementing fingerprinting on panda

devtekve: This is also the reason why we artificially lower to 80% of the limits, so that we are for the most part below the actual safety limits

sshane: I see. There's probably some unique identifier we can use. We can accept something that maps to a platform at this point, doesn't need to have the exact vehicle params we need.

sshane: We're going to need this to continue supporting Hyundais since they're all becoming so different.

devtekve: To be honest, nothing I've found give me the confidence to single positively ID a model, `DATC_SwVerX` may come close (see about it on the description of this closed PR https://github.com/commaai/opendbc/pull/1676) 

But this cars have OTA updates, so I wouldn't be surprised if a car received one of those and the values are changed, and neither im positive that all different models would have different values... 

BTW, this is how the test I mentioned looks, this happened when we were adding the Santa Fe 
image
https://github.com/sunnypilot/opendbc/actions/runs/16521014418/job/46722636940?pr=199

and this is the code for the test https://github.com/sunnypilot/opendbc/blob/c41b0af0ea8242d9da79f6182c01672e8b9ed9bf/opendbc/safety/tests/test_hyundai_canfd.py#L329-L356

opendbc/safety/tests/test_hyundai_canfd.py
```diff
@@ -22,10 +28,16 @@
...
+  scaled = int(angle_deg / 0.1)
+  scaled += can_offset
+  return scaled * 0.1
+
+
 class TestHyundaiCanfdBase(HyundaiButtonBase, common.CarSafetyTest, common.DriverTorqueSteeringSafetyTest, common.SteerRequestCutSafetyTest):
 
   TX_MSGS = [[0x50, 0], [0x1CF, 1], [0x2A4, 0]]
-  STANDSTILL_THRESHOLD = 12  # 0.375 kph
+  STANDSTILL_THRESHOLD = 0.375 * 0.03125  # kph
```
devtekve:
```suggestion
  STANDSTILL_THRESHOLD = 12 # 0.375 kph
```

opendbc/safety/tests/test_hyundai_canfd.py
```diff
@@ -22,10 +28,16 @@
...
+  scaled = int(angle_deg / 0.1)
+  scaled += can_offset
+  return scaled * 0.1
+
+
 class TestHyundaiCanfdBase(HyundaiButtonBase, common.CarSafetyTest, common.DriverTorqueSteeringSafetyTest, common.SteerRequestCutSafetyTest):
 
   TX_MSGS = [[0x50, 0], [0x1CF, 1], [0x2A4, 0]]
-  STANDSTILL_THRESHOLD = 12  # 0.375 kph
+  STANDSTILL_THRESHOLD = 12 # 0.375 kph
```
devtekve:
```suggestion
  STANDSTILL_THRESHOLD = 12  # 0.375 kph
```

PR #2428, opened 2025-07-02
Add support for generating clang compilation database by default

devtekve:
> [!NOTE]
> Sister PRs:
> * https://github.com/commaai/panda/pull/2225
> * https://github.com/commaai/openpilot/pull/35629

This brings the `compilation_db` functionality to **opendbc**, similar to what we already have in **openpilot**, which helps with IDE integration and better code navigation.

_This feature was originally introduced in openpilot via [commaai/openpilot#19533](https://github.com/commaai/openpilot/pull/19533) by @adeebshihadeh._

Previously, generating `compile_commands.json` in opendbc required passing the `--compile-db` flag. We've now improved this:

### What's changed
- **`compile_commands.json` is now always generated**—no need to pass a flag.
- Whether you build opendbc standalone or as part of openpilot, the compilation database is created automatically.

### Behavior
1. **Standalone Build**  
   When building opendbc on its own, a `compile_commands.json` will be created in the `opendbc/` folder containing only opendbc-related entries.

2. **Submodule in openpilot**  
   When opendbc is built as a submodule of openpilot, its entries will be automatically added to the top-level `compile_commands.json` in the openpilot root directory. This is done by updating the build environment in `opendbc/SConscript`.

This ensures that IDEs and static analysis tools work seamlessly across the full openpilot + opendbc codebase.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -34,10 +34,6 @@ AddOption('--mutation',
           action='store_true',
           help='generate mutation-ready code')
 
-AddOption('--compile_db',
-          action='store_true',
-          help='build clang compilation database')
-
 ccflags_asan = ["-fsanitize=address", "-fno-omit-frame-pointer"] if GetOption('asan') else []
 ldflags_asan = ["-fsanitize=address"] if GetOption('asan') else []
 
@@ -67,8 +63,7 @@ env = Environment(
   tools=["default", "cython", "compilation_db"]
 )
 
-if GetOption('compile_db'):
-  env.CompilationDatabase('compile_commands.json')
+env.CompilationDatabase('compile_commands.json')
 
 common = ''
 Export('env', 'arch', 'common')


adeebshihadeh:
Can we just do this all the time (i.e. remove the optional flag)? How much time would it add to the build?

devtekve:
> Can we just do this all the time (i.e. remove the optional flag)? How much time would it add to the build?

@adeebshihadeh  Yes, we can, and I think it’s negligible. It already has to build, it’s just now being asked to write down the “instructions” sort of speak. I’m down for having it always on and I can do the PRs for all the projects that currently have the flag

adeebshihadeh:
Yeah, let's do that then

devtekve:
updated @adeebshihadeh and I've created PRs for panda and op 

* https://github.com/commaai/panda/pull/2225
* https://github.com/commaai/openpilot/pull/35629

PR #2433, opened 2025-07-04
mazda: fix unreliable cruiseState.standstill

jakethesnake420:
I have witnessed cases where the cp.vl["PEDALS"]["STANDSTILL"] signal is not accurate which makes it so openpilot will not create the resume button command as can be seen in the plot. 

image

It is possible that in this case, the vehicles cruise control system will not allow a resume signal, in which case this is a mazda ECU bug. 

[c585568edc695aed_00000002--af81e42324--2--qlog.zip](https://github.com/user-attachments/files/21050225/c585568edc695aed_00000002--af81e42324--2--qlog.zip)


adeebshihadeh:
How certain are you that there is no correct signal for this on CAN?

jakethesnake420:
Can't be sure. _Usually_ the pedal signal is pretty good. I can tell you that cp.vl["PEDALS"]["STANDSTILL"] is active when the car is moving in reverse, which tells me that the signal is not indicating a true standstill state, and not indicating the willingness of the mazda cruise control system to accept a resume command. Given that the purpose of cruiseState.standstill is simply to prevent resume commands from being spammed (since long control isn't supported), this can be safely merged imo.

PR #2445, opened 2025-07-08
VW MQB: Implement vEgoCluster

jyoung8607:
As many other vehicles do, certain VW MQB have a small positive offset for the speed displayed in the instrument cluster. It isn't always present; it doesn't happen on any of the cars I own. I think it may only be cars outside the USA.

Solving for the stock ACC case is easy: the instrument cluster reports the speed it displayed. We'll just always use that. This PR implements the change.

Solving for openpilot longitudinal is more nuanced. We need to know when/how to offset `vCruiseCluster`. We might not need to offset it at all, the cluster may do it for us. If we need to, the offset looks to be `vEgo * ~1.045` with some rounding effects. If the USA-only theory is correct, the cluster also exports a handy bit-flag for USA clusters. All of this will need to be checked against commaCarSegments, and will have to wait until I can spend time on VW alpha long.

**Route:** `9d09cc205c254c4b/00000143--601c3ee13a`

image


dparring:
Following this.  It may be needed on MLB/Macan also

jyoung8607:
Tested successfully.

**Route:** `9d09cc205c254c4b/00000000--0d24caa859`

Also verified we do need to offset `vCruiseCluster` under OP long, but that will be a separate PR.

PR #2448, opened 2025-07-09
Portably find DBC files

sshane:
Related: https://github.com/commaai/opendbc/pull/2423/files


sshane:
nvm, this is relative to the cwd :(

PR #2450, opened 2025-07-09
Fix external project build

sshane:
fixes openpilot build. from: https://github.com/commaai/opendbc/pull/2423


opendbc/can/SConscript
```diff
@@ -13,6 +13,7 @@ libdbc = envDBC.SharedLibrary('libdbc', src, LIBS=[common, ], LINKFLAGS=LINKFLAG
 
 # Build packer and parser
 lenv = envCython.Clone()
+lenv["LIBPATH"].append(Dir("."))
```
sshane, line 16: causes harmless duplicate -L in opendbc, fixes external project building

PR #2451, opened 2025-07-10
Rivian: Accel should be zero when longitudinal control is inactive.

lukasloetkolben:
Distinguishes between enabled and active longitudinal control. Ensures acceleration is zero when inactive to prevent safety code from blocking frames, for example, during gas pedal override.

### DIFF
--- a/opendbc/car/rivian/carcontroller.py
+++ b/opendbc/car/rivian/carcontroller.py
@@ -36,7 +36,7 @@ class CarController(CarControllerBase):
     # Longitudinal control
     if self.CP.openpilotLongitudinalControl:
       accel = float(np.clip(actuators.accel, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX))
-      can_sends.append(create_longitudinal(self.packer, self.frame, accel, CC.enabled))
+      can_sends.append(create_longitudinal(self.packer, self.frame, accel, CC.enabled, CC.longActive))
     else:
       interface_status = None
       if CC.cruiseControl.cancel:
--- a/opendbc/car/rivian/riviancan.py
+++ b/opendbc/car/rivian/riviancan.py
@@ -62,10 +62,10 @@ def create_wheel_touch(packer, sccm_wheel_touch, enabled):
   return packer.make_can_msg("SCCM_WheelTouch", 2, values)
 
 
-def create_longitudinal(packer, frame, accel, enabled):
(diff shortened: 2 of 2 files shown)


opendbc/car/rivian/carcontroller.py
```diff
@@ -36,7 +36,7 @@ def update(self, CC, CS, now_nanos):
     # Longitudinal control
     if self.CP.openpilotLongitudinalControl:
       accel = float(np.clip(actuators.accel, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX))
-      can_sends.append(create_longitudinal(self.packer, self.frame, accel, CC.enabled))
+      can_sends.append(create_longitudinal(self.packer, self.frame, accel, CC.enabled, CC.longActive))
```
sshane:
```suggestion
      can_sends.append(create_longitudinal(self.packer, self.frame, accel, CC.enabled))
```

opendbc/car/rivian/riviancan.py
```diff
@@ -62,10 +62,10 @@ def create_wheel_touch(packer, sccm_wheel_touch, enabled):
   return packer.make_can_msg("SCCM_WheelTouch", 2, values)
 
 
-def create_longitudinal(packer, frame, accel, enabled):
+def create_longitudinal(packer, frame, accel, enabled, active):
```
sshane:
```suggestion
def create_longitudinal(packer, frame, accel, enabled):
```

opendbc/car/rivian/riviancan.py
```diff
@@ -62,10 +62,10 @@ def create_wheel_touch(packer, sccm_wheel_touch, enabled):
...
   values = {
     "ACM_longitudinalRequest_Counter": frame % 15,
-    "ACM_AccelerationRequest": accel if enabled else 0,
+    "ACM_AccelerationRequest": accel if active else 0,
```
sshane:
```suggestion
    "ACM_AccelerationRequest": accel,
```

sshane:
accel is designed to go inactive when long isn't active. Multiple ports assume this behavior already, so we can remove this check here

lukasloetkolben:
Screenshot_from_2025-06-30_10-24-23

Accel is not immediately 0 when the gas pedal is pressed. It's running SunnyPilot. I will double check it by enabling long on stock openpilot.

sshane:
Sounds like a SP bug then

PR #2455, opened 2025-07-11
Add average frequency check

adeebshihadeh:
------
https://chatgpt.com/codex/tasks/task_e_68707844cdd08325be03441410845fdb

### DIFF
--- a/opendbc/can/tests/test_packer_parser.py
+++ b/opendbc/can/tests/test_packer_parser.py
@@ -349,6 +349,25 @@ class TestCanParserPacker:
       "CHECKSUM": 0,
     }
 
+  def test_avg_frequency_check(self):
+    dbc_file = "honda_civic_touring_2016_can_generated"
+    freq = 100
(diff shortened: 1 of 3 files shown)


opendbc/can/tests/test_packer_parser.py
```diff
@@ -349,6 +349,25 @@ def test_track_all_signals(self):
...
+    for _ in range(5):
+      parser.update_strings([int(t), [packer.make_can_msg("VSA_STATUS", 0, {})]])
+      t += int(0.01 * 1e9)
+
+    # slower messages should fail the avg check
+    for _ in range(15):
+      parser.update_strings([int(t), [packer.make_can_msg("VSA_STATUS", 0, {})]])
+      t += int(0.03 * 1e9)
+
+    assert not parser.can_valid
```
sshane, line 369: not a good test. what if we aren't sending correctly in the first loop and `can_valid` was never True?

adeebshihadeh:
Definitely not Codex's best work

PR #2463, opened 2025-07-14
Honda: Door open message refactor, part 1

jyoung8607:
Convert the all-doors-state message to auto-detection.

Doors beyond the driver's door aren't necessarily of critical value, but it won't be too hard to keep with auto-detect, and I'd like to minimize behavior changes during the refactor project.

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -52,6 +52,8 @@ def get_can_messages(CP, gearbox_msg):
   if CP.flags & HondaFlags.BOSCH_ALT_BRAKE:
     messages.append(("BRAKE_MODULE", 50))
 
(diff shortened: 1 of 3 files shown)


mvl-boston:
You can clean further with ideas from pr 2168

There are 3 categories:
1) 3 exception nidec cars: CAR.HONDA_ODYSSEY_CHN, CAR.HONDA_FREED, CAR.HONDA_HRV
2) All other Nidec plus Bosch A CRV
3) All other Bosch (including every open Bosch A/B/C pr)

Per abdeeb's request in 2168, I have a Bosch A CRV tester creating a door open log to see if the Bosch standard signal also works, so that this one exception can be cleaned up too.

jyoung8607:
> You can clean further with ideas from pr 2168

This is what I was thinking with the TODO: #2465

PR #2465, opened 2025-07-14
Honda: Door open message refactor, part 2

jyoung8607:
Get a little clever:

* Auto-detect and use any DRIVERS_DOOR_OPEN signal in the SCM messages for this car
* If the DOORS_OPEN message is available, OR those results with any SCM signal we got

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -128,6 +128,7 @@ class CarState(CarStateBase):
     # panda checks if the signal is non-zero
     ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 1e-5
 
+    # doorOpen is true if we can find any door open, but signal locations vary, and we may only see the driver's door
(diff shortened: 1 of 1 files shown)


opendbc/car/honda/carstate.py
```diff
@@ -128,15 +128,13 @@ def update(self, can_parsers) -> structs.CarState:
     # panda checks if the signal is non-zero
     ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 1e-5
 
+    # doorOpen is true if we can find any door open, but signal locations vary, and we may only see the driver's door
+    ret.doorOpen = bool(cp.vl["SCM_FEEDBACK"].get("DRIVERS_DOOR_OPEN", False)) or \
+                   bool(cp.vl["SCM_BUTTONS"].get("DRIVERS_DOOR_OPEN", False))
```
sshane, line 133: This feels like it's possible for no source to be available, and silently pass. We should design it so that's not possible

jyoung8607: That's fair, let me think on that.

jyoung8607: #2468

PR #2472, opened 2025-07-15
Docs: Move nonfunctional code out of [brand]/values.py

ugtthis:
_This is still in the works_, but wanted to put it up to get some thoughts, once I felt there was a concrete path I could see

Goals: 
- move `CarDocs` away from functional code in `values.py`
- make it easier to extend `CarDocs`
- backwards compatible for now. `car_docs = get_brand_metadata(platform) or platform.config.car_docs`
- eventually move/rename `docs.py` and `docs_definition.py` into `/metadata`
- slowly lean into `metadata` naming pattern

Assumptions:
- `CarDocs` will be getting more attributes (road noise attribute, car port status/PR Links, drivetrain, more video links, etc)
-  manual information input from community. "some info will be populated by openpilot, and the rest will be done manually" (comment link [here](https://github.com/commaai/opendbc/issues/1144#issue-2476581055))
- `metadata` naming is more accurate and forward looking than `CarDocs`. opendbc is for robotics not specifically just cars i.e. Billy
- I used a script to generate the current specs of `CarDocs` to compare against the migrated `CarDocs` to make sure nothing accidentally got changed
- Removing `init_make` gets `CarDocs` less dependent on `values.py` and more clear to contributers what current data individual cars have

Related:
#1144 
 Metadata Framework Discord [current-projects](https://discord.com/channels/469524606043160576/1318660916329451540)

Related PRs:
Ready: #2406 - Make Ford models more obvious
In Progress: - Remove `init_make` from `CarDocs` and use static solutions

### DIFF
--- a/opendbc/car/hyundai/values.py
+++ b/opendbc/car/hyundai/values.py
@@ -1,11 +1,10 @@
 import re
 from dataclasses import dataclass, field
-from enum import Enum, IntFlag
+from enum import IntFlag
 
 from opendbc.car import Bus, CarSpecs, DbcDict, PlatformConfig, Platforms, uds
 from opendbc.car.common.conversions import Conversions as CV
 from opendbc.car.structs import CarParams
-from opendbc.car.docs_definitions import CarFootnote, CarHarness, CarDocs, CarParts, Column
 from opendbc.car.fw_query_definitions import FwQueryConfig, Request, p16
 
 Ecu = CarParams.Ecu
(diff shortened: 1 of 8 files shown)


opendbc/car/hyundai/values.py
sshane, line 1: There's a balance here of putting everything easily in one place that's static, since it's easy to see the entire state of a specific car, or entire brand. And splitting what is static (values.py) and dynamic (interface.py).

This split of static car docs and static car specs doesn't feel necessary to me at this point in time. That's not to say it won't be a useful refactor in the future, but this isn't blocking any impending clean up as far as I'm aware.

ugtthis: The main north star was to create something for "doc-only info" that Adeeb mentioned in a comment [here](https://github.com/commaai/opendbc/pull/1478#issuecomment-2506764854). Splitting static car docs seemed to be the simplest solution that provides a place for doc-only info

By not splitting static car docs we end up w/more noise in `values.py`(i.e. video links) and it slows down the addition of model specific doc info(which could possibly not be the current priority)

The split helps with adding more doc-only info i've been wanting to add, to then show on opendbc.com, i.e. PR links to show what car ports are depending on([jyoung's comment](https://github.com/commaai/opendbc/pull/1365#issue-2583201438)), better explanations why a car port is in its current status, videos to show how to open car's camera casing(my comment on [discord](https://discord.com/channels/469524606043160576/1318660916329451540/1394733979722317886)) to name a few. Doing PR's for these didn't seem like it was pushing forward in a meaningful way towards the core problem, the "doc-only info" framework. 
<br>
#### Why prioritize this?
Solving the "doc-only info" framework seems to be the core bottle neck to giving a better web experience for both devs and consumers allowing opendbc.com to have all sorts of doc only info that could be leveraged to solve and improve existing issues(install experience, car porting experience, car buying experience(Adeeb's [X post](https://x.com/adeebshihadeh/status/1914851620124529029)), contributor experience, etc)

sshane:
Thanks for the work, but this PR is now out of date and adds another hidden file you need to know to add when doing a brand port, or change when doing a model port or change. We want to reduce surface area right now, then we can add functionality

PR #2473, opened 2025-07-15
Honda - match stock hud lane lines

mvl-boston:
Update Honda lane lines in HUD to match stock ACC:
- When ACC enabled but LKAS unavailable, show dashed lines.  Per tester request, show this when steer is saturated.
- When ACC enabled and steering is controlled, show solid lines.

Validation
* Dongle ID: d7233a428eb7d0b5
* Route: 00000002--afc9b6dd9b (see 9:06 timestamp)

Stock Honda controls are:
lkas disabled:
LKAS_HUD / DASHED_LANES = 0
LKAS_HUD / SOLID_LANES = 0

lkas enabled
LKAS_HUD / DASHED_LANES = 1
LKAS_HUD / SOLID_LANES = 0

lkas active (helps steer)
LKAS_HUD / DASHED_LANES = 1
LKAS_HUD / SOLID_LANES = 1

### DIFF
--- a/opendbc/car/honda/carcontroller.py
+++ b/opendbc/car/honda/carcontroller.py
@@ -228,8 +228,9 @@ class CarController(CarControllerBase):
     # Send dashboard UI commands.
     # On Nidec, this controls longitudinal positive acceleration
     if self.frame % 10 == 0:
+      display_lines = hud_control.lanesVisible and (abs(apply_torque) < self.params.STEER_MAX)
       hud = HUDData(int(pcm_accel), int(round(hud_v_cruise)), hud_control.leadVisible,
-                    hud_control.lanesVisible, fcw_display, acc_alert, steer_required, hud_control.leadDistanceBars)
+                    display_lines, fcw_display, acc_alert, steer_required, hud_control.leadDistanceBars)
       can_sends.extend(hondacan.create_ui_commands(self.packer, self.CAN, self.CP, CC.enabled, pcm_speed, hud, CS.is_metric, CS.acc_hud, CS.lkas_hud))
 
       if self.CP.openpilotLongitudinalControl and self.CP.carFingerprint not in HONDA_BOSCH:
--- a/opendbc/car/honda/hondacan.py
+++ b/opendbc/car/honda/hondacan.py
@@ -171,6 +171,7 @@ def create_ui_commands(packer, CAN, CP, enabled, pcm_speed, hud, is_metric, acc_
     'SET_ME_X41': 0x41,
     'STEERING_REQUIRED': hud.steer_required,
     'SOLID_LANES': hud.lanes_visible,
+    'DASHED_LANES': int(enabled),
     'BEEP': 0,
   }
 


opendbc/car/honda/carcontroller.py
```diff
@@ -228,8 +228,9 @@ def update(self, CC, CS, now_nanos):
     # Send dashboard UI commands.
     # On Nidec, this controls longitudinal positive acceleration
     if self.frame % 10 == 0:
+      display_lines = hud_control.lanesVisible and (abs(apply_torque) < self.params.STEER_MAX)
```
jyoung8607: Within `create_ui_commands` this turns into `hud.lanes_visible`, which used to equal and is now extremely easy to confuse with `hud_control.lanesVisible`. I know it confused me when reading it.

If we're going to do this (which is TBD) I'd rather this logic live with the rest, and pass in a `steer_saturated` bool.

opendbc/car/honda/hondacan.py
```diff
@@ -171,6 +171,7 @@ def create_ui_commands(packer, CAN, CP, enabled, pcm_speed, hud, is_metric, acc_
     'SET_ME_X41': 0x41,
     'STEERING_REQUIRED': hud.steer_required,
     'SOLID_LANES': hud.lanes_visible,
```
jyoung8607: Your change to the Bosch SOLID_LANES logic will also apply to the DASHED_LANES logic for Bosch radarless and CAN-FD. That may not have been intentional. This code is pretty tangled and it's on my refactor list.

opendbc/car/honda/hondacan.py
```diff
@@ -171,6 +171,7 @@ def create_ui_commands(packer, CAN, CP, enabled, pcm_speed, hud, is_metric, acc_
...
+    'DASHED_LANES': int(enabled),
```
jyoung8607, line 169: The description in the PR talks about dashed lines being used for LKA available but disengaged, and solid lines being used for active steering. That makes sense. Given those conditions though, do we have the signals backward?

Pre-existing bugs are definitely a possibility here, just trying to understand.

opendbc/car/honda/hondacan.py
mvl-boston: The plan was to apply the solid lane logic to all Honda cars, all have limited steering so early limit warning is beneficial to all.  Steer saturation comma beeps are 1-2 seconds later as the car has to veer off course for comma to detect.

opendbc/car/honda/hondacan.py
mvl-boston: Per the stock behavior above, the "dashed lanes" signal is active during solid or dashed lanes.   It only becomes zero when there is no lane.   Thus I set it to "enabled" which is whenever cruise is active and lane control is possible.

The "Solid lanes" signal is what fills in the dashed lanes to make them solid.

So I don't think this is backwards.   I clarified the stock conditions above, maybe that explains?

opendbc/car/honda/carcontroller.py
mvl-boston: Yes, the variables were named differently in existing carcontroller vs hondacan.   But I see your point about passing in the bool to not confuse code readers, that would have been my first choice but I understood comma had a preference for "change as little as possible".  Will revise.

Thinking longer term for odyssey and rdx which will also use this when car ecus (radar/eps) block steering (lowspeed or wheeljerks), maybe a more generic term like steer_restricted (instead of saturated) is more appropriate and I can set this bool in carstate?

mvl-boston: code has been modified to use a steer_restricted bool

opendbc/car/honda/carcontroller.py
```diff
@@ -150,6 +151,7 @@ def update(self, CC, CS, now_nanos):
     # steer torque is converted back to CAN reference (positive when steering right)
     apply_torque = int(np.interp(-limited_torque * self.params.STEER_MAX,
                                  self.params.STEER_LOOKUP_BP, self.params.STEER_LOOKUP_V))
+    self.steer_restricted = (abs(apply_torque) == self.params.STEER_MAX) # for hud lane line display
```
sshane: Let's not arbitrarily define what already exists. Also, is there no hands on alert we can show instead?
 https://github.com/commaai/opendbc/blob/7e5239c747a0b47e2ba3e6ad1aa10acff14bfe38/opendbc/car/car.capnp#L419

jyoung8607: > Also, is there no hands on alert we can show instead?

I was actually going to ask, what happens if we set `STEERING_REQUIRED` with `BEEP` 0? Is that suitable?

jyoung8607:
Just to set expectations, this PR is alive but won't be merged for at least a couple weeks:

* No matter what direction we go, it'll cause a visible behavior change for a lot of Honda drivers, and I want to avoid that for this release cycle.
* We're not nearly done bikeshedding, and I want to see how this works on a rental.

opendbc/car/honda/carcontroller.py
mvl-boston: steering required with beep zero might be acceptable, but it should be limited to a not steeringpressed state.

From a driving experience standpoint, steering_required creates a hud handsoff alert which is a blinking orange (fault color) alert on the dash similar to a CMBS active-event alert.   That signifies an emergency condition which should only be relevant when a driver is doing something demanding an immediate reaction.  When the driver resumes steering manually, the blinking alert should turn off.

However the dashed lane lines state should remain while steeringpressed, to signal that manual steering remains required, consistent with how stock behaves.

If you're on board with this approach, I can rewrite the signals in the pr.

mvl-boston:
> Just to set expectations, this PR is alive but won't be merged for at least a couple weeks:

I do believe a warning solution of some kind is required before porting the ALT_RADAR models (21+ Odyssey and 22+ RDX).   They disable steer unexpectedly upon high g-force (eg: wheel jerk or driver-steered high speed turn, sometimes needing minute-long cool-down period before reenabling), and also always disable steer below 45mph threshold.

And I found this lack of warning very unsafe as noted in the bug https://github.com/commaai/opendbc/issues/1637, as a driver shouldn't be surprised by a silent loss of steer, and especially in a cool down disable timer state shouldn't be trial-and-error "guessing" when comma can start steering again.

jyoung8607:
All of that stuff is why the steerFault flags and low speed alert exist. Nothing in this PR is substitute for correctly detecting loss of control.

mvl-boston:
Ok.   I'll look for the most matching status flags.   In this PR I'll
update UI to match driver expectations for all steer alerts, based on
alignin with what stock instrument cluster does.

In the carports I'll point Odyssey and RDX steer blocks to the most
matching existing comma flag.

Plan to do this tonight (Eastern US timezone).

mvl-boston:
Completed the redo of this logic to be based on existing openpilot visualhud state, with a focus on better matching the two (regular vs emergency) steer_unavailable instrument cluster messages from stock Honda.

Revised the initial PR description above to provide a detailed explanation of new thinking/approach.

mvl-boston:
Moved to draft - I found separate existing carstate status values for emergency and regular steer alerts, and per shane's request will attempt to use the existing ones rather than recreate.

Will move to ready after testers try new carstate values.

mvl-boston:
> Moved to draft - I found separate existing carstate status values for emergency and regular steer alerts, and per shane's request will attempt to use the existing ones rather than recreate.
> 
> Will move to ready after testers try new carstate values.

Unfortunately it was an event status which is not exposed to opendbc.   So I had to copy/paste the logic into carcontroller to match.

Will ask testers to roadtest how it works.

mvl-boston:
Moving back to draft - looks like steer status alone isn't sufficient as it shows status "normal" with "steer control" = 0.   Need to revise logic.

jyoung8607:
Having driven this for awhile, I see what you want to accomplish and I'm in general agreement. There's some bugs to be worked out first. I'll merge this (or code in this spirit) in different pieces as I validate them.

mvl-boston:
Two interface issues raised by Odyssey test drivers.  For consideration as you clean this up, both seem to be in core OP warnings rather than Odyssey specific:

1) For undetermined reasons, steer_status sometimes moves to low_speed_lockout in the final 3mph braking to a stop, rather than staying normal while stopping.   The silent_steer_warning simulation I wrote works well for the instrument cluster (just showing dashed lines), but the "TAKE CONTROL IMMEDIATELY" error while the car is stopped is too harsh for a silent steer warning, especially if the driver is stopped behind a lead car.   silent_steer_warning probably needs a downgrade to no 3x screen message at all while the car is in standstill, or at least a less obtrusive alert until the car moves again.   Can't turn off the the steerFaultTemporary status, because I don't want the car to attempt to steer which would trigger a longer-term lockout.

2) The lowSpeedAlert message and the above steerFaultTemporary message felt misworded to a tester.  The phrase "Steering Unavailable" can easily be misread as the driver is blocked from being able to manually steer.   They would recommend something like "LKAS unavailable" or "AutoSteer unavailable".   My personal preference would be "Assisted Steering Unavailable".

Happy to open standalone pr's for either if needed since technically independent from Odyssey.  Please let me know.

jyoung8607:
Picking up these concepts in #2773.

PR #2474, opened 2025-07-15
draft - Honda - Rdx 2025 support

mvl-boston:
(not needed - see closure note)

Additional changes seem to be needed to stabilize 2025 RDX, especially in long control.

Draft until I can get a tester for further evaluation.

Honda recently published a 25 RDX EPS recall, will see if this is no longer needed after recall software is applied.


mvl-boston:
Remaining unresolved use case is that the car continues to shutdown EPS with steerstatus=7 when the driver jerks the wheel during an ACC braking event, even if comma steering is disabled.

Ran a dashcam test and found a different/unmapped steer control bit set during stock ACC braking.   Initial attempts to set this bit resulted in inability to regain comma steer control even when not braking.
Todo 1: find if this bit can be used to prevent EPS faults, potentially setting it only when braking and clearing it during gas.   

Separately, found an unmapped steer status bit that might be a hands on sensor.
Todo 2: test if disabling steer control during hands-on is a reliable way to prevent shutdown of EPS

Awaiting another 2025 loaner to proceed with todo1 and todo2.

mvl-boston:
Closing, the problems were confirmed fixed by a driver after implementing the NHTSA power steering safety recall.

PR #2475, opened 2025-07-15
Honda: CRV 2023-2025 (canfd)

mvl-boston:
Support for CANFD CRV 2023-2025
- merge and update of PR 2117 + 2119 from vanillagorillaa
- revised alt_brake flag to autodetection (used by ICE CRV but not Hybrid CRV)
- renamed pilot 2023 dbc to honda_canfd_common since all canfd cars so far are identical
- include safety changes from from shane and vanillagorillaa in pr 2129 

Extensively roadtested as part of combined-honda builds, inclusive of separate long control and tuning code.

Validation (ICE):
* Dongle ID: 63568e3e2f56c8ad
* Route: 0000000a--a254e90429

Validation (hybrid):
* Dongle ID: 864613393836f643
* Route: 00000002--3cef07213f

dashcam mode:
• 0a84f6dd8372aef6/00000001--9ade6fe7d3

### DIFF
--- a/opendbc/car/honda/values.py
+++ b/opendbc/car/honda/values.py
@@ -211,8 +211,8 @@ class CAR(Platforms):
   )
   HONDA_CRV_6G = HondaBoschCANFDPlatformConfig(
     [
-      HondaCarDocs("Honda CR-V 2024-25", "All"),
-      HondaCarDocs("Honda CR-V Hybrid 2024-25", "All"),
+      HondaCarDocs("Honda CR-V 2023-25", "All"),
+      HondaCarDocs("Honda CR-V Hybrid 2023-25", "All"),
     ],
     CarSpecs(mass=1703, wheelbase=2.7, steerRatio=16.2, centerToFrontRatio=0.42),
   )


mvl-boston:
One of the earlier signal refactors broke canfd.   Working with testers to figure out which one.   Moving to draft until a workaround can be found.

mvl-boston:
Reopening.   Drivers said the bugs only impacted long-control, which isn't part of this port yet.

jyoung8607:
- [ ] Get a dashcam route with strong factory LKA actuation to find known good actuation range
- [ ] Convert to lateral accel torque controller
- [ ] Refine params with test drives
- [ ] Update CI test routes

mvl-boston:
Had to use an older log from another drive to set values, the torque controller was blank on this log.

Also, need to find someone who can test ICE.   Or should I separate into its own PR?

mvl-boston:
dashcam mode 0a84f6dd8372aef6/00000001--9ade6fe7d3

jyoung8607:
Going to collapse the ICE and hybrid variants:
* The test route didn't have any engagement (fixable, but no need I think)
* Mass difference isn't enough to worry about
* We collapse ICE and hybrid for every other model except the older CR-V, which could be cleaned up
* We're auto-detecting the need for ALT_BRAKE, which is probably why the older CR-V is split

jyoung8607:
LGTM pending some additional drive-time to refine the lat accel controller factors. Notes for merge:

* Support verified for MY2023 through MY2025 inclusive
* The fingerprint list was recreated from seven known dongles, notes in commit history/git blame
* Startup values are near live-learned values for liveParameters and liveDelay
  * Want some more drive-time before trusting liveTorqueParameters
  * #2632
  * #1100

PR #2479, opened 2025-07-15
update steerRatio to match the actual Volkswagen Steer Ratio

tealtwo:
update steerRatio to match the actual Volkswagen Steer Ratio which is `16.4`, instead of the 15.6 that was used previously (which is incorrect)


dkiiv:
https://media.vw.com/assets/documents/original/6238-1953807316536a81198b65a.pdf
https://media.vw.com/assets/documents/original/6426-198352647857d84b5610a8c.pdf

dkiiv:
this would need to be an edit made specifically for PQ. not VW as a whole

jyoung8607:
What's there isn't correct, but this PR isn't an improvement. Any fine-tuning would need to be per-model, not global, and the value shouldn't be set to the mechanical ratio in the EPS, but the params learner output after some driving (which includes additional suspension effects).

PR #2485, opened 2025-07-16
Honda: Steer control status refactor, part 1

jyoung8607:
The change in message length for Nidec `STEER_STATUS` happens to correlate nicely with its companion control messages, which were already commonized.

This PR common-izes the length 7 variant. The length 6 variants disagree on the meaning of one signal. There's an excellent chance that's a bug in the DBC rather than a real difference, but will have to examine separately.


jyoung8607:
Ugh, the VAL enums are slightly different, and there's a specific CI speed-bump for this change so it could be legit.

PR #2488, opened 2025-07-18
Honda: Honda Odyssey 2021-25

mvl-boston:
Add support for 2021 to 2025 Honda Odyssey

- car fingerprint named 5G MMR. That is honda-speak for 5th generation mid-model refresh
- Added BOSCH_ALT_RADAR flag as most changes are shared with 3G_MMR RDX, and possibly other Bosch A unported cars.
- US Odysseys use imperial speed/cruise units on the canbus, Canadian Odysseys use metric.
- Odyssey radar removes steering when under 45mph with radar on (eg: when longitudinal is off). Used the canbus flag that identifies the steering lockout, to display lockout on the instrument cluster consistent with OEM display.
- Due to the intrusiveness of comma steering lockout on Hondas, the driving experience will be distracting without https://github.com/commaai/opendbc/pull/2473.

Hardware: Comma 3x
Harness: Comma Honda Bosch A
Test routes:

Hardware: Comma 3x
Harness: Comma Honda Bosch A
Test routes:
- d7233a428eb7d0b5/00000001--9b99b04d43
- 558ac87ada6a88cd/0000000f--a0a05b334e
- 558ac87ada6a88cd/00000036--7d59e2af08
- 558ac87ada6a88cd/0000005b--0d40d84bb6

Dashcam route to prove standard Honda max steer of 2560:
- 805f87f5e96d128c/00000002--c257be37f0

Todo; update test route with latest logic

### DIFF
--- a/opendbc/car/honda/fingerprints.py
+++ b/opendbc/car/honda/fingerprints.py
@@ -562,6 +562,7 @@ FW_VERSIONS = {
   CAR.HONDA_ODYSSEY_5G_MMR: {
     (Ecu.vsa, 0x18da28f1, None): [
       b'57114-THR-A240\x00\x00',
+      b'57114-THR-A520\x00\x00',
     ],
     (Ecu.fwdRadar, 0x18dab0f1, None): [
       b'36802-THR-A220\x00\x00',
@@ -576,21 +577,26 @@ FW_VERSIONS = {
     (Ecu.transmission, 0x18da1ef1, None): [
       b'28102-5MX-A100\x00\x00',
       b'28102-5MX-A200\x00\x00',
+      b'28102-5MX-A410\x00\x00',
     ],
     (Ecu.srs, 0x18da53f1, None): [
       b'77959-THR-A220\x00\x00',
       b'77959-THR-A230\x00\x00',
+      b'77959-THR-A320\x00\x00',
     ],
     (Ecu.electricBrakeBooster, 0x18da2bf1, None): [
       b'46114-THR-A530\x00\x00',
       b'46114-THR-A540\x00\x00',
+      b'46114-THR-A720\x00\x00',
     ],
     (Ecu.gateway, 0x18daeff1, None): [
       b'38897-THR-A130\x00\x00',
+      b'38897-THR-A320\x00\x00',
       b'38897-THR-A410\x00\x00',
     ],
     (Ecu.eps, 0x18da30f1, None): [
       b'39990-THR-A050\x00\x00',
+      b'39990-THR-A110\x00\x00',
     ],
   },
   CAR.HONDA_PILOT: {


jyoung8607:
Drove around earlier to collect a dashcam route for actuator limits, and had some interesting outcomes.

**Route:** `f39cf149898833ff/0000004b--05d3acf28f`

1) There are *two* clear actuation maxima from the stock LKA system, one at 2560 (which we've been using as a common limit for CAN-FD and other cars) and one at 3840 (which was formerly in openpilot for some cars). This may be a difference in operating envelope for Lane Keeping Assist vs. Road Departure Mitigation; the higher peak was preceded by both a HUD warning and steering wheel haptic feedback. I'll try to establish if 3840 is usable during normal driving.

image

image

2) During a driver input correction on a curve, the EPS went to state 2, followed by an immediate torque cut and another control bit reaction from the camera. Control resumed immediately, along the prior torque curve (not rising from zero) after the EPS returned to state 0. This all happened in a fraction of a second. EPS state is almost certainly the trigger.

image

mvl-boston:
Check torque on higher steering control, the jump from 2560 to 3840 appears to be lower slope vs prior numbers, and possibly even negative slope.   I'll prioritize reviewing some donuts (scripted increasing steer control) logs for review.

I also noted that the UI is not behaving as intended with pr2473.   Looks like steer_status is staying at normal but steer_control_active is dropping to zero when going below 45mph minspeed.   Need to write steer_control_active as a minspeed fault to properly set the LKAS lines to dashed.
805f87f5e96d128c/00000013--24784bc32c/7

There are half a dozen Odyssey and RDX testers on alert for any tests/driver requests here, they all know you have the car this week.

jyoung8607:
- [x] ~~Stock sends all three LKAS_HUD message variants on this car, but we are only sending two?~~
- [ ] #2473 and children thereof
- [x] Pick up RDM changes from #252, they appear correct
- [x] ~~Investigate RDM control bit, see if it unlocks the 2560-to-3840 range~~
- [ ] ~~Re-check steer-to-zero bit~~
- [ ] ~~Re-check send direct to powertrain~~

The RDM thing is probably just a haptic warning, but the radar is clipping the relevant bit on the way through. I tried increasing torque to 3840 and the radar passes that, but it does seem to fall flat. Will try to experiment more when using OP long and the radar isn't an intermediary.

mvl-boston:
Wanna drop alt_radar value to 512 now that its free?

jyoung8607:
- [ ] ~~Fix RDM startup disable/error~~ it's there deliberately, and should probably stay
- [x] Try to get control of the orange LKAS HUD icon 
- [x] Go through the applicable long tuning PR backlog
  * Slightly undershooting target speeds, could be a displayed speed offset thing
  * Feels like we're not using coast effectively, feels like there's a deadband between 0.0 accel and friction braking
- [x] Lat accel torque controller isn't handling this car super well, try legacy PIF
- [x] ~~Not showing AEB disabled under OP long, not sure we're passing thru stock with radar disable, doubt it~~ we are, it's just subtle

jyoung8607:
- [ ] No longitudinal closed loop control at all, no PI correction
- [ ] This thing really needs pitch compensated accel

image

mvl-boston:
Long control for all Bosch A and Bosch C:
- Brake has factory PID, no adjustments needed
- Gas needs pitch correction, drag correction, and different-per-car pedal magnitude adjustment
- Improper existing carcontroller logic that does not apply brake until -0.2

Possible experiment: Radarless has factory PID on gas and brake.   Will Bosch A/C accept radarless control, to allow more accurate gas?

mvl-boston:
>       * [ ]  Slightly undershooting target speeds, could be a displayed speed offset thing

Need drag adjuster to fix, already found/fixed in testing.   Problem is that drag varies per car/elevation so needs a PID.

The OP long problems you are experiencing impact all Honda Bosch A & C.   That is why no one bothers to drive OP Long without long tuning.

jyoung8607:
I don't disagree on any point in principle, just feeling the underlying issues before I try your solutions.

mvl-boston:
> I don't disagree on any point in principle, just feeling the underlying issues before I try your solutions.

Remaining gaps that testers found in my solutions:
- wind drag needs a PID - I'm thinking I can just remove the pid on gas, and rely on the hardcoded pedal factors instead.  Then restrict PID to wind drag only
- zero cutoff is bad - I'm thinking of setting to 0 across the board and get rid of -0.2 start in scaling.   Needs testing to see if there is a 0 creep force in gas.

jyoung8607:
So we're trying to map accel to gas, which works in a very crude way, but I think the gas actuator is actually torque.

~~FWIW I don't buy the *scaling* on these alleged newton-meters, but that's the right unit, we're just not scaling it right.~~ The scaling may actually be correct as-is.

image

In this graph, GAS_PEDAL_2 contains engine requested torque and estimated actual torque output. Requested torque will be the sum of idle and accessory demand, driver accelerator pedal demand, and ACC control demand. Here we're engaged and there's no driver accel pedal, just the sum of ACC demand and overhead demand. Lines right up.

That's why we're having to mess with scaling between cars, because our actuator takes torque through the drivetrain, not simple acceleration. But now that we know that, there's well understood vehicle dynamics equations to get the necessary torque based on `CP.mass`, and deal with rolling resistance and wind resistance and the like.

The spikes you see are upshifts and downshifts. We can also see here how much drag torque the drivetrain is applying under compression braking, and from there know how much negative acceleration that contributes apart from friction braking.

The graph shows the drivetrain is actually doing a good job closing the loop on openpilot's torque demand. The problem is that openpilot is sending dumb commands. We can fix this for sure. We'll still want to do pitch compensation on the acceleration before converting to torque.

mvl-boston:
> That's why we're having to mess with scaling between cars, because our actuator takes torque through the drivetrain, not simple acceleration. But now that we know that, there's well understood vehicle dynamics equations to get the necessary torque based on `CP.mass`, and deal with rolling resistance and wind resistance and the like.

Makes sense - questions are a matter of implementation:
- are there enough inputs to do the calculations?   For example Honda doesn't publish drag coefficients, can opendbc learn this stuff similar to lat_accel_factors?   Or does it require more support/infrastructure from openpilot to create and store values that can persist across routes?   Seems the opendbc PID learned factors don't persist across drives, can a crude one sit in carcontroller and have access to persistent/learned factors?
- does 3x have the horsepower to do full calcs realtime?   Or is it better to simplify into an interp curve like I did with wind drag in my pr?
- Most impactful in my drive experience is the raw gas pedal factors.   I've created them observationally from long maneuver logs and will chart them vs mass specs to see how close they become see if a generic formula can be used.

mvl-boston:
Below are the mass factors based on the longitudinal logs - seems like there is a little more going on than purely linear calc, my guess is deliberate Honda "gas pedal feel" tuning.   But we can get much better with a basic mass factor than we are today.


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/manue/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/manue/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


Car | mass | BOSCH_GAS_LOOKUP_V   - max | factor
-- | -- | -- | --
CIVIC_BOSCH | 1326 | 750 | 57%
ACCORD_11G | 1577 | 1060 | 67%
CRV_6G | 1703 | 1460 | 86%
RDX_3G_MMR | 1850 | 2000 | 108%
ODYSSEY_5G_MMR | 2082 | 2000 | 96%
PILOT_4G | 2114 | 2200 | 104%



</body>

</html>

jyoung8607:
The calculations are trivial for 3X. I had Copilot vibe something for me to play with last night.

The minimums, no reasonable defaults:
* mass, get from CP.mass
* wheel radius, Google for a tire calculator, enter the stock tire size, 1/2 the diameter
* current velocity, get from CS.out.vEgo
* target acceleration, the accel passed down into carcontroller

There's more nice-to-have data, but we have defaults we can get started with, and tinker as we need to. CD doesn't vary that much. Frontal area you can spitball from manufacturer spec width and height. Altitude changes are probably small enough to plaster over with PI. From toying with things last night, the torque difference was important but not insanely huge measuring with 10mph vs 70mph headwind, and these aren't airplanes.

Basically we want to fill in `compute_gb_honda_bosch` by first pitch-compensating the accel and then running the calcs below. Then you can pass its `gas` output to ACC_CONTROL (ignoring the brake) as the torque component. Keep it simple to start and see what the error/mismatch looks like, see if we can get it coarsely correct, and THEN layer on PI.

You're welcome to take a shot at it. Time permitting I'll also come back to hack on this, but my current priority is to get the ALT_RADAR cars merged, and I'll iterate on alpha long after.

```python
import math

def calculate_wheel_torque(target_acceleration, vehicle_mass, velocity, wheel_radius, 
                          rolling_resistance_coeff=0.015, air_density=1.225, 
                          drag_coefficient=0.3, frontal_area=2.5, drivetrain_efficiency=0.9):
    """
    Calculate the wheel torque needed to achieve a target vehicle acceleration.
    
    Args:
        target_acceleration (float): Desired acceleration in m/s²
        vehicle_mass (float): Vehicle mass in kg
        velocity (float): Current vehicle velocity in m/s
        wheel_radius (float): Effective wheel radius in meters
        rolling_resistance_coeff (float): Rolling resistance coefficient (default: 0.015 for car tires on asphalt)
        air_density (float): Air density in kg/m³ (default: 1.225 at sea level, 15°C)
        drag_coefficient (float): Aerodynamic drag coefficient (default: 0.3 for typical car)
        frontal_area (float): Vehicle frontal area in m² (default: 2.5 m² for typical car)
        drivetrain_efficiency (float): Drivetrain efficiency factor (default: 0.9)
    
    Returns:
        float: Required wheel torque in N⋅m
    """
    
    # Force needed to accelerate the vehicle (F = ma)
    acceleration_force = vehicle_mass * target_acceleration
    
    # Rolling resistance force (F_rr = Crr × N × g)
    # where N is normal force (approximately mg for flat ground)
    gravity = 9.81  # m/s²
    rolling_resistance_force = rolling_resistance_coeff * vehicle_mass * gravity
    
    # Aerodynamic drag force (F_d = 0.5 × ρ × Cd × A × v²)
    drag_force = 0.5 * air_density * drag_coefficient * frontal_area * velocity**2
    
    # Total force needed at the wheels
    total_force = acceleration_force + rolling_resistance_force + drag_force
    
    # Account for drivetrain efficiency
    # Higher force needed at wheels due to losses in drivetrain
    wheel_force = total_force / drivetrain_efficiency
    
    # Convert force to torque (T = F × r)
    wheel_torque = wheel_force * wheel_radius
    
    return wheel_torque

def calculate_wheel_torque_detailed(target_acceleration, vehicle_mass, velocity, wheel_radius,
                                  rolling_resistance_coeff=0.015, air_density=1.225,
                                  drag_coefficient=0.3, frontal_area=2.5, drivetrain_efficiency=0.9,
                                  return_breakdown=False):
    """
    Enhanced version that optionally returns a breakdown of force components.
    
    Args:
        Same as calculate_wheel_torque, plus:
        return_breakdown (bool): If True, returns dict with force breakdown
    
    Returns:
        float or dict: Wheel torque in N⋅m, or dict with detailed breakdown
    """
    
    gravity = 9.81
    
    # Calculate individual force components
    acceleration_force = vehicle_mass * target_acceleration
    rolling_resistance_force = rolling_resistance_coeff * vehicle_mass * gravity
    drag_force = 0.5 * air_density * drag_coefficient * frontal_area * velocity**2
    
    total_resistance = rolling_resistance_force + drag_force
    total_force = acceleration_force + total_resistance
    wheel_force = total_force / drivetrain_efficiency
    wheel_torque = wheel_force * wheel_radius
    
    if return_breakdown:
        return {
            'wheel_torque_nm': wheel_torque,
            'forces': {
                'acceleration_force_n': acceleration_force,
                'rolling_resistance_n': rolling_resistance_force,
                'drag_force_n': drag_force,
                'total_resistance_n': total_resistance,
                'total_force_n': total_force,
                'wheel_force_n': wheel_force
            },
            'parameters': {
                'vehicle_mass_kg': vehicle_mass,
                'velocity_ms': velocity,
                'wheel_radius_m': wheel_radius,
                'target_acceleration_ms2': target_acceleration,
                'rolling_resistance_coeff': rolling_resistance_coeff,
                'drag_coefficient': drag_coefficient,
                'frontal_area_m2': frontal_area,
                'drivetrain_efficiency': drivetrain_efficiency
            }
        }
    
    return wheel_torque

# Example usage and test cases
if __name__ == "__main__":
    # Example: Typical passenger car
    mass = 1500  # kg
    velocity = 20  # m/s (72 km/h)
    wheel_radius = 0.32  # m (typical for 15-16 inch wheels)
    target_accel = 2.0  # m/s²
    
    torque = calculate_wheel_torque(target_accel, mass, velocity, wheel_radius)
    print(f"Required wheel torque: {torque:.1f} N⋅m")
    
    # Detailed breakdown
    result = calculate_wheel_torque_detailed(target_accel, mass, velocity, wheel_radius, 
                                           return_breakdown=True)
    print(f"\nDetailed breakdown:")
    print(f"Total wheel torque: {result['wheel_torque_nm']:.1f} N⋅m")
    print(f"Acceleration force: {result['forces']['acceleration_force_n']:.1f} N")
    print(f"Rolling resistance: {result['forces']['rolling_resistance_n']:.1f} N")
    print(f"Drag force: {result['forces']['drag_force_n']:.1f} N")
    print(f"Total resistance: {result['forces']['total_resistance_n']:.1f} N")
    
    # Test different scenarios
    print(f"\nScenario comparison (for {mass}kg vehicle, {wheel_radius}m wheel radius):")
    scenarios = [
        ("City acceleration (0 km/h)", 0, 2.0),
        ("Highway merge (50 km/h)", 50/3.6, 1.5),
        ("Highway acceleration (100 km/h)", 100/3.6, 1.0),
        ("Gentle acceleration (80 km/h)", 80/3.6, 0.5)
    ]
    
    for description, vel, accel in scenarios:
        torque = calculate_wheel_torque(accel, mass, vel, wheel_radius)
        print(f"{description}: {torque:.1f} N⋅m")
```

jyoung8607:
With stock ACC and the radar filtering, when you first set the enable bit, it delays setting it toward the EPS, and then applies some kind of slow-start damping algorithm to camera's torque request for a few seconds.

This thing is so weird.

image


image

jyoung8607:
Here's some quality engineering. During a RDM event, the stock camera tries to generate a haptic warning with torque commands, but the radar just eats it on the way through. The HAPTIC_WARNING bit made it happen anyway at the EPS, but I wonder if HAPTIC_WARNING might disable damping or rate-limiting on Honda EPS other than ALT_RADAR.

image

jyoung8607:
LGTM, or good enough anyway. Notes:

* There are confirmed examples of MY2021 through MY2025 inclusive
* The fingerprint list was recreated from six known dongles, notes in commit history/git blame
* Learned lateral accel factors are finally converging, but keep an eye on this

Caveats:
* Not much torque, and it's heavily damped, not very good at fine corrections for changing road roll
* In stock ACC mode our torque commands get tampered with a lot, and the stock ACC sucks at hills
* Alpha long drives serviceably, but needs tuning, will iterate on that later

PR #2489, opened 2025-07-18
Tesla coop steer

dzid26:
Adds steer angle offset based on measured driver torque value. 
Uses +/-0.2Nm deadzone to filter out torque sensor offset and other disturbances. 

- [x] target angle from lat accel
- [ ] enable full angle range and higher steering velocity at low speed - but ensure stability
- [ ] ~~learn steering torque offset~~ eh, just increase deadzone to 0.5Nm, since it's somewhat angle dependant
- [ ] 100Hz angle request won't work?

-------

Validation
* Dongle ID: 
* Route:

### DIFF
--- a/opendbc/car/tesla/carcontroller.py
+++ b/opendbc/car/tesla/carcontroller.py
@@ -16,7 +16,9 @@ MAX_LATERAL_ACCEL = ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_R
 MAX_LATERAL_JERK = 3.0 + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)  # ~3.6 m/s^3
 
 STEER_BIAS_MAX = 0.2 # Nm
-STEER_VIRTUAL_SPRING_COEFF = 4.0
+STEER_OVERRIDE_MAX_TORQUE = 2.5 # Nm max torque before EPS disengages
+STEER_OVERRRIDE_MAX_LAT_ACCEL = 2.0 # m/s^2 - similar to Tesla comfort steering mode
+STEER_OVERRRIDE_GAIN_LIMIT = 10 # jerky but stable
 
 
 def get_max_angle_delta(v_ego_raw: float, VM: VehicleModel):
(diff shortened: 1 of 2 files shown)


opendbc/car/tesla/carcontroller.py
```diff
@@ -76,9 +79,16 @@ def update(self, CC, CS, now_nanos):
     # Canceling is done on rising edge and is handled generically with CC.cruiseControl.cancel
     lat_active = CC.latActive and CS.hands_on_level < 3
 
+    steeringAngleDeg = actuators.steeringAngleDeg
+
+    # ignore potential steer bias
+    steeringTorqueDeadzone = CS.out.steeringTorque - np.clip(CS.out.steeringTorque, -STEER_BIAS_MAX, STEER_BIAS_MAX)
+    # create virtual spring effect
+    steeringAngleDeg += steeringTorqueDeadzone * STEER_VIRTUAL_SPRING_COEFF
```
sshane: You can't really directly modify angle from torque. Torque translates to lateral accel. I tried a fork that had this and you couldn't override at all at low speeds. Need to translate to lateral accel, then translate to angle to do this properly

dzid26: Agree. Also using lat accel solves oversensitivity at highway speeds.  

At low speed it will be tricky due to high gain (torque to angle) feedback loop. To prevent oscillations, probably the real driver torque will have to be decoupled from steering inertia torque.  I am thinking maybe a feedforward steering inertia estimator based on the commanded angle could help stabilize it.  

For now I clipped the gain so it doesn't oscillate. - This feature is already is a massive driving experience improvement. 

One thing that should be fixed is torque sensor offset (my steering has pretty consistent -0.2Nm ).  Would be interesting to check values across the fleet.  I don't want to increase a deadzone too much to filter this out else it may become noticable. The offset should be learnt and saved, so I guess it should be added to paramsd ?

dzid26:
Better implementation is here https://github.com/dzid26/opendbc/commit/6331d05cf26a63dccd9dd41a5ed6a894d8eb876d ([dzid26/vtb](https://github.com/dzid26/opendbc/commits/vtb/))

PR #2493, opened 2025-07-20
Genesis GV80: Add 2021 firmware

mattjoubert:
Add firmware for Genesis GV80 2021


jyoung8607:
There's no firmware added in this PR, and no route or dongle ID provided to check. If you can tell us a route, or at least your dongle ID, we may be able to help.

magicbear2025:
Understood, I'll beef it up with a new commit.

magicbear2025:
ok - updated the commit with dongle/route and fingerprint adds.

magicbear2025:
@jyoung8607 hope this does the trick. In sunny pilot it doesn't auto fingerprint and I have to select it. But I have other ecus added .. guessing that's why.

jyoung8607:
How well does it drive? Controls response looks very jittery compared to the MY2023 test route.

magicbear2025:
It has been excellent, seems to do better in traffic than solo. A wobble when accelerating, 7/10... The 2021 I realized today I'm not using lka steering, missed the flag because I'm hda2  on my trim

magicbear2025:
@jyoung8607  so i've been using sunnypilot because i auto-fingerprinting wasnt working - as it turns out, in a moment of trying to get my head around the submodules, I somehow overlooked the actual firmware values. It worked in sunny because of manual selection and q harness. But I didn't have that LKA flag on. so I guess HDA1 was torque only steering.

I see an angle steering as well, not sure if that's available to me (untested), but I think that's related to LFA2? 

In any event, the 2021 is a weird year, just be for the chip shortage and so it has hda2 on the prestige trim, whereas the 22/23 have hda1 as a result of said shortage (if the internet has educated me correctly).

anyway - i've just flipped on the lka flag, it works, wobbles, and my tuning probably made things worse at thiw point, so i'm going to set my branch up to copy the 2023 settings (except having the hda2 stuff on).

magicbear2025:
a5205fb7e46c605b 0000004c--e582e41999 is a public route.

This guy has a little tuning, better model. Less insane settings.

Feels good to me. any change @jyoung8607 ... Or anybody who can help .

Thank you !

mattjoubert:
two weeks of no action - closing pull request.  I have a comma 3x for sale now

PR #2499, opened 2025-07-21
Slimming down `carState`

adeebshihadeh:
The current car port backlog is untenable. While this isn't a magic bullet, I think the first step is to strip out anything that's not strictly necessary for openpilot. These fields may have become luxuries we can no longer afford.

Our API to the car is essentially `carstate.py` and `carcontroller.py`, and the more we reduce that surface area, the easier porting, fingerprinting, and maintenance will be.

@sshane @jyoung8607 Any thoughts? This is a fairly aggressive pass at `carState`. As a side benefit, the smaller `carState` means we can throw more of them in qlogs.

### DIFF
--- a/opendbc/car/car.capnp
+++ b/opendbc/car/car.capnp
@@ -168,7 +168,6 @@ struct CarState {
   vCruiseCluster @54 :Float32; # best estimate of set speed shown on car's instrument cluster, used for UI
 
   standstill @18 :Bool;
-  wheelSpeeds @2 :WheelSpeeds;
 
   gasPressed @4 :Bool;    # this is user pedal only
 
@@ -290,6 +289,7 @@ struct CarState {
   genericToggleDEPRECATED @23 :Bool;
   clutchPressedDEPRECATED @28 :Bool;
   cumLagMsDEPRECATED @50 :Float32;
+  wheelSpeedsDEPRECATED @2 :WheelSpeeds;
 }
 
 # ******* radar state @ 20hz *******
--- a/opendbc/car/gm/carstate.py
+++ b/opendbc/car/gm/carstate.py
@@ -68,16 +68,16 @@ class CarState(CarStateBase):
     # An Equinox has been seen with an unsupported status (3), so only check if either wheel is in reverse (2)
     left_whl_sign = -1 if pt_cp.vl["EBCMWheelSpdRear"]["RLWheelDir"] == 2 else 1
     right_whl_sign = -1 if pt_cp.vl["EBCMWheelSpdRear"]["RRWheelDir"] == 2 else 1
-    ret.wheelSpeeds = self.get_wheel_speeds(
+    wheel_speeds = self.get_wheel_speeds(
       left_whl_sign * pt_cp.vl["EBCMWheelSpdFront"]["FLWheelSpd"],
       right_whl_sign * pt_cp.vl["EBCMWheelSpdFront"]["FRWheelSpd"],
       left_whl_sign * pt_cp.vl["EBCMWheelSpdRear"]["RLWheelSpd"],
       right_whl_sign * pt_cp.vl["EBCMWheelSpdRear"]["RRWheelSpd"],
     )
-    ret.vEgoRaw = float(np.mean([ret.wheelSpeeds.fl, ret.wheelSpeeds.fr, ret.wheelSpeeds.rl, ret.wheelSpeeds.rr]))
+    ret.vEgoRaw = float(np.mean([wheel_speeds.fl, wheel_speeds.fr, wheel_speeds.rl, wheel_speeds.rr]))
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
     # sample rear wheel speeds, standstill=True if ECM allows engagement with brake
-    ret.standstill = abs(ret.wheelSpeeds.rl) <= STANDSTILL_THRESHOLD and abs(ret.wheelSpeeds.rr) <= STANDSTILL_THRESHOLD
+    ret.standstill = abs(wheel_speeds.rl) <= STANDSTILL_THRESHOLD and abs(wheel_speeds.rr) <= STANDSTILL_THRESHOLD
 
     if pt_cp.vl["ECMPRDNL2"]["ManualMode"] == 1:
       ret.gearShifter = self.parse_gear_shifter("T")
--- a/opendbc/car/body/carcontroller.py
+++ b/opendbc/car/body/carcontroller.py
@@ -45,12 +45,12 @@ class CarController(CarControllerBase):
       speed_desired = CC.actuators.accel / 5.
       speed_diff_desired = -CC.actuators.torque / 2.
 
-      speed_measured = SPEED_FROM_RPM * (CS.out.wheelSpeeds.fl + CS.out.wheelSpeeds.fr) / 2.
+      speed_measured = CS.out.vEgoRaw
       speed_error = speed_desired - speed_measured
 
       torque = self.wheeled_speed_pid.update(speed_error, freeze_integrator=False)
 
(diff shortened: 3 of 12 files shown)


opendbc/car/car.capnp
```diff
@@ -220,15 +209,11 @@ struct CarState {
...
   rightBlinker @21 :Bool;
-  genericToggle @23 :Bool;
 
   # lock info
   doorOpen @24 :Bool;
   seatbeltUnlatched @25 :Bool;
 
-  # clutch (manual transmission only)
-  clutchPressed @28 :Bool;
-
```
sshane, line 231: is this used for USER_DISABLE?

opendbc/car/ford/carcontroller.py
```diff
@@ -108,7 +108,7 @@ def update(self, CC, CS, now_nanos):
         apply_curvature = actuators.curvature
 
       # apply rate limits, curvature error limit, and clip to signal range
-      current_curvature = -CS.out.yawRate / max(CS.out.vEgoRaw, 0.1)
+      current_curvature = 0  # TODO: calculate from alternative source, yawRate deprecated
```
sshane, line 111: this is for the safety. just pull it from CS.yaw_rate

opendbc/car/ford/carstate.py
```diff
@@ -34,15 +34,13 @@ def update(self, can_parsers) -> structs.CarState:
     # car speed
     ret.vEgoRaw = cp.vl["BrakeSysFeatures"]["Veh_V_ActlBrk"] * CV.KPH_TO_MS
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
-    ret.yawRate = cp.vl["Yaw_Data_FD1"]["VehYaw_W_Actl"]
```
sshane, line 37: `self.yaw_rate = cp.vl["Yaw_Data_FD1"]["VehYaw_W_Actl"]`

opendbc/car/toyota/carcontroller.py
```diff
@@ -103,8 +103,8 @@ def update(self, CC, CS, now_nanos):
     new_torque = int(round(actuators.torque * self.params.STEER_MAX))
     apply_torque = apply_meas_steer_torque_limits(new_torque, self.last_torque, CS.out.steeringTorqueEps, self.params)
 
-    # >100 degree/sec steering fault prevention
-    self.steer_rate_counter, apply_steer_req = common_fault_avoidance(abs(CS.out.steeringRateDeg) >= MAX_STEER_RATE, lat_active,
+    # >100 degree/sec steering fault prevention (steeringRateDeg deprecated, disabling check)
+    self.steer_rate_counter, apply_steer_req = common_fault_avoidance(False, lat_active,
```
sshane, line 107: this is a huge downgrade...

sshane:
I think it makes sense to start to remove unnecessary fields

opendbc/car/car.capnp
adeebshihadeh: apparently not
```python
(openpilot)  1 batman@workstation-adeeb2:~/fourpilot$ rg -i clutchpress selfdrive/
(openpilot)  1 batman@workstation-adeeb2:~/fourpilot$
```

opendbc/car/gm/carstate.py
```diff
@@ -68,16 +68,16 @@ def update(self, can_parsers) -> structs.CarState:
     # An Equinox has been seen with an unsupported status (3), so only check if either wheel is in reverse (2)
     left_whl_sign = -1 if pt_cp.vl["EBCMWheelSpdRear"]["RLWheelDir"] == 2 else 1
     right_whl_sign = -1 if pt_cp.vl["EBCMWheelSpdRear"]["RRWheelDir"] == 2 else 1
-    ret.wheelSpeeds = self.get_wheel_speeds(
+    wheel_speeds = self.get_wheel_speeds(
```
sshane, line 71: you can make get_wheel_speeds just return a tuple

jyoung8607:
```
  gasDEPRECATED @3 :Float32;        # this is user pedal only
  brakeDEPRECATED @5 :Float32;      # this is user pedal only
  engineRpmDEPRECATED @46 :Float32;
  genericToggleDEPRECATED @23 :Bool;
  cumLagMsDEPRECATED @50 :Float32;
  wheelSpeedsDEPRECATED @2 :WheelSpeeds;
```

I'm good with these. Wheel speeds are obviously safety relevant, but we keep the important outcomes in the form of `vEgoRaw` and `standstill`. Likewise for `gasPressed` and `brakePressed`.

```
  clutchPressedDEPRECATED @28 :Bool;
```

Weak preference to keep this. It's just a bool, and it'll be necessary later for a good OP long experience. Cars with ACC and MT tend to stay engaged for a few seconds while shifting, but we want to reject initial entry, and we'll want to freeze or even reset the gas/accel integrator. We're not doing these things yet, but I don't want to lose the prerequisite signals.

```
  steeringRateDegDEPRECATED @15 :Float32;
  yawRateDEPRECATED @22 :Float32;   # best estimate of yaw rate
```

Not stoked about losing convenient views of these, given their active use in safety/fault-mitigation in a few ports, but life would go on.

**Other fields**

* At a glance, `canErrorCounter` is recorded by `card` but doesn't seem like it's doing anything useful
* `steeringAngleOffsetDeg` is Toyota only and could just live there, but also safety relevant, hmm
* Can we consolidate `vCruise` and `cruiseState.speed`?
* Likewise `vCruiseCluster` and `cruiseState.speedCluster`
* `cruiseState.speedOffset` appears to be unused

**Other ideas**
* Some of our Float32s **might** be represented sufficiently with Float16s
* Perhaps consider a new `carSensors` type of object for nice-but-optional data, exclude it from qlogs

adeebshihadeh:
> 
> ```
>   clutchPressedDEPRECATED @28 :Bool;
> ```
> 
> Weak preference to keep this.

Removed for now. Easy enough to add back if necessary.

> * Can we consolidate `vCruise` and `cruiseState.speed`?
> * Likewise `vCruiseCluster` and `cruiseState.speedCluster`

Potentially. There is a meaningful difference in that `cruiseState` is the car's own state. Regardless of whether we decide to keep it, this difference is not at all clear, and that should be fixed.

> * Some of our Float32s **might** be represented sufficiently with Float16s

`float32` is the smallest type in capnproto.

> * Perhaps consider a new `carSensors` type of object for nice-but-optional data, exclude it from qlogs

So the qlogs were really just a bonus. It's more about separating the necessary vs the nice-to-haves, and at the very least, I think we need a different file if we want to bring back the nice-to-haves.

adeebshihadeh:
Alright, pretty much done with the first pass at this.

It's clear that there's lot more to do here on a per-port basis. Some examples:
* do we really need this weighting? https://github.com/commaai/opendbc/blob/de7d108f443e91b7bf040e40ebf22fb4d86b8025/opendbc/car/honda/carstate.py#L159 
* can we remove `carState.standstill` or just define it as `vEgoRaw < 0.1`?
* `carState.wheelSpeeds` can go once the body doesn't use it for closed loop control on the joystick

jyoung8607:
> Removed for now. Easy enough to add back if necessary.

Reasonable, and thanks for doing it standalone (#2510) so it's easy to cross that bridge when we come to it.
 
> do we really need this weighting?

Been awhile, and haven't looked at all Honda, but for at least one example the wheel speed sensor signals all drop to zero below about 2mph or thereabout. Not usable for FtS/SnG purposes. It may be possible to discard the wheel speeds entirely and go with the single transmission output speed signal.

> So the qlogs were really just a bonus. It's more about separating the necessary vs the nice-to-haves, and at the very least, I think we need a different file if we want to bring back the nice-to-haves.

I like this idea, as a future-someday thing.

> can we remove carState.standstill or just define it as vEgoRaw < 0.1?

Stock ACC resume spam is conditioned on standstill (commaai/openpilot#24873) and VW don't like to be spammed with the drivetrain coordinator in the wrong state (commaai/openpilot#22176). Haven't checked other ports.

PR #2503, opened 2025-07-21
Refactor steering angle limit functions and add common application logic

devtekve:


opendbc/car/tesla/values.py
```diff
@@ -85,20 +85,20 @@ class CANBUS:
...
   ACCEL_MIN = -3.48  # m/s^2
   JERK_LIMIT_MAX = 4.9  # m/s^3, ACC faults at 5.0
   JERK_LIMIT_MIN = -4.9  # m/s^3, ACC faults at 5.0
 
+  ANGLE_LIMITS: AngleSteeringLimits = AngleSteeringLimits(
+    360,  # deg, EPAS faults above this angle
+    # Tesla uses a vehicle model instead, check carcontroller.py for details
+    ([], []),
+    ([], []),
+    (1 / (DT_CTRL * STEER_STEP)),  # 50 Hz
```
sshane: this is redundant information

devtekve: > this is redundant information

The frequency? Or the 4 lines ? What makes it redundant? Is it because of the way it’s being calculated instead of simply putting the hz? 

Thanks for looking at this 👍

opendbc/safety/tests/test_tesla.py
```diff
@@ -33,7 +34,7 @@ class TestTeslaSafetyBase(common.PandaCarSafetyTest, common.AngleSteeringSafetyT
   GAS_PRESSED_THRESHOLD = 3
 
   # Angle control limits
-  STEER_ANGLE_MAX = 360  # deg
+  STEER_ANGLE_MAX = CarControllerParams.ANGLE_LIMITS.STEER_ANGLE_MAX  # deg
```
devtekve: I have 2 voices in my head 😄 :
1. Test should have it static, so don't take it from values.
2. The test should help validate the existing setup, so it should follow what's set on setup :D 

```suggestion
  STEER_ANGLE_MAX = 360  # deg
```

I am fine either way, so feel free to apply the suggestion or discard it entirely 😅

opendbc/car/tesla/values.py
devtekve: We can: 

1. Simply put 50 there as value instead of the calculation (I prefer that, but kept it as you had it) 
2. We can make a new `AngleSteeringLimits` named something like `VmAngleSteeringLimits` with only the relevant parts for this control scheme?
3. Make the caller provide it independently on every call to the methods instead of bundled as the `AngleSteeringLimits`?

Let me know which one should we do or if you had something else in mind @sshane

sshane: I mean `STEER_STEP` also tells you the frequency

devtekve: I am not seeing the change you want. Can you be more specific please? Do you want me to add STEER_STEP into the angle limits instead of frequency ?

devtekve: well I've made it just "50" for tesla. I need this so that I can pass it down to the angle steering calulation without having to add infinite amount of arguments to the methods

devtekve:
@adeebshihadeh @sshane can we get this one moving to progress better with the HKG Angle Steering? thanks!

opendbc/car/__init__.py
```diff
@@ -7,14 +8,24 @@
...
+from opendbc.car.vehicle_model import VehicleModel
 
 DT_CTRL = 0.01  # car state and control loop timestep (s)
 
 # kg of standard extra cargo to count for drive, gas, etc...
 STD_CARGO_KG = 136.
 
+# ISO 11270
+ISO_LATERAL_ACCEL = 3.0  # m/s^2
+ISO_LATERAL_JERK = 5.0  # m/s^3
```
adeebshihadeh: these are already defined in opendbc.car.interfaces

adeebshihadeh (changes requested):
Shane's comment still isn't addressed, and the first thing I see when reviewing this PR is duplicated code.

If you're going to directly ping us like this and ask for a prompt review, please make sure it's 100% ready to go. If that bar isn't clear, happy to talk about it on the dev call.

It's also fine if you aren't interested in hitting that bar, but please don't ping us in that case.

opendbc/car/__init__.py
devtekve: I could swear I had removed them from there..

devtekve: Gone. it is no longer on interfaces, I wanted to remove it form there, otherwise it'd be a circular dependency tho

opendbc/car/__init__.py
```diff
@@ -100,6 +114,41 @@ class Bus(StrEnum):
...
+
+
+def get_max_angle(v_ego_raw: float, VM: VehicleModel, max_lateral_accel: float = MAX_LATERAL_ACCEL):
+  """ Calculate the maximum steering angle based on lateral acceleration limits. """
+  max_curvature = max_lateral_accel / (v_ego_raw ** 2)  # 1/m
+  return math.degrees(VM.get_steer_from_curvature(max_curvature, v_ego_raw, 0))  # deg
+
+
+def apply_common_steer_angle_limits(apply_angle: float, apply_angle_last: float, v_ego_raw: float, steering_angle: float, lat_active: bool,
+                                    limits: AngleSteeringLimits, steer_step: int, VM: VehicleModel) -> float:
```
sshane:
```suggestion
                                    limits: CarControllerParams, VM: VehicleModel) -> float:
```

like torque limiting

opendbc/car/tesla/carcontroller.py
```diff
@@ -1,54 +1,11 @@
...
-from opendbc.car import ACCELERATION_DUE_TO_GRAVITY, Bus, AngleSteeringLimits, DT_CTRL, rate_limit
-from opendbc.car.interfaces import CarControllerBase, ISO_LATERAL_ACCEL
+from opendbc.car import Bus, apply_common_steer_angle_limits
+from opendbc.car.interfaces import CarControllerBase
 from opendbc.car.tesla.teslacan import TeslaCAN
 from opendbc.car.tesla.values import CarControllerParams
 from opendbc.car.vehicle_model import VehicleModel
 
-# limit angle rate to both prevent a fault and for low speed comfort (~12 mph rate down to 0 mph)
-MAX_ANGLE_RATE = 5  # deg/20ms frame, EPS faults at 12 at a standstill
```
sshane, line 12: This is to prevent a fault, not for safety. I think you can keep it here

opendbc/car/__init__.py
sshane: make sure you clearly distinguish these angle functions from the existing breakpoint-based function that will be deprecated in the future. it's not clear they're separate or different in any meaningful way

> `apply_std_steer_angle_limits` vs `apply_common_steer_angle_limits`? oh let me just delete one!

devtekve: I was going for that but again, didn't feel convinced given that it is not inheriting from anywhere which makes it too loose, although python is happy with that. I will comply but I don't like it too much 😅

devtekve: will add the `_vm_` to them

opendbc/safety/tests/test_tesla.py
sshane: you can propose this change, but it's separate from this PR

opendbc/car/__init__.py
devtekve: going for `apply_vm_steer_angle_limits`

devtekve: Done

devtekve: Done

sshane: yeah just copy what safety.h calls it

sshane: I've wanted to make CarControllerParamsBase for a while, can you do that in a separate PR?

opendbc/car/__init__.py
```diff
@@ -7,14 +8,24 @@
...
 
+# ISO 11270
+ISO_LATERAL_ACCEL = 3.0  # m/s^2
+ISO_LATERAL_JERK = 5.0  # m/s^3
+
 ACCELERATION_DUE_TO_GRAVITY = 9.81  # m/s^2
 
+# Add extra tolerance for average banked road since safety doesn't have the roll
+AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
+MAX_LATERAL_ACCEL = ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL)  # ~3.6 m/s^2
```
sshane: This shouldn't be a global in `/car/__init__.py`. It can be very confusing for users to know what to import/use

opendbc/car/__init__.py
```diff
@@ -25,6 +36,8 @@ class AngleSteeringLimits:
   STEER_ANGLE_MAX: float
   ANGLE_RATE_LIMIT_UP: tuple[list[float], list[float]]
   ANGLE_RATE_LIMIT_DOWN: tuple[list[float], list[float]]
+  MAX_LATERAL_ACCEL: float = MAX_LATERAL_ACCEL  # Used primarily to lower the limits for more conservative approaches.
+  MAX_LATERAL_JERK: float = MAX_LATERAL_JERK  # Used primarily to lower the limits for more conservative approaches.
```
sshane, line 40: What is the point of this? You don't use it? I'd argue you should though, and move the above constants I just commented on to tesla/values.py

opendbc/car/__init__.py
```diff
@@ -100,6 +113,44 @@ class Bus(StrEnum):
   ap_party = auto()
 
 
+def get_max_angle_delta_vm(v_ego_raw: float, steer_step: int, VM: VehicleModel, max_lateral_jerk: float = MAX_LATERAL_JERK):
```
sshane: less defaults scattered around are better

opendbc/car/__init__.py
```diff
@@ -100,6 +113,44 @@ class Bus(StrEnum):
...
+def apply_vm_steer_angle_limits(apply_angle: float, apply_angle_last: float, v_ego_raw: float, steering_angle: float, lat_active: bool,
+                                limits, VM: VehicleModel, MAX_ANGLE_RATE: float = None) -> float:
+  """Apply rate, accel, and safety limit constraints to steering angle."""
+  v_ego_safe = max(v_ego_raw, 1.)
+
+  # *** max lateral jerk limit ***
+  max_angle_delta = get_max_angle_delta_vm(v_ego_safe, limits.STEER_STEP, VM, limits.ANGLE_LIMITS.MAX_LATERAL_JERK)
+
+  if MAX_ANGLE_RATE is not None:
+    max_angle_delta = min(max_angle_delta, MAX_ANGLE_RATE)
```
sshane: comment please. something like `Prevent fault on cars that need it`

opendbc/car/__init__.py
```diff
@@ -100,6 +113,44 @@ class Bus(StrEnum):
...
+
+  # *** max lateral accel limit ***
+  max_angle = get_max_angle_vm(v_ego_safe, VM, limits.ANGLE_LIMITS.MAX_LATERAL_ACCEL)
+  new_apply_angle = np.clip(new_apply_angle, -max_angle, max_angle)
+
+  # angle is current angle when inactive
+  if not lat_active:
+    new_apply_angle = steering_angle
+
+  # *** final hard limit to prevent fault ***
```
sshane: like this one
```suggestion
  # final hard limit to prevent fault
```

opendbc/car/__init__.py
sshane: It's important to scope things to the minimum required level until we need otherwise for maintainability/readability

opendbc/car/__init__.py
sshane: How does this not fail? These have no defaults

opendbc/car/__init__.py
sshane, line 1: This file is huge. I think we should split it out like we have done for safety. @adeebshihadeh do you like simply `lateral.py` and `longitudinal.py`?

devtekve: I am not named Adeeb, but I was tempted with this as well

devtekve: but pleeeeeeeeease, let's do that on another PR

opendbc/car/__init__.py
devtekve: You don't have these methods defined there  you do it differently so not 1:1 

```cpp
const float max_curvature_rate_sec = MAX_LATERAL_JERK / (fudged_speed * fudged_speed);
const float max_angle_rate_sec = get_angle_from_curvature(max_curvature_rate_sec, curvature_factor, params);

// finally get max angle delta per frame
const float max_angle_delta = max_angle_rate_sec / (float)limits.frequency;
```

Are we okay with the current names I did? 
* `apply_vm_steer_angle_limits`
* `get_max_angle_delta_vm`
* `get_max_angle_vm`

opendbc/car/__init__.py
adeebshihadeh: yeah, I was thinking the same thing actually

opendbc/car/__init__.py
devtekve: Well, this is actually a significant amount of work. Some brands have the class as if it was dataclass, others have it as a regular class with init... Not all the values on them are meant to be common... It's tough :D

opendbc/car/__init__.py
devtekve: Well I realized CarControllerParamsBase would be easier if we do this first. So... Is this what you had in mind? https://github.com/commaai/opendbc/pull/2594

devtekve: done 
* commaai/opendbc/pull/2594

opendbc/car/__init__.py
devtekve: with the split of init and lateral this has another chance of succeeding so I'll try again that

opendbc/car/__init__.py
devtekve: moving this to `lateral.py` feels like a more sensible approach, do you agree @sshane ?

opendbc/car/__init__.py
devtekve: the `CarControllerParamsBase` will be done on a separate PR but will take it for later, I think right now with the changes on the split this PR is much more manageable.

opendbc/car/interfaces.py
```diff
@@ -26,10 +26,6 @@
 ACCEL_MAX = 2.0
 ACCEL_MIN = -3.5
 
-# ISO 11270
-ISO_LATERAL_ACCEL = 3.0  # m/s^2
-ISO_LATERAL_JERK = 5.0  # m/s^3
-
```
devtekve, line 32: I had to move this if I want to be able to use them from `values.py` because otherwise I was getting a circular dependency

devtekve:
converting this to draft again until we address https://github.com/commaai/openpilot/pull/35866 and https://github.com/commaai/openpilot/pull/35866 since they caused changes in process replay due to float precision. So I would like to get those sorted out before this PR to avoid mixing things up

devtekve:
> converting this to draft again until we address [commaai/openpilot#35866](https://github.com/commaai/openpilot/pull/35866) and [commaai/openpilot#35866](https://github.com/commaai/openpilot/pull/35866) since they caused changes in process replay due to float precision. So I would like to get those sorted out before this PR to avoid mixing things up

alright, this is done now ! After splitting gravity we now have it consistently working

adeebshihadeh:
Looks pretty clean to me now, but will let @sshane do the final review/merge.

opendbc/car/tesla/carcontroller.py
```diff
@@ -79,9 +39,8 @@ def update(self, CC, CS, now_nanos):
 
     if self.frame % 2 == 0:
       # Angular rate limit based on speed
-      self.apply_angle_last = apply_tesla_steer_angle_limits(actuators.steeringAngleDeg, self.apply_angle_last, CS.out.vEgoRaw,
-                                                             CS.out.steeringAngleDeg, lat_active,
-                                                             CarControllerParams.ANGLE_LIMITS, self.VM)
+      self.apply_angle_last = apply_vm_steer_angle_limits(actuators.steeringAngleDeg, self.apply_angle_last, CS.out.vEgoRaw, CS.out.steeringAngleDeg,
+                                                          lat_active, CarControllerParams, self.VM, MAX_ANGLE_RATE)
```
adeebshihadeh, line 43: note for followup: `MAX_ANGLE_RATE` should probably go in `CarControllerParams`

opendbc/car/lateral.py
```diff
@@ -1,15 +1,26 @@
...
+ISO_LATERAL_JERK = 5.0  # m/s^3
+
 
 @dataclass
 class AngleSteeringLimits:
   STEER_ANGLE_MAX: float
   ANGLE_RATE_LIMIT_UP: tuple[list[float], list[float]]
   ANGLE_RATE_LIMIT_DOWN: tuple[list[float], list[float]]
+  MAX_LATERAL_ACCEL: float = ISO_LATERAL_ACCEL * .7  # ~2.1m/s^2, 70% ISO limits as default, but ideally this is set manually for comfortable limits?
+  MAX_LATERAL_JERK: float = ISO_LATERAL_JERK * .5  # ~2.5m/s^3, 50% of ISO limits as default, but ideally this is set manually for comfortable limits?
```
sshane, line 23: Where did these come from? It shouldn't be possible for cars that aren't tested with this function yet to be able to use it by default

opendbc/car/lateral.py
```diff
@@ -1,15 +1,26 @@
...
 import numpy as np
 from dataclasses import dataclass
-from opendbc.car import structs
+from opendbc.car import structs, rate_limit, DT_CTRL
+from opendbc.car.vehicle_model import VehicleModel
 
 FRICTION_THRESHOLD = 0.3
 
+# Add extra tolerance for average banked road since safety doesn't have the roll
+AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
```
sshane, line 10: This should be namespaced/prefixed with angle so it's clear what this is for

opendbc/car/lateral.py
devtekve: Some conservative values / sensible defaults since `AngleSteeringLimits` seems used by cars that did not use the VM methods. Whether `safety` would allow them or not would be another topic... Should we do `None` instead?

PR #2507, opened 2025-07-21
Chevy Bolt: lower lat accel factor

sshane:
https://github.com/commaai/opendbc/pull/2506 should fix the heavy potential oscillations, but also re-introduced undershooting in turns. This should fix that. Note that friction is artificially kept low to mask torque controller bug. torqued actually learns values around what we set it to, but can introduce oscillations over bumps


sshane:
@ZwX1616 actually since the Bolt uses the nano model for ff, latAccelFactor does nothing lol. Your change only adjusted the friction back then

PR #2508, opened 2025-07-21
Slimming `carState`, part 2

adeebshihadeh:
Can't remove these as they're used for control in some cars, but no need to parse it out on the other cars.

This even resolves this Honda TODO:
https://github.com/commaai/opendbc/blob/cdf3e4a199995cc46b4169eadb8920135b45a7c3/opendbc/car/honda/carstate.py#L33


sshane:
Eventually we may want to limit Honda torque

adeebshihadeh:
Let's worry about future concerns in the future!

PR #2511, opened 2025-07-21
torque controller: fix non-linear friction

sshane:
Validation
* Dongle ID: 
* Route:


sshane:
Superseded by https://github.com/commaai/opendbc/pull/2528

PR #2519, opened 2025-07-22
hypothesis: use valid CAN DLCs

sshane:


opendbc/car/tests/test_car_interfaces.py
```diff
@@ -21,13 +21,16 @@
 
 ALL_REQUESTS = {tuple(r.request) for config in FW_QUERY_CONFIGS.values() for r in config.requests}
 
+# From panda/python/__init__.py
```
adeebshihadeh, line 24: fyi this isn't a panda thing: https://elearning.vector.com/mod/page/view.php?id=368

PR #2528, opened 2025-07-22
Torque controller: refactor calculations to be in accel space

sshane:
friction now in lat accel space


opendbc/car/__init__.py
```diff
@@ -203,10 +203,11 @@ def rate_limit(new_value, last_value, dw_step, up_step):
 
 def get_friction(lateral_accel_error: float, lateral_accel_deadzone: float, friction_threshold: float,
                  torque_params: structs.CarParams.LateralTorqueTuning) -> float:
+  friction_lateral_accel = torque_params.friction / torque_params.latAccelFactor
```
sshane:
```suggestion
  # TODO: make all the params to be in lat accel space. required torqued changing how it outputs friction
  friction_lateral_accel = torque_params.friction / torque_params.latAccelFactor
```

PR #2531, opened 2025-07-23
CANParser: prep for dynamic parsing

adeebshihadeh:


adeebshihadeh:
This is gonna be a mess to do in the Cython one -- will just do the Python rewrite first.

PR #2540, opened 2025-07-23
Toyota: fix silent frequency truncation

adeebshihadeh:
This got silently truncated to zero with the old CANParser (float -> int), but new one catches this https://github.com/commaai/opendbc/pull/2539.

Will add checks once new CANParser is in.

### DIFF
--- a/opendbc/car/toyota/carstate.py
+++ b/opendbc/car/toyota/carstate.py
@@ -192,7 +192,7 @@ class CarState(CarStateBase):
   def get_can_parsers(CP):
     pt_messages = [
       ("LIGHT_STALK", 1),
-      ("BLINKERS_STATE", 0.15),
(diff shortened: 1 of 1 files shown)


opendbc/car/toyota/carstate.py
```diff
@@ -192,7 +192,7 @@ def update(self, can_parsers) -> structs.CarState:
   def get_can_parsers(CP):
     pt_messages = [
       ("LIGHT_STALK", 1),
-      ("BLINKERS_STATE", 0.15),
+      ("BLINKERS_STATE", 0),
```
sshane, line 195: set to 1 and run test_models?

adeebshihadeh: I want to get my refactor in first

PR #2549, opened 2025-07-24
Use dynamic CANParser for all radar interfaces

adeebshihadeh:


adeebshihadeh:
https://github.com/commaai/opendbc/issues/2605 is a prereq.

PR #2552, opened 2025-07-24
safety: misc cleanups

adeebshihadeh:


opendbc/safety/modes/body.h
```diff
@@ -3,9 +3,6 @@
 #include "opendbc/safety/safety_declarations.h"
 
 static void body_rx_hook(const CANPacket_t *to_push) {
-  // body is never at standstill
-  vehicle_moving = true;
```
sshane, line 7: please remove from carstate if you change safety state, fails test models. we really need this test in opendbc ASAP!

PR #2559, opened 2025-07-25
Honda: EPS fault value updates

jyoung8607:
Picked from #1967 and #2488 for separate consideration.

The existing code is weird. Every steerFaultPermanent is *also* steerFaultTemporary and we also fault for previously undefined states. One might consider that a feature, but no other port does it. It would be better to grow explicit monitoring of steering-ready and steering-active for all ports.

- [ ] DRIVER_STEERING would have been treated as a fault before, see if this is present in commaCarSegments
- [ ] Make this logic a bit more conventional

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -83,10 +83,10 @@ class CarState(CarStateBase):
     ret.seatbeltUnlatched = bool(cp.vl["SEATBELT_STATUS"]["SEATBELT_DRIVER_LAMP"] or not cp.vl["SEATBELT_STATUS"]["SEATBELT_DRIVER_LATCHED"])
 
     steer_status = self.steer_status_values[cp.vl["STEER_STATUS"]["STEER_STATUS"]]
-    ret.steerFaultPermanent = steer_status not in ("NORMAL", "NO_TORQUE_ALERT_1", "NO_TORQUE_ALERT_2", "LOW_SPEED_LOCKOUT", "TMP_FAULT")
+    ret.steerFaultPermanent = steer_status not in ("NORMAL", "NO_TORQUE_ALERT_1", "NO_TORQUE_ALERT_2", "LOW_SPEED_LOCKOUT", "TMP_FAULT", "DRIVER_STEERING")
     # LOW_SPEED_LOCKOUT is not worth a warning
     # NO_TORQUE_ALERT_2 can be caused by bump or steering nudge from driver
-    ret.steerFaultTemporary = steer_status not in ("NORMAL", "LOW_SPEED_LOCKOUT", "NO_TORQUE_ALERT_2")
+    ret.steerFaultTemporary = steer_status not in ("NORMAL", "LOW_SPEED_LOCKOUT", "NO_TORQUE_ALERT_2", "DRIVER_STEERING")
 
     if self.CP.carFingerprint in HONDA_BOSCH_RADARLESS:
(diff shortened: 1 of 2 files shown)


jyoung8607:
@mvl-boston Does DRIVER_STEERING mean the EPS won't accept LKA control, or is it just informational?

vanillagorillaa:
> @mvl-boston Does DRIVER_STEERING mean the EPS won't accept LKA control, or is it just informational?

FWIW, https://github.com/commaai/opendbc/pull/2104#issuecomment-2800153147 and https://github.com/commaai/opendbc/pull/2240 

"Bosch" vehicles (at least tested on a 2018 Accord) will still accept commands in fault state 1, "Bosch Radarless" (at least it appeared that way on @MarcoTheDingo 's 2022 Civic) do not accept command in fault state 1

mvl-boston:
It  definitely accepts, but I think it ignores as I don't remember feeling
the nudges.

jyoung8607:
> FWIW, [#2104 (comment)](https://github.com/commaai/opendbc/pull/2104#issuecomment-2800153147) and #2240

VERY interesting, wasn't aware of that and would love to see it get back in. Thanks!
 
> "Bosch" vehicles (at least tested on a 2018 Accord) will still accept commands in fault state 1, "Bosch Radarless" (at least it appeared that way on @MarcoTheDingo 's 2022 Civic) do not accept command in fault state 1

Has anyone tried sending that bit to an ALT_RADAR car? Might bypass their min steer speed.

mvl-boston:
On my phone so hard to see links: if you are asking about settozero bit, it
did not help on the 22rdx nor odyssey.

jyoung8607:
@mvl-boston Does LOW_SPEED_LOCKOUT work on ALT_RADAR cars, or anything else on the outstanding PR list that might have a non zero minimum steer speed?
 
https://github.com/commaai/opendbc/blob/fdd0c5a37dff55aadea3a8b47ff51f1b7f5595ad/opendbc/car/honda/carstate.py#L90

I'm not seeing the code in the Honda port (or openpilot car_specific) to alert when below minimum steer speed. We alert on this for other brands. Not sure why it's missing, maybe just because it's the very oldest car port, but that could be added with some appropriate nuisance filtering.

@vanillagorillaa I know you've moved on, but as a Honda community elder who seems to be watching GitHub still, your thoughts would be valuable. Especially any interaction with #2104, which I'd love to resurrect within its usable scope.

mvl-boston:
> @mvl-boston Does LOW_SPEED_LOCKOUT work on ALT_RADAR cars, or anything else on the outstanding PR list that might have a non zero minimum steer speed?
> 
> https://github.com/commaai/opendbc/blob/fdd0c5a37dff55aadea3a8b47ff51f1b7f5595ad/opendbc/car/honda/carstate.py#L90
> 
> I'm not seeing the code in the Honda port (or openpilot car_specific) to alert when below minimum steer speed. We alert on this for other brands. Not sure why it's missing, maybe just because it's the very oldest car port, but that could be added with some appropriate nuisance filtering.
> 
> @vanillagorillaa I know you've moved on, but as a Honda community elder who seems to be watching GitHub still, your thoughts would be valuable. Especially any interaction with #2104, which I'd love to resurrect within its usable scope.

My daily driver steers at all speeds, along with any recent loaners I've driven.  The only limited experience I have is the few times I turn long off on the RDX for testing.  Based on this limited experience, I think https://github.com/commaai/opendbc/pull/2473 is sufficient for an alert.

Issue #1102, opened 2023-05-31
Civic: permanent LKAS fault when jerking wheel


sshane:
Could be a diff of torque, max torque, or angle rate

https://connect.comma.ai/f299d0a03c4e965d/1685509313071/1685509343277

https://connect.comma.ai/f299d0a03c4e965d/1685509362131/1685509446705

https://connect.comma.ai/f299d0a03c4e965d/1685509216391/1685509226635

jyoung8607:
I suspect we have angle-rate limits on some of the Hondas we're working to merge now, with similar permanent faults. We'll be testing an adaptation of the common fault avoidance code in Toyota and Subaru.

If that helps, and the limits aren't very onerous, we may just apply them to all Honda and resolve this as well.

mvl-boston:
> If that helps, and the limits aren't very onerous, we may just apply them to all Honda and resolve this as well.

All cars with this issue reported are Bosch A.   So whatever fix is found, I don't think it needs to be expanded beyond Bosch A.

vanillagorillaa:
> > If that helps, and the limits aren't very onerous, we may just apply them to all Honda and resolve this as well.
> 
> All cars with this issue reported are Bosch A. So whatever fix is found, I don't think it needs to be expanded beyond Bosch A.

No, this is incorrect. Shane's initial report, and my follow up investigations both come from Nidec Civics.

mvl-boston:
Ok, haven't seen it on B or C yet.  Drivers didn't like the jerk shutoffs I
made global in earlier attempts to fix this as they liked the support of
LKAS even while manual steering.  So I would still recommend limiting to
only impacted car models.

PR #2560, opened 2025-07-25
Honda: Bosch alternate radar support

jyoung8607:
Picked from #1967 and #2488 for separate consideration.

- [ ] Poke around for any alternative ways to obtain lower speed control, may not be possible
- [x] Review how this will change with OP long
- [x] #2562
- [ ] Any better way to handle aggressive driver overrides? Temp fault or drop to zero on DRIVER_STEERING?


mvl-boston:
With OP long, the STEER_CONTROL_ACTIVE bit will remain at 1 down to 4mph, and drop to zero at 3mph (losing steering).   This seemed too low a speed to alert the driver of lkas loss.

Without OP long, this same behavior occurs under 40mph, and once dropping the bit to zero it will not reset back to 1 until 45mph.   This is why carspecs are set at 45mph so that built in warning that says no steer under 45mph is accurate.   This higher speed cutoff is hardcoded in the radar, and because Honda stock-long relays commands via the radar we can't seem to defeat it.

Let me know if any further questions.   The coding approach here seems appropriate.

mvl-boston:
Also, do you think csouers carfootnote should be added?
https://github.com/csouers/opendbc/blob/4171e87556c75ea300f89f8e3dbfd0fdba55029f/opendbc/car/honda/values.py#L108

jyoung8607:
> because Honda stock-long relays commands via the radar we can't seem to defeat it.

Has someone actually seen this on a wiring diagram? Can you point me toward working long control code for this configuration?

opendbc/car/honda/carstate.py
```diff
@@ -111,6 +112,18 @@ def update(self, can_parsers) -> structs.CarState:
...
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
 
+    if self.CP.carFingerprint in HONDA_BOSCH_ALT_RADAR:
+      ret.lowSpeedAlert = False
+      # under 4mph not worth warning, but don't create falling edge if already in warning state
+      if ret.cruiseState.enabled and (cp.vl["STEER_STATUS"]["STEER_CONTROL_ACTIVE"] != 1) and (self.steer_blocked_prev or ret.vEgo > 4.0 * CV.MPH_TO_MS):
+        if ret.vEgo <= self.CP.minSteerSpeed:
+          ret.lowSpeedAlert = True
+        else:
+          ret.steerFaultTemporary = True
```
jyoung8607, line 122: Can you help me understand how we arrive at line 122? Accepting the various speed limitations for now, is there a situation where we're at a supported speed and there is neither a temporary nor a permanent steer fault and yet the EPS isn't acting on control requests? How does this get triggered?

mvl-boston: The car does steer cutoff with wheel jerks or prolonged driver wheel overrides.   That's the need for line 122.   There seems to be some type of cooldown where it reestablishes steer after 30 seconds or so.    Usually cruise disengage/reengage can wake steer up sooner.   And these drivers are asking for some type of notification (eg: pr 2473) so they know when this is occurring.

mvl-boston:
> > because Honda stock-long relays commands via the radar we can't seem to defeat it.
> 
> Has someone actually seen this on a wiring diagram? Can you point me toward working long control code for this configuration?

This occurs in all Bosch A cars, including ones already upstreamed.   LKAS goes to the radar when it is on and radar relays to powertrain.   And LKAS goes direct to powertrain when radar is off.   I presume powertrain faults on dupe signals on the canbus, and with the harness not in line with radar we have no way of stopping radar LKAS.

https://github.com/commaai/opendbc/blob/5e58025dc5882e68ccfb5fafabc474ad908ee83d/opendbc/car/honda/hondacan.py#L19

ALT_RADAR models are unique that the radar locks out the LKAS relay under stock LKAS speeds.   Honda's low-speed LKAS is a "feature" they sell as "traffic jam assist", as a way to get people to buy higher end models (eg: MDX has traffic jam assist, RDX doesn't).   Seems like this is a deliberate nerf to enforce the upsell.

Bosch C seems to allow direct-to-powertrain LKAS while allowing the car to ignore radar inputs.   I can ask the ALT_RADAR testers to try again, but it seems this was an issue for years so never tried to re-validate.

jyoung8607:
@mvl-boston Do you have a test route that demonstrates the wheel jerk issue? If not, can you have one made?

I think we may be rediscovering #1102, and we just need to solve it generally for Honda. If that's true, all we really need for ALT_RADAR cars is dynamic min steer speeds.

mvl-boston:
Agree that bug describes the trigger well.  But the symptom has has two classes of cars:
1) Odyssey 21-25 and RDX 22-24:  Symptom is it causes a steer_control_active=0.  LKAS disables but EPS remains.   There is no warning currently on the dash when this occurs, hence the pr2473 ask.  Driver can usually reenable LKAS by disengage/reengage of cruise, or waiting for about 30 seconds.   I also got a few reports of already-upsteamed RDX 2019-2021 drivers experiencing this.
2) RDX 2025 (their second mid-model refresh), seems to match the 2019-21 Civic from bug 1102: Symptom is that EPS dies with a steer_status=7 and a flashing red "stop driving now" message on the dash.   This kills EPS meaning the driver needs the full force of manual steering (dangerous).   And EPS does not reengage until engine off/on.

I've tried for days in June during a RDX 2025 loaner to reduce frequency of this, and found the workarounds in pr2474 helped 90% of cases.  The one I can't seem to solve is a wheel jerk during hard braking (eg: quick lange change as you are stopping, wanting to switch to a lane with less backed-up cars).   Even disabling steer control during all hard braking still caused this fault.

ad9840558640c31d/00000047--5780f5aecd - this is the brake/change lane scenario I couldn't fix.

I went through all my other logs in useradmin and no other steer faults have rlogs.   Almost as if useradmin deletes stale rlogs, I really thought I downloaded more rlogs during that loaner cycle.

jyoung8607:
> `ad9840558640c31d/00000047--5780f5aecd`

I suspect this you might not have a torque-with-LKA limit, but an angle-rate-with-LKA limit. Looking at your route, at the point of the fault, there were times before that where you overrode with more force, but not at this speed.

image


This is somewhat common. We already need mitigation for this exact thing on Toyota and Subaru. They are nice clean implementations you can use as a model to see if it helps. Your test route kissed 90deg/s for one frame, not sure if that's your actual limit but I'd test with something more conservative. Say, 60deg/s and 10 frames. 


https://github.com/commaai/opendbc/blob/fcf851ae22f356fc7a30f9e07550d6bf40f5e4c2/opendbc/car/toyota/carcontroller.py#L28-L31

https://github.com/commaai/opendbc/blob/fcf851ae22f356fc7a30f9e07550d6bf40f5e4c2/opendbc/car/toyota/carcontroller.py#L106-L108


Don't worry about where it fits in just yet. Just hack it in, and see if this prevents your fault conditions. If it works, then we can refine the exact limits, merge it as the fix for #1102, and then all your car ports get simpler.

jyoung8607:
Superseded by #2685

PR #2567, opened 2025-07-26
rm GET_ADDR and GET_BUS

adeebshihadeh:
These macros are leftover from when we were using the STM's native CAN struct instead of our own nice CANPacket.


adeebshihadeh:
Merged these separately

PR #2569, opened 2025-07-26
Honda: CANFD long control

mvl-boston:
Adding alphalong control for Honda CANFD models.

Based on closed PR2162, rebaseling to current opendbc codebase, with addition of safety code.

validation route: ad9840558640c31d/00000195--03d04b0d1c

Additional note:
All newer Hondas (all CANFD and the MDX_4G) are sensitive to the unresolved TODO [here](https://github.com/commaai/openpilot/blob/bd176cb6a7c9aec51bb92577b7e3a6035a6986c6/openpilot/selfdrive/car/card.py#L229), which causes 8-15ms of dropped messages in Honda road testing.  Hondas will fault if the message drop is 13+ms, so there are occasional faults that can occur in this generation of alphalong (since it relies on ECU knockout).

I have successfully tested a large rewrite of Honda alphalong that fixes above, but I had planned to put it in a separate PR after this is merged.  If comma prefers I include it here, let me know.

The ECU knockout does not occur outside alphalong, so there is no impact to release branches.


jyoung8607:
- [ ] See if BRAKE_ERROR can be made global/common, it's present in a Nidec DBC as well
- [ ] #2638
- [ ] Drive testing
- [ ] Why does it fault on quick restarts?

jyoung8607:
> 1. Does comma3+red panda usage require bus 4 as written?

That's not necessary, each Panda only knows about what it's hooked up to.

> 2. This makes startup faults temporary as car does this if engine is restarted too soon after shutdown.  Is there another way to buffer flush at shutdown to prevent?

Will have to look and see what's broken.

jyoung8607:
> * [ ]  Why does it fault on quick restarts?

This *could* be connected to #2632. Not saying it is for sure, but Panda is dealing with a disruption that can back up queues and affect other buses as well, so I wouldn't spend a bunch more time on it until that's been resolved.

mvl-boston:
Reopening as this is testing well with the cruise fault workaround linked in the description.

mvl-boston:
In separate AEB testing, found that radar hud was previously blocked by safety.  Changed this PR to allow it through to properly displays that CMBS is off on the instrument cluster.  No driving impact noticed in testdrive.

PR #2577, opened 2025-07-28
packer & parser logging

adeebshihadeh:


opendbc/can/packer.py
```diff
@@ -43,6 +46,7 @@ def make_can_msg(self, name_or_addr, bus: int, values: dict[str, float]):
     else:
       msg = self.dbc.name_to_msg.get(name_or_addr)
       if msg is None:
+        carlog.error(f"msg not found for {name_or_addr=}")
```
sshane, line 49: we don't want to raise an exception yet?

any of these used to throw an exception before the Python rewrite?

adeebshihadeh: no they didn't. the problem is that these are conditional code paths (e.g. resume spam), otherwise it'd be fine to just throw an exception on startup

PR #2584, opened 2025-07-29
Test: Increase body test coverage

ssysm:


opendbc/safety/modes/body.h
```diff
@@ -16,8 +15,7 @@ static bool body_tx_hook(const CANPacket_t *msg) {
   }
 
   // Allow going into CAN flashing mode even if controls are not allowed
-  bool flash_msg = (msg->addr == 0x250U) && (GET_LEN(msg) == 8U);
-  if (!controls_allowed && (GET_BYTES(msg, 0, 4) == 0xdeadfaceU) && (GET_BYTES(msg, 4, 4) == 0x0ab00b1eU) && flash_msg) {
+  if (!controls_allowed && (GET_BYTES(msg, 0, 4) == 0xdeadfaceU) && (GET_BYTES(msg, 4, 4) == 0x0ab00b1eU)) {
```
adeebshihadeh: why remove this check?

ssysm: Originality thought that RxCheck will caught message that's does met condition `(msg->addr == 0x250U) && (GET_LEN(msg) == 8U);`, but since the CAN flasher *could* have bytes that aligns with the magic bytes, added a test to cover the check but removed the length check.

opendbc/safety/modes/body.h
```diff
@@ -29,8 +28,11 @@ static safety_config body_init(uint16_t param) {
     {.msg = {{0x201, 0, 8, 100U, .ignore_checksum = true, .ignore_counter = true, .ignore_quality_flag = true}, { 0 }, { 0 }}},
   };
 
-  static const CanMsg BODY_TX_MSGS[] = {{0x250, 0, 8, .check_relay = false}, {0x250, 0, 6, .check_relay = false}, {0x251, 0, 5, .check_relay = false},  // body
-                                        {0x1, 0, 8, .check_relay = false}};  // CAN flasher
+  static const CanMsg BODY_TX_MSGS[] = {
+    {0x250, 0, 8, .check_relay = false},
+    {0x250, 0, 6, .check_relay = false},
+    {0x251, 0, 5, .check_relay = false},  // body
+    {0x1, 0, 8, .check_relay = false}};  // CAN flasher
```
adeebshihadeh: no need to change this

opendbc/safety/tests/test_body.py
```diff
@@ -55,6 +55,15 @@ def test_can_flasher(self):
...
     self.assertFalse(self._tx(common.make_msg(0, 0x251, dat=b'\xce\xfa\xad\xde\x1e\x0b\xb0\x0a')))  # wrong address
 
+  def test_can_flasher_invalid(self):
+    # Invalid CAN flasher message
+    self.safety.set_controls_allowed(False)
+    self.assertTrue(self._tx(common.make_msg(0, 0x1, 8)))
+    self.assertFalse(self._tx(common.make_msg(0, 0x250, dat=b'\xce\xfa\xad\xde\x1e\x0b\xb0\x1a')))
+
+    self.assertTrue(self._tx(common.make_msg(0, 0x250, dat=b'\xce\xfa\xad\xde\x1e\x0b\xb0\x0a')))
+    # cover test but this could be a valid message(bytes in the CAN firmware)
```
adeebshihadeh: I don't understand this comment

adeebshihadeh (changes requested):

opendbc/safety/tests/test_body.py
ssysm: Just saying is asserting true for the test, and the reason why is true is that this could represent a real byte sequence in a can flasher/firmware payload.

PR #2585, opened 2025-07-29
Tesla: Switch condition check to switchcase and add default values

ssysm:


adeebshihadeh:
Can we do this in a more minimal way?

PR #2588, opened 2025-07-29
Honda: Steering jerk EPS fault mitigation

### LINKED ISSUE #1102: Civic: permanent LKAS fault when jerking wheel

jyoung8607:
Resolves #1102.

- [ ] Establish whether angle-rate limits are effective against the perm fault condition
- [ ] Establish the scope where we need to do this (Nidec, Bosch A, ?)
- [ ] Binary search for best parameters
- [ ] Cleanup

### DIFF
--- a/opendbc/car/honda/carcontroller.py
+++ b/opendbc/car/honda/carcontroller.py
@@ -12,7 +12,7 @@ LongCtrlState = structs.CarControl.Actuators.LongControlState
 
 # LKA angle-rate limits for avoiding EPS faults
 MAX_STEER_RATE = 60  # deg/s
-MAX_STEER_RATE_FRAMES = 10  # tx control frames needed before torque can be cut
+MAX_STEER_RATE_FRAMES = 3  # tx control frames needed before torque can be cut
 
 
 def compute_gb_honda_bosch(accel, speed):
--- a/opendbc/car/honda/carcontroller.py
+++ b/opendbc/car/honda/carcontroller.py
@@ -2,13 +2,17 @@ import numpy as np
 from collections import namedtuple
 
 from opendbc.can import CANPacker
-from opendbc.car import Bus, DT_CTRL, rate_limit, make_tester_present_msg, structs
+from opendbc.car import Bus, DT_CTRL, common_fault_avoidance, rate_limit, make_tester_present_msg, structs
 from opendbc.car.honda import hondacan
 from opendbc.car.honda.values import CruiseButtons, VISUAL_HUD, HONDA_BOSCH, HONDA_BOSCH_RADARLESS, HONDA_NIDEC_ALT_PCM_ACCEL, CarControllerParams
 from opendbc.car.interfaces import CarControllerBase
 
 VisualAlert = structs.CarControl.HUDControl.VisualAlert
 LongCtrlState = structs.CarControl.Actuators.LongControlState
+
+# LKA angle-rate limits for avoiding EPS faults
+MAX_STEER_RATE = 60  # deg/s
+MAX_STEER_RATE_FRAMES = 10  # tx control frames needed before torque can be cut
 
 
 def compute_gb_honda_bosch(accel, speed):
@@ -109,6 +113,7 @@ class CarController(CarControllerBase):
     self.apply_brake_last = 0
     self.last_pump_ts = 0.
     self.stopping_counter = 0
+    self.steer_rate_counter = 0
 
     self.accel = 0.0
     self.speed = 0.0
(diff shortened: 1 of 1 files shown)


jyoung8607:
@vanillagorillaa @mvl-boston Give this code a shot on cars that are going into a permanent EPS fault state. The 60 deg/s angle-rate is intended to be conservative, just to confirm we can actually prevent the fault.

If it works, try a binary search for the best available angle-rate. Change the limit to 120 deg/s and make sure the fault comes back, then change it to 90 deg/s and see what happens, etc etc until we've found the likely limit.

No matter what results you get, please upload and share routes of what happens.

jyoung8607:
Nick's Nidec Civic goes into a fault state really fast, 10 frames was too much allowance. I've cut the frame count to 3.

`39563a18fe07169f/000000e9--716972ebc2`

image

mvl-boston:
Was planning to test the different buses to see if I can get minspeed below
45mph first.  Hope to get to the mph build for the testers later tonight.

I recall from my log that I already cut steer and it still faulted.   Might
need to inspect what stock was attempting at that time to see if anything
different, or possibly pass through stock steer instead of sending steer
command off.   Will get whatever findings you have for the RDX/Odysshey
testers once minspeed is done.

mvl-boston:
Initial build didn't work for an Odyssey tester.   Can you see if this log commanded as you intended?  (eg: I integrated the Odyssey fingerprints/commands properly?)

805f87f5e96d128c/00000023--bd00f23e91/

edit: asking tester to try a new build at 40 deg/s, instead of 60 deg/s

mvl-boston:
Also looked at my own status 7 log from June, seems that the set_me values were different from stock when steer control was off, so I passed those through to see if it helps.

mvl-boston:
Second Odyssey log at 40 still caused issues.  
805f87f5e96d128c/0000002b--919828229d

There is a "no torque allowed" state lasts longer than the angle rate clearing.  Will write a new build to both try at 39 and wait for "no torque allowed" to clear before resuming.

mvl-boston:
Latest updates:
- Odyssey could not prevent all the way down to 39 angle change.   In process of testing 25 angle max.
- Also, adding a "don't steer if steer status <> normal" seems to have prevented the lockout on a 22 RDX, awaiting Odyssey results.
- I have a 25RDX loaner (wife is using but I can probably drive it tonight), and will try to get as much fault-prevention coded as I can tonight.

jyoung8607:
> * Also, adding a "don't steer if steer status <> normal" seems to have prevented the lockout on a 22 RDX, awaiting Odyssey results.

This might be the way. I was hoping to poke at this on the CANFD car, but it seems to stay in normal all the time. If there's no torque applied in some of those modes anyway, it might be best to just back off. I'd like to develop a better understanding of those states in #2559.

mvl-boston:
25 loaner still faulted status 7 despite all of the above remediation attempts.    Will need to look at the "good" wheel jerks vs "bad" one to see if anything different.  Log below if anything comes to mind.

ad9840558640c31d/00000014--d6a155b560

mvl-boston:
There were two things I noticed in the log:
- the "bad" wheeljerk had prior steering within 1 second of a lockout.   I'm going to try a 1-sec steer delay timer after any lockouts to see if it helps
- I noticed angle_rate hits its max at 1600 steer_control, and the pump actually weakens after that.   So I'm going to cut the max commanded rate to 1600 to see whether it helps settle the pump down.   That will also decrease the steerdelta since it is proportional to max, so might help too.

I'm trading cars with my wife today so I'm driving the 25RDX all day (or at least until the dealership tells me to return the loaner).

mvl-boston:
1 sec delay and 1600 even with torque controller didn't work.   Any other ideas?

ad9840558640c31d/00000002--5894bf6861

mvl-boston:
I returned the 25 RDX before my trip.  Couldn't find any working mitigations for it, There are a few other random steer bits sent by stock during ACC braking that might prevent faults, just need more dedicated time for trial/error.

I recommend at this point that we focus on the 22-24 RDX, plus the 21-25 Odyssey.  These cars don't do the status 7 eps disable, and the goal was to prevent their post-steer lkas lockout.  The angle prevention alone didn't work for these cars, but angle limit plus non-normal status did.   Next step will be to make a non-normal status only build, and ask testers if that is good enough to keep things simple.

jyoung8607:
> I recommend at this point that we focus on the 22-24 RDX, plus the 21-25 Odyssey. These cars don't do the status 7 eps disable, and the goal was to prevent their post-steer lkas lockout.

That's fine.

> The angle prevention alone didn't work for these cars, but angle limit plus non-normal status did. Next step will be to make a non-normal status only build, and ask testers if that is good enough to keep things simple.

Agreed. I have a feeling we'll have a better understanding of what states 2/3/4/5 really mean after this. Once you have something that sidesteps the loss-of-control conditions, even if it's ugly, I can help iterate on it.

mvl-boston:
> > I recommend at this point that we focus on the 22-24 RDX, plus the 21-25 Odyssey. These cars don't do the status 7 eps disable, and the goal was to prevent their post-steer lkas lockout.
> 
> That's fine.
> 
> > The angle prevention alone didn't work for these cars, but angle limit plus non-normal status did. Next step will be to make a non-normal status only build, and ask testers if that is good enough to keep things simple.
> 
> Agreed. I have a feeling we'll have a better understanding of what states 2/3/4/5 really mean after this. Once you have something that sidesteps the loss-of-control conditions, even if it's ugly, I can help iterate on it.

LKAS mitigation PR opened as https://github.com/commaai/opendbc/pull/2685

jyoung8607:
This approach didn't work. We'll iterate on the approach in #2685,

PR #2589, opened 2025-07-29
Test: Increase rivian test coverage

ssysm:
- Removed address & bus checks, should be done by panda safety 
- Removed branch that cannot be tested with coverage

### DIFF
--- a/opendbc/safety/modes/rivian.h
+++ b/opendbc/safety/modes/rivian.h
@@ -34,24 +34,53 @@ static uint8_t _rivian_compute_checksum(const CANPacket_t *msg, uint8_t poly, ui
   return crc ^ xor_output;
 }
 
+typedef struct {
+  uint32_t addr;
+  uint8_t poly;
+  uint8_t xor_output;
+} ChecksumConfig;
+
 static uint32_t rivian_compute_checksum(const CANPacket_t *msg) {
+  static const ChecksumConfig checksum_configs[] = {
+    {0x208U, 0x1D, 0xB1},  // ESP_Vehicle_Speed_Checksum
+    {0x150U, 0x1D, 0x9A},  // VDM_VehicleSpeed_Checksum
+  };
+
   uint8_t chksum = 0;
-  if (msg->addr == 0x208U) {
-    chksum = _rivian_compute_checksum(msg, 0x1D, 0xB1);
-  } else {
-    // msg->addr == 0x150U
-    chksum = _rivian_compute_checksum(msg, 0x1D, 0x9A);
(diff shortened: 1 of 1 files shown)


opendbc/safety/modes/rivian.h
```diff
@@ -43,9 +38,9 @@ static uint32_t rivian_compute_checksum(const CANPacket_t *msg) {
   uint8_t chksum = 0;
   if (msg->addr == 0x208U) {
     chksum = _rivian_compute_checksum(msg, 0x1D, 0xB1);
-  } else if (msg->addr == 0x150U) {
-    chksum = _rivian_compute_checksum(msg, 0x1D, 0x9A);
   } else {
+    // msg->addr == 0x150U
+    chksum = _rivian_compute_checksum(msg, 0x1D, 0x9A);
```
adeebshihadeh: we need to be careful to not let the tests make the actual code less readable. this is one of those cases; can you rewrite it in a better way?

opendbc/safety/modes/rivian.h
```diff
@@ -54,9 +49,9 @@ static bool rivian_get_quality_flag_valid(const CANPacket_t *msg) {
   bool valid = false;
   if (msg->addr == 0x208U) {
     valid = ((msg->data[3] >> 3) & 0x3U) == 0x1U;  // ESP_Vehicle_Speed_Q
-  } else if (msg->addr == 0x150U) {
-    valid = (msg->data[1] >> 6) == 0x1U;  // VDM_VehicleSpeedQ
   } else {
+    // msg->addr == 0x150U
```
adeebshihadeh: same here

opendbc/safety/modes/rivian.h
```diff
@@ -94,10 +89,8 @@ static void rivian_rx_hook(const CANPacket_t *msg) {
 
   if (msg->bus == 2U) {
     // Cruise state
-    if (msg->addr == 0x100U) {
```
adeebshihadeh, line 88: same here, i'd just bring this up into the same if as `if (msg->bus == 2U) {`

adeebshihadeh (changes requested):
While these changes might improve coverage, we can't trade-off readability for that, and we shouldn't need to.

Make sure to carefully consider that with all of these changes.

ssysm:
Changed up a bit, now checksum and quality check will go thru a config list instead of going thru a if-else logic. However, seems like MULL really doesn't like for-loop's `<` and `<=` on matching item. Added a falling-thru case for quality check and if somehow the code does mutate at runtime, it will simply return invalid.

PR #2591, opened 2025-07-30
Hyundai: make LFA and LKAS steering consistent

adeebshihadeh:
Fixes https://github.com/commaai/openpilot/issues/35683

---

For both steer control types, `DAMP_FACTOR` was set to 0, which is apparently invalid. However, the ADRV ECU seems to default this to a different value than the EPS. ADRV clips to the min (3), and the EPS seems to use around 100 if it gets an invalid value. It's possible there's a more optimal value, but starting with that since it's what the stock system uses.

In `5bef1e6eece568e8/0000006d--302abdb32f`, `LKAS/DAMP_FACTOR` is cycled such that we can see what the ADRV does:
image


opendbc/dbc/generator/hyundai/hyundai_canfd.dbc
```diff
@@ -98,6 +98,7 @@ BO_ 272 LKAS_ALT: 32 XXX
  SG_ ADAS_StrAnglReqVal : 82|14@1- (0.1,0) [0|176.7] "Deg" GW_RGW,MDPS,SFA
  SG_ ADAS_ACIAnglTqRedcGainVal : 96|8@1+ (0.004,0) [0|1] "" GW_RGW,MDPS,SFA
  SG_ NEW_SIGNAL_3 : 111|8@0+ (1,0) [0|255] "" XXX
+ SG_ DAMP_FACTOR : 64|8@1+ (1,0) [0|255] "" XXX
```
crwusiz, line 101: address 64 right.?

STEER_MODE has overlapping bits
NEW_SIGNAL_2 has overlapping bits

adeebshihadeh: want to write and upstream a test for this?

royjr:
my previous testing with damp control, trigger up = 3, down = 200

https://github.com/user-attachments/assets/6dbdb263-570f-463c-8b3b-87793bf380ba

PR #2592, opened 2025-07-30
Clarify that CoPilot-360 must be a version that includes ACC

alan-polk:
This PR updates the text string that gets displayed on the compatibility page of the website.  The Ford channel on discord continues to frequently see users who are angry because they have bought a Comma for their Ford equipped with Co-Pilot360, only to find out they have the wrong version of Co-Pilot360 and their car is not compatible.

### DIFF
--- a/opendbc/car/ford/values.py
+++ b/opendbc/car/ford/values.py
@@ -68,7 +68,7 @@ class Footnote(Enum):
 
 @dataclass
 class FordCarDocs(CarDocs):
-  package: str = "Co-Pilot360 Assist+"
+  package: str = "A version of Co-Pilot360 which includes adaptive cruise control such as Co-Pilot360 Assist or Co-Pilot360 Active"
   hybrid: bool = False
   plug_in_hybrid: bool = False
 


opendbc/car/ford/values.py
```diff
@@ -68,7 +68,7 @@ class Footnote(Enum):
 
 @dataclass
 class FordCarDocs(CarDocs):
-  package: str = "Co-Pilot360 Assist+"
+  package: str = "A version of Co-Pilot360 which includes adaptive cruise control such as Co-Pilot360 Assist or Co-Pilot360 Active"
```
sshane:
```suggestion
  package: str = "Adaptive Cruise Control"
```

If `Co-Pilot360 Assist+` includes packages without ACC on some cars, what about this?

alan-polk: Co-Pilot360 (No Assist/Active) = No ACC
Co-Pilot360 Assist = ACC but only LKA
Co-Pilot Assist+ = ACC and LCA
Co-Pilot Assist 2.0 ACC and LCA
Co-Pilot Active = BlueCruise.

Technically, the current requirement is correct.  That hasn't stopped dozens of Maverick and Escape owners with regular Co-Pilot from buying a Comma and then complaining that the Comma website says it should be compatible.  I'm making it explicitly clear that they need to have ACC for the Comma to work, since the Co-Pilot naming convention is not very consistent or clear for Ford.  I'm open to other ways to try and make it clear that not all Co-Pilot versions include ACC.

ugtthis: What if we leave the package as is and add a [footnote](https://github.com/commaai/opendbc/blob/master/opendbc/car/subaru/values.py#L90)? This seems like a similar issue with Subaru except more geared towards Non-US models EyeSight package not including lane keep assist <br>

incognitojam: @alan-polk `Co-Pilot360 Assist = ACC but only LKA` - I think this isn't true for the Maverick 2023+, they use that name for the package with ACC and LCA. I think that's the only model with this difference though.

Adding a footnote to the Maverick 2023+ seems like a good idea.

incognitojam:
There is also an unfortunate line-break in the vehicle docs which separates `Co-Pilot360` from the rest of the package name, which might be causing people to miss it. This will have to be fixed in the website repo.
Screenshot 2025-07-31 at 18 22 09

opendbc/car/ford/values.py
alan-polk: Since the Co-Pilot levels are not sequential or numeric (How do you know that Active is better than Assist+?), I think we should list all known working trim packages.  I updated the PR and also added a footnote about ACC.  Can you look at the latest code please?

opendbc/car/ford/values.py
```diff
@@ -64,11 +64,16 @@ class Footnote(Enum):
     "North and South America/Southeast Asia.",
     Column.MODEL,
   )
+  GLOBAL = CarFootnote(
+    "OpenPilot on Ford requires adaptive cruise control. Ford frequently renames their Co-Pilot packages, "
+    "but not all packages include adaptive cruise control. Co-Pilot360 Assist+, Co-Pilot360 Assist 2.0,"
+    "and Co-Pilot360 Active are known to include adaptive cruise control")
```
ugtthis: CleanShot 2025-07-31 at 14 23 27@2x
^test is failing

<br>

refer to the footnote example above the `GLOBAL` footnote for example

alan-polk: I see, I had accidently deleted the column.package argument.  I think I have fixed it.

sshane:
Make it a nbsp?

ugtthis:
Doing nbsp seems like a temp fix that may have odd responsive text layout issues depending on the length of the package, and still doesn't help with clarity as much

opendbc/car/ford/values.py
```diff
@@ -64,11 +64,18 @@ class Footnote(Enum):
...
+    "and Co-Pilot360 Active are known to include adaptive cruise control",
+    Column.PACKAGE,
+  )
+
 
 
 @dataclass
 class FordCarDocs(CarDocs):
-  package: str = "Co-Pilot360 Assist+"
+  package: str = "Co-Pilot360 Assist+, Co-Pilot360 Assist 2.0, Co-Pilot360 Active"
```
sshane: How about `Adaptive Cruise Control (ACC) & Lane Keep Assist` and a footnote? This list is redundant information with the footnote, and confusing imo. Is it and or or?

alan-polk: This brings up another sore spot.  LKA versus LCA.  I've updated the package to be "Adaptive Cruise Control and Lane Centering Assist" and I've added a second note about LKA versus LCA.

opendbc/car/ford/values.py
```diff
@@ -123,41 +134,41 @@ def init(self):
 
 class CAR(Platforms):
   FORD_BRONCO_SPORT_MK1 = FordPlatformConfig(
-    [FordCarDocs("Ford Bronco Sport 2021-24")],
+    [FordCarDocs("Ford Bronco Sport 2021-24", footnotes=[Footnote.HW_TYPE])],
```
incognitojam: If you want a footnote to show up on all cars you should be able to set a default for `footnotes` in FordCarDocs instead

incognitojam (changes requested):

opendbc/car/ford/values.py
```diff
@@ -64,11 +64,22 @@ class Footnote(Enum):
     "North and South America/Southeast Asia.",
     Column.MODEL,
   )
-
+  GLOBAL = CarFootnote(
+    "OpenPilot on Ford requires adaptive cruise control (ACC) and Lane Centering Assist (LCA). Ford's Co-Pilot naming scheme is complicated"
```
ugtthis: > "OpenPilot on Ford requires adaptive cruise control (ACC) and Lane Centering Assist (LCA). Ford's Co-Pilot naming scheme is complicated" "and evolving with new package names,"

The words above dont provide much value. I think stating the packages that are known to have ACC & LCA is a lot more clearer, since the `package: str`(line 82) just says ACC and LCA. Now this footnote(minus the fluff) can provide the bare minimum helpful context to make it clear what is known to be compatible. I think each word should fight to be there or we end up using footnotes as a crutch.

I also have similar opinions for the second footnote(line 74) "Ford offers two different lane assist packages." I would consider fluff since that doesn't clarify compatibility and takes the user longer to read the footnote.

btw, nice bringing up the issues ford users were having in that channel, I wasn't really aware of that!

alan-polk: Since Ford is unlikely to introduce new variants of Co-Pilot360 that do not include LCA and ACC (All canFD vehicles have LCA), how about this:

Package = Co-Pilot360 Assist+ or Co-Pilot360 Assist 2.0 or Co-Pilot360 Active
Global Note = Co-Pilot360 and Co-Pilot360 Assist do not include the required features for OpenPilot compatibility.

ugtthis: - I like how the package is more specific so it's easier to criticize if we are wrong, so then there is a faster feedback loop versus finding out later that its not clear. Then the note emphasizes what doesn't work.
- I think the note could be shorter -> `Co-Pilot360 and Co-Pilot360 Assist are NOT compatible`

<br>

---

I would imagine a seperate PR would be needed to then adjust the usage of the below packages:
- Adaptive Cruise Control with Lane Centering
- LARIAT Luxury
- Essentially any package that doesn't say "all"

<br>

Question:
Do you know if `Co-Pilot360 Plus` is referring to `Co-Pilot360 Assist+`? 
If so then this would fall under the package update PR too
https://github.com/commaai/opendbc/blob/c6f01a6039faeb0d54d3d5a284d7cd2422eac2a8/opendbc/car/ford/values.py#L148

alan-polk: Co-Pilot360 Plus is the Lincoln equivalent of Ford Co-Pilot360 Assist+ (thanks, Ford, this wasn't confusing enough without having Lincoln packages having different names).

opendbc/car/ford/values.py
```diff
@@ -3,7 +3,8 @@
 from dataclasses import dataclass, field, replace
 from enum import Enum, IntFlag
 
-from opendbc.car import AngleSteeringLimits, Bus, CarSpecs, DbcDict, PlatformConfig, Platforms, uds
+from opendbc.car import Bus, CarSpecs, DbcDict, PlatformConfig, Platforms, uds
+from opendbc.car.lateral import AngleSteeringLimits
```
ugtthis: How does this change relate to this PR?

opendbc/car/ford/values.py
ugtthis: Unless I'm missing something it seems like "Adaptive Cruise Control with Lane Centering", and "LARIAT Luxury" as a listed required package wouldn't be needed anymore? Since the default package is now updated to include the specific Co-Pilot360 variations

opendbc/car/ford/values.py
alan-polk: It doesn't, I just reverted to the wrong commit when I rolled back the HW_TYPE change I had proposed.  Fixed.

opendbc/car/ford/values.py
```diff
@@ -313,4 +318,4 @@ def ford_asbuilt_block_response(block_id: int):
   match_fw_to_car_fuzzy=match_fw_to_car_fuzzy,
 )
 
-DBC = CAR.create_dbc_map()
+DBC = CAR.create_dbc_map()
```
ugtthis: This seems to be another one? maybe white space? or maybe it was because of the revert

alan-polk: This one confuses me, I don't understand what changed.  Both lines are identical.

opendbc/car/ford/values.py
ugtthis: @alan-polk Subaru has an example https://github.com/commaai/opendbc/blob/master/opendbc/car/subaru/values.py

opendbc/car/ford/values.py
ugtthis: It doesn't show as clear online but it's because your missing a new line at the end. Add an extra line and this diff should go away

CleanShot 2025-08-03 at 19 04 58@2x

ugtthis:
I'd mark this PR as a draft until its fully ready for shane/incognitojam to review

alan-polk:
> I'd mark this PR as a draft until its fully ready for shane/incognitojam to review

marked as draft and awaiting feedback from Cameron.

ugtthis:
There is still a requested change above from incognitojam from a couple days ago in regards to a default footnote

alan-polk:
> There is still a requested change above from incognitojam from a couple days ago in regards to a default footnote

I messaged him on Discord to make sure I understand what he wants, and reduce the number of commits.  Waiting on his response.

ugtthis:
In the review above he was pointing out that you could set a default footnote. Currently no models will be getting a footnote with this current code

sshane:
Can we just come up with the simplest thing to merge sooner rather than later?

opendbc/car/ford/values.py
```diff
@@ -64,13 +64,18 @@ class Footnote(Enum):
     "North and South America/Southeast Asia.",
     Column.MODEL,
   )
+  GLOBAL = CarFootnote(
+    "Co-Pilot360 and Co-Pilot360 Assist are NOT COMPATIBLE with OpenPilot",
```
sshane: openpilot is never stylized with capitals

opendbc/car/ford/values.py
```diff
@@ -65,7 +65,7 @@ class Footnote(Enum):
     Column.MODEL,
   )
   GLOBAL = CarFootnote(
-    "Co-Pilot360 and Co-Pilot360 Assist are NOT COMPATIBLE with OpenPilot",
+    "Co-Pilot360 and Co-Pilot360 Assist are not compatible with OpenPilot",
```
sshane: I meant openpilot lol

sshane:
This is all too confusing. Can you just merge both changes into one PR to review?

alan-polk:
> This is all too confusing. Can you just merge both changes into one PR to review?

I'll update this PR this weekend to includes all of ugtthis ford docs changes so it's all in one place.

PR #2594, opened 2025-07-30
refactor: move lateral methods from init to lateral.py

devtekve:
Addressing https://github.com/commaai/opendbc/pull/2503#discussion_r2238469479 

The goal is to simplify `__init__.py` 

> [!NOTE]
> Causes:
> * https://github.com/commaai/openpilot/pull/35856


opendbc/car/lateral.py
```diff
@@ -0,0 +1,121 @@
...
+
+# TODO: I couldn't find references to the code below in the codebase.
+# def get_friction(lateral_accel_error: float, lateral_accel_deadzone: float, friction_threshold: float,
+#                  torque_params: structs.CarParams.LateralTorqueTuning) -> float:
+#   friction_interp = np.interp(
+#     apply_center_deadzone(lateral_accel_error, lateral_accel_deadzone),
+#     [-friction_threshold, friction_threshold],
+#     [-torque_params.friction, torque_params.friction]
+#   )
+#   return float(friction_interp)
```
devtekve: Is it okay to remove this unused method?

sshane: This is used in openpilot, no

opendbc/car/__init__.py
```diff
@@ -100,117 +92,6 @@ class Bus(StrEnum):
...
-
-
-def apply_center_deadzone(error, deadzone):
-  if (error > - deadzone) and (error < deadzone):
-    error = 0.
-  return error
-
-
-def rate_limit(new_value, last_value, dw_step, up_step):
-  return float(np.clip(new_value, last_value + dw_step, last_value + up_step))
```
sshane: this is not only used in lateral functions, but also longitudinal. it's worth it to spend time reading references to make sure how things work are intuitive after refactors. maybe we also need a `car/controls/__init__.py` with common stuff (move get_friction, PID, etc. there).

or just leave this here and import from here in lateral.py?

opendbc/car/__init__.py
```diff
@@ -100,117 +92,6 @@ class Bus(StrEnum):
...
-def get_friction(lateral_accel_error: float, lateral_accel_deadzone: float, friction_threshold: float,
-                 torque_params: structs.CarParams.LateralTorqueTuning) -> float:
-  friction_interp = np.interp(
-    apply_center_deadzone(lateral_accel_error, lateral_accel_deadzone),
-    [-friction_threshold, friction_threshold],
-    [-torque_params.friction, torque_params.friction]
-  )
-  return float(friction_interp)
-
-
```
sshane, line 213: it's best to make unrelated changes in separate PRs. this should be not changed though

opendbc/car/lateral.py
devtekve: Oh I see. I searched for this but I searched it on SP without realizing it and didn't find references there because this change hasn't been synced! my bad :D

opendbc/car/__init__.py
devtekve: Yeah, I was thinking this made more sense being on `lateral.py` considering it's a function used explicitly for lateral control, but at the same time I was hesitant on whether it was used because I was looking for usages in the wrong place.

Are we okay with it being on `lateral.py` ? I will update the reference in OP too if so

opendbc/car/__init__.py
devtekve: I think the MVP is leaving it here and importing it on `lateral.py`, and a future improvement is having it in `car/controls/__init__.py`

devtekve: We didn't need to import it in `lateral.py` the reason I moved it initially was because I thought it was only used on lateral controls but had a bit of confirmation bias since toyota uses for accel, honda for braking... So leaving it untouched on __init__.py for now

opendbc/car/__init__.py
devtekve: if we are good, there's this then:
* https://github.com/commaai/openpilot/pull/35856

opendbc/car/__init__.py
devtekve: Im having second thoughts on this, I will make the `car/controls/` and move `lateral.py` there and make the `__init__.py` it kinda feels like the right thing to do, but let's see how it looks

devtekve: Alright I've done this, but I didn't really find what else we should put on `car/controls/__init__.py` that's not too disruptive. I found `car/common/pid.py` and I could put it on `car/controls/` but not sure if it's what we want? or putting the contents of `pid.py` inside the `__init__.py` but that would be like going backwards maybe?

PR #2596, opened 2025-07-30
Test: Unwrap Nissan Cruise Check

ssysm:
Believe it or not, unwrap a nested if condition can pass a branch coverage test.


sshane:
is every branch tested?

ssysm:
image
Yep... 100% Branch Coverage

PR #2600, opened 2025-07-31
EV6: raise torque limit

sshane:


adeebshihadeh:
Superseded by https://github.com/commaai/opendbc/pull/3224

PR #2601, opened 2025-07-31
parser: optimize MessageState buffer management

deanlee:
using `deque(maxlen=MESSAGE_HISTORY_SIZE) `instead of manual buffer size checks and popping.

### DIFF
--- a/opendbc/can/parser.py
+++ b/opendbc/can/parser.py
@@ -9,7 +9,6 @@ from opendbc.can.dbc import DBC, Signal
 
 MAX_BAD_COUNTER = 5
 CAN_INVALID_CNT = 5
-MESSAGE_HISTORY_SIZE = 500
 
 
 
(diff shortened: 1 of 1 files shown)


opendbc/can/parser.py
```diff
@@ -82,13 +83,10 @@ def parse(self, nanos: int, dat: bytes) -> bool:
...
 
     self.timestamps.append(nanos)
-    max_buffer = 500
-    while len(self.timestamps) > max_buffer:
-      self.timestamps.popleft()
 
     if self.frequency < 1e-5 and len(self.timestamps) >= 3:
       dt = (self.timestamps[-1] - self.timestamps[0]) * 1e-9
-      if (dt > 1.0 or len(self.timestamps) >= max_buffer) and dt != 0:
+      if (dt > 1.0 or len(self.timestamps) >= MESSAGE_HISTORY_SIZE) and dt != 0:
```
adeebshihadeh:
```suggestion
      if (dt > 1.0 or len(self.timestamps) >= self.timestamps.maxlen) and dt != 0:
```

opendbc/can/parser.py
```diff
@@ -40,7 +41,7 @@ class MessageState:
   timeout_threshold: float = 1e5  # default to 1Hz threshold
   vals: list[float] = field(default_factory=list)
   all_vals: list[list[float]] = field(default_factory=list)
-  timestamps: deque[int] = field(default_factory=deque)
+  timestamps: deque[int] = field(default_factory=lambda: deque(maxlen=MESSAGE_HISTORY_SIZE))
```
adeebshihadeh:
```suggestion
  timestamps: deque[int] = field(default_factory=lambda: deque(maxlen=500))
```

PR #2604, opened 2025-07-31
pre-cache counter and checksum signals

deanlee:
pre-caches counter and checksum signals for each message address, reducing repeated lookups and improving packing performance.


sshane:
Even if we don't run it in CI, it would be useful to have a script to print realistic performance results without a route to post on PRs.

adeebshihadeh:
We do, though it could be better

https://github.com/commaai/opendbc/blob/master/opendbc/can/tests/benchmark.py

PR #2610, opened 2025-08-01
Honda: Accord 2023-2025 (canfd)

mvl-boston:
PR to support 2023-2025 Honda Accord

- Based on [pr 2120](https://github.com/commaai/opendbc/pull/2179) from vanillagorillaa, with additional 2025 fingerprints, gas pedal calibration, and rebaselined to current opendbc
- include safety changes from from shane and vanillagorillaa in [pr 2129](https://github.com/commaai/opendbc/pull/2129)
- Supports OP steering/lateral control at all speeds
- Limited torque as typical for Honda models
- Uses stock Acura long control, works at all speeds, includes stop/go support
- OP long control is in separate https://github.com/commaai/opendbc/pull/2569
-  changes consistent with canfd CRV https://github.com/commaai/opendbc/pull/2475


mvl-boston:
Honda steer is not linear.   Need to determine min-max and breakpoints along curve.

jyoung8607:
I haven't pushed it to the limits yet, but it's looking just fine on the lat accel torque controller so far. Feedforward is a good match for what the car needs, very limited proportional cleanup. It builds up a little integral on one side, but it looks like that's just nulling out some EPS built-in roll comp. It does the same thing on other cars.

image

image

jyoung8607:
This drives pretty good overall. This will be a good openpilot car. Notes from drive testing:

* Acts like it's still wound-up briefly coming out of over-limit turns, but I'm not seeing extra integrator in it, actuator output looks right. Will retest with legacy PIF control. This could also be a deadzone at the top (which happens for other Honda) or an angle-rate limit in the EPS.
* Steer touch threshold is pretty heavy, adjust this as we've done for other Honda.
* Stock engage-at-standstill behavior is inconsistent without a lead car, haven't figured out what the rules are. Sometimes it refuses, sometimes it'll engage but wants a gas touch to leave, resume button doesn't work. With a lead car, seems like we have fully touchless S&G working as expected.
* See if blind spot monitor alerts are available on CAN-FD cars.

jyoung8607:
We're dropping a frame or two of STEERING_CONTROL on disengagements.

image

jyoung8607:
The CAN-FD common DBC's copy of LKAS_HUD has some damage, will sort this out later.

image

mvl-boston:
4096 is pretty low for their smaller Hondas, should be able to max out go much higher in the code.   The high ends tend to be lower slope needing a breakpoint.  You should be able to keep trying higher numbers as hardcoded turns in carcontoller at 5mph cruse in a parking lot until it faults.    I believe Accord has been tested to over 10000.

I've never seen a Bosch C fail resume buttons.   Can you doublecheck that maybe safety or something wasn't broken (I'm happy to look at a log if needed).  Edit: I'm reading this is stock resume, frankly I don't remember the rules, Bosch C was more restrictive than earlier Hondas.   I think you have to press set, not resume.

Most Hondas put BSM on the body bus, which isn't accessible via the camera.

jyoung8607:
> 4096 is pretty low for their smaller Hondas, should be able to max out go much higher in the code. The high ends tend to be lower slope needing a breakpoint. You should be able to keep trying higher numbers as hardcoded turns in carcontoller at 5mph cruse in a parking lot until it faults. I believe Accord has been tested to over 10000.

Do we have any hard data that shows an actual increase in lateral accel from doing that, beyond just not faulting?

I'm going to do some more drive-testing today, but I'm seeing potential indications of a deadzone (or at minimum a decrease in response) even within the existing 4096. Dashcam logs show the stock camera peaking at 2560. Our control is very good under about 1.5 m/s2 and starts getting questionable above that.

Support for a wider non-linear control envelope can be discussed, if it's proven to exist and we have a quality implementation with appropriate safety, but what's going in the initial upstream merge is what we can make predictable and performant on the test car before Wednesday.

> I've never seen a Bosch C fail resume buttons. Can you doublecheck that maybe safety or something wasn't broken (I'm happy to look at a log if needed). Edit: I'm reading this is stock resume, frankly I don't remember the rules, Bosch C was more restrictive than earlier Hondas. I think you have to press set, not resume.

I think it may just be timing, I'm going to play with it more.

> Most Hondas put BSM on the body bus, which isn't accessible via the camera.

Just wanted to have a look and see if the car's internal gateway forwarded a copy in the new CAN-FD design. I'm not seeing it right away.

mvl-boston:
> Do we have any hard data that shows an actual increase in lateral accel from doing that, beyond just not faulting?
> 

I'll make my donuts build (drive in circles) so we can create hard data on all the cars.

> Support for a wider non-linear control envelope can be discussed, if it's proven to exist and we have a quality implementation with appropriate safety, but what's going in the initial upstream merge is what we can make predictable and performant on the test car before Wednesday.
> 
edit - looks like the breakpoint change was left out of the PR, it is only in my combined-honda merge.   Regardless will propose some ideas after donut driver logs.   Agree that 4096 already drives well enough that this isn't needed for day1 PR.

jyoung8607:
Notes to self for today:

- [x] Test drive and data review with lat accel torque controller rescaled to 2560 peak
- [x] Test drive and data review with reduced steering touch threshold
- [x] Re-test engagement at standstill, may be nothing to fix here, but seeking understanding
- [x] Fix LKA_HUD message
- [x] Investigate lack of VIN or any firmware from OBD via comma power
- [ ] Review CAN-FD applicable parts of #2473
- [x] Remove inapplicable TX messages from CAN-FD safety (minor refactor needed)

mvl-boston:
> 
>     * [ ]  Remove inapplicable TX messages from CAN-FD safety (minor refactor needed)

I did most of this in [pr 2569](https://github.com/commaai/opendbc/pull/2569) if you want to grab the non-long portion.

jyoung8607:
- [x] Take Control Immediately / Cruise is Off on regen paddle

mvl-boston:
Known bug with regen paddles.   There is an abandoned PR for it, can reopen
tonight.  Also have other regen testers who can help after car return.

Stock regen kills ACC like brake pedal, so openpilot ahould do same.

opendbc/car/honda/interface.py
```diff
@@ -69,6 +69,9 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
     if 0x1C2 in fingerprint[CAN.pt]:
       ret.flags |= HondaFlags.HAS_EPB.value
 
+    if candidate in HONDA_BOSCH_CANFD and 0x184 not in fingerprint[CAN.pt]:
+      ret.flags |= HondaFlags.BOSCH_ALT_BRAKE.value
+
```
jyoung8607: @mvl-boston do you know what's going on here? 0x184 exists on the bus, but is undefined in the DBC, it's not the alt brake message 0x1BE. If there are CAN-FD cars that need it, I would just add `*HONDA_BOSCH_CANFD` to the 0x1BE existence check, but I don't know if there was a good reason behind this 0x184 detection.

https://github.com/commaai/opendbc/blob/28a16218639ff1a5a2bd3e36e750a52bc7f968b1/opendbc/car/honda/interface.py#L209-L212

jyoung8607:
> * [ ]  Investigate lack of VIN or any firmware from OBD via comma power

Having an odd issue wherein a manual query_fw_versions is able to see the usual stuff at the usual addresses responding to 0xF181 queries, but fingerprinting at startup only picks up the radar and camera. Not sure what's going on here, will have to continue debugging tomorrow. I extended the car out to Thursday noon.

`f39cf149898833ff/00000040--b40a6ac230`

mvl-boston:
> > * [ ]   Investigate lack of VIN or any firmware from OBD via comma power
> 
> Having an odd issue wherein a manual query_fw_versions is able to see the usual stuff at the usual addresses responding to 0xF181 queries, but fingerprinting at startup only picks up the radar and camera. Not sure what's going on here, will have to continue debugging tomorrow. I extended the car out to Thursday noon.
> 
> `f39cf149898833ff/00000040--b40a6ac230`

I have seen over the past few months that there are waves where lots of fingerprints come back (even without comma power) for a month, and then a few fingerprints come back in a different month.

I'll look through my logs and find a breakpoint, so that you have a range of diffs to inspect.   My presumption is that it is something lower level, and I didn't have time to investigate.

opendbc/car/honda/interface.py
mvl-boston: I see you resolved this.   Let me know if you still need me to look at it.   I built the alt-brake detection a few weeks back to align with all the other Honda autodetection, half the canfd used one brake signal and half used another, and it was particularly important in long control.  Don't remember specific signals but if you still need me to dig into it I can.

jyoung8607:
Moved the FW/VIN query issue to #2632, don't love it but it's not a blocker.

There's a question around UI/HUD stuff I'll follow up on later, everything else LGTM.

PR #2611, opened 2025-08-02
Toyota: fake lkas buttonEvents

Quantizr:
Create fake lkas buttonEvents by using the car's built in HUD LKAS_STATUS message

TSS-P has an actual button press/unpress signal on Bus 1 0x160, but that message does not exist on TSS2. This effectively does the same thing and works on all toyotas.

### DIFF
--- a/opendbc/car/toyota/carstate.py
+++ b/opendbc/car/toyota/carstate.py
@@ -46,6 +46,8 @@ class CarState(CarStateBase):
 
     self.pcm_follow_distance = 0
 
+    self.lkas_status = None
+
     self.acc_type = 1
     self.lkas_hud = {}
     self.gvc = 0.0
@@ -178,12 +180,24 @@ class CarState(CarStateBase):
     if self.CP.carFingerprint not in UNSUPPORTED_DSU_CAR:
       self.pcm_follow_distance = cp.vl["PCM_CRUISE_2"]["PCM_FOLLOW_DISTANCE"]
 
+
+    button_events = []
+
     if self.CP.carFingerprint in (TSS2_CAR - RADAR_ACC_CAR):
       # distance button is wired to the ACC module (camera or radar)
       prev_distance_button = self.distance_button
       self.distance_button = cp_acc.vl["ACC_CONTROL"]["DISTANCE"]
 
-      ret.buttonEvents = create_button_events(self.distance_button, prev_distance_button, {1: ButtonType.gapAdjustCruise})
+      button_events.extend(create_button_events(self.distance_button, prev_distance_button, {1: ButtonType.gapAdjustCruise}))
+
+    # fake pressed/unpressed lkas button events
+    prev_lkas_status = self.lkas_status
+    self.lkas_status = cp_cam.vl["LKAS_HUD"]["LKAS_STATUS"]
+    if self.lkas_status != prev_lkas_status and prev_lkas_status is not None:
+      button_events.extend(create_button_events(1, 0, {1: ButtonType.lkas}))
+      button_events.extend(create_button_events(0, 1, {1: ButtonType.lkas}))
+
+    ret.buttonEvents = button_events
 
     return ret
 


AlexandreSato:
In TSS2 vehicles, you need to make sure the steering assists are turned off `ret.invalidLkasSetting = cp_cam.vl["LKAS_HUD"]["LDA_SA_TOGGLE"] == 1`, because when the icon pulses, it can activate without any button being pressed.

history:
https://github.com/AlexandreSato/opendbc/commit/79e45c1c58cddf4f36e6c5557d6500a51f549498
https://github.com/AlexandreSato/opendbc/commit/984b1f9490b1a7925b1ee8eeff5c5909e3ef40f2
https://github.com/AlexandreSato/opendbc/commit/d6c0e8ff0641522c41e4a11892526f760112940b
and
https://github.com/AlexandreSato/opendbc/commit/4100159c76688f3908e0bfb9ec164e837f5a16c6

Quantizr:
> In TSS2 vehicles, you need to make sure the steering assists are turned off `ret.invalidLkasSetting = cp_cam.vl["LKAS_HUD"]["LDA_SA_TOGGLE"] == 1`, because when the icon pulses, it can activate without any button being pressed.
> 
> history: [AlexandreSato@79e45c1](https://github.com/AlexandreSato/opendbc/commit/79e45c1c58cddf4f36e6c5557d6500a51f549498) [AlexandreSato@984b1f9](https://github.com/AlexandreSato/opendbc/commit/984b1f9490b1a7925b1ee8eeff5c5909e3ef40f2) [AlexandreSato@d6c0e8f](https://github.com/AlexandreSato/opendbc/commit/d6c0e8ff0641522c41e4a11892526f760112940b) and [AlexandreSato@4100159](https://github.com/AlexandreSato/opendbc/commit/4100159c76688f3908e0bfb9ec164e837f5a16c6)

so when LDA_SA_TOGGLE is high, LKAS_STATUS randomly turns on/off?

seems like checking if LDA_ON_MESSAGE goes from nonzero to zero and vice versa instead of looking at LKAS_STATUS would avoid this?

AlexandreSato:
In the corollas of Brazil yes, it is the icon of the panel. It is possible to reproduce by activating the assists and driftout lanes, at least in TSS2. Not sure about LDA_ON_MESSAGE

opendbc/car/toyota/carstate.py
```diff
@@ -172,18 +174,30 @@ def update(self, can_parsers) -> structs.CarState:
...
+    button_events = []
+
     if self.CP.carFingerprint != CAR.TOYOTA_PRIUS_V:
       self.lkas_hud = copy.copy(cp_cam.vl["LKAS_HUD"])
 
-    if self.CP.carFingerprint not in UNSUPPORTED_DSU_CAR:
-      self.pcm_follow_distance = cp.vl["PCM_CRUISE_2"]["PCM_FOLLOW_DISTANCE"]
+      # fake pressed/unpressed lkas button events
+      prev_lkas_status = self.lkas_status
+      self.lkas_status = cp_cam.vl["LKAS_HUD"]["LKAS_STATUS"]
```
sshane, line 187: This doesn't work on my camry, it's always true.

sshane: Oh it works if I hold my LKAS button for 3 seconds

sshane:
On TSS2 at least, LKAS_STATUS is for the LKAS alerts, not actuation. When you press the "LKAS" button it toggles lane departure assist features, not alerts (between centering and lkas-only). TSS-P is probably different from this, which is why this PR doesn't work on TSS2 if you only tested TSS-P. I would add an if statement for not TSS2 in this PR to make it work properly

Done TSS2 here: https://github.com/commaai/opendbc/pull/2634

sshane:
@Quantizr can you check if my PR works on TSS-P? If not, let's merge this when you can

Quantizr:
> @Quantizr can you check if my PR works on TSS-P? If not, let's merge this when you can

seems like its the opposite in TSS-P

LDA_ON_MESSAGE goes high on button press but goes to 0 after 5 seconds of not touching anything (if below certain speed) or upon a second press within 5 seconds

LKAS_STATUS seems to do what we want it to for TSS-P but I have no clue if the whole turns itself off while driving thing happens in TSS-P.

PR #2618, opened 2025-08-04
PSA: allow unpadded tester_present responses for PSA AEE2010_R3 ECUs

elkoled:
PSA AEE2010_R3 ECUs send 3-byte (unpadded) responses to tester_present requests, but openpilot expects 8 bytes. This causes ECU discovery to fail and blocks firmware queries.

Fix: Accept variable-length tester_present responses instead of requiring 8 bytes.

Logs of official PSA dongle:
Request:
image
Response:
image

### DIFF
--- a/opendbc/car/ecu_addrs.py
+++ b/opendbc/car/ecu_addrs.py
@@ -7,9 +7,11 @@ from opendbc.car.fw_query_definitions import EcuAddrBusType
 
 
 def _is_tester_present_response(msg: CanData, subaddr: int = None) -> bool:
+  # ISO-TP messages may use CAN frame optimization (not always 8 bytes)
   # tester present response is always a single frame
   dat_offset = 1 if subaddr is not None else 0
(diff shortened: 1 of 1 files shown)


sshane:
Do we need to revert https://github.com/commaai/panda/pull/1312 too?

elkoled:
I did not modify panda apart from ignition CAN, FW query works with this opendbc change.

sshane:
Can you provide a route with attempted fingerprinting?

sshane:
I see I already reverted it: https://github.com/commaai/panda/pull/1411

This PR makes sense then. Would like to see a response on CAN though to verify

elkoled:
Here is a route attempting fingerprint without the opendbc fix:
6a7075a4fdd765ee/00000006--f5a03e8cee

sshane:
lgtm! something like this would improve things here https://github.com/commaai/opendbc/issues/2651

PR #2621, opened 2025-08-04
ruff: enable preview rules

sshane:
removes need of reviewing for new line formatting issues. we should fix the indentation issue soon


pyproject.toml
```diff
@@ -107,7 +107,13 @@ ignore = [
   "UP031",
   "NPY002",  # new numpy random syntax is worse
   "C420",
+  "E111",
```
adeebshihadeh: does this actually work for two spaces yet?

sshane: yes! this adds more i need to disable though, fixing

sshane: nope :cry: https://github.com/astral-sh/ruff/issues/8705

PR #2624, opened 2025-08-05
Honda: Disengage stock ACC on regen paddle

jyoung8607:
Act on the regen paddle signals added in #2622. Resolves occurrences of #1100 with stock ACC.

Leaving that bug open, as some safety refactoring is needed to solve it for alpha OP long.

**Before:** `f39cf149898833ff/0000003e--1fd6c210b4`
**After:** `f39cf149898833ff/0000003f--f8fa7f2c17`

Requires commaai/openpilot#35929.

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -131,8 +131,7 @@ class CarState(CarStateBase):
     else:
       gear_position = self.shifter_values.get(cp.vl[self.gearbox_msg]["GEAR_SHIFTER"], None)
       ret.gearShifter = self.parse_gear_shifter(gear_position)
-      if self.CP.transmissionType == TransmissionType.auto:
-        ret.regenBraking = cp.vl["GEARBOX_AUTO"]["REGEN_STAGE_SELECTION"] > 0
+      ret.regenBraking = cp.vl[self.gearbox_msg].get("REGEN_STAGE_SELECTION", 0) > 0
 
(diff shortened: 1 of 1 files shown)


opendbc/car/honda/carstate.py
```diff
@@ -131,6 +131,7 @@ def update(self, can_parsers) -> structs.CarState:
     else:
       gear_position = self.shifter_values.get(cp.vl[self.gearbox_msg]["GEAR_SHIFTER"], None)
       ret.gearShifter = self.parse_gear_shifter(gear_position)
+      ret.regenBraking = cp.vl[self.gearbox_msg].get("REGEN_STAGE_SELECTION", 0) > 0
```
sshane, line 134: What is this? Can we be explicit? Now you can silently skip checking regen paddle if someone renames it in DBC

jyoung8607: Your question is reasonable. I'm trying to resolve #1100 incrementally. It was a blocker for the CAN-FD Accord.

Shortly we'll catch it by shipping #2569 with additions to read the regen paddle state, along the way resolving #1100, and then reverting commaai/openpilot#35929 (nice test BTW).

If I can't do that by tomorrow, I'll make this a normal read gated by `transmissionType == automatic and not CAR.ACURA_RDX:`, that's basically what the above code does.

sshane: I'd prefer we do the correct change now rather than lumping it in a larger PR

PR #2628, opened 2025-08-06
Tesla: Brake to ESP_status

lukasloetkolben:
The new Tesla Model Y Juniper does not include the `IBST_status` message.
`ESP_status` is the same across all vehicles. 
This is likely due to the [new brake system](https://www.teslaoracle.com/2025/03/27/tesla-re-engineers-the-brake-system-in-the-new-model-y-juniper-for-autopilot-and-human-drivers/) implemented in the Juniper.

Validation
* Route: da74b4dc7cea694c/00000001--cdff470bcf


opendbc/safety/modes/tesla.h
```diff
@@ -128,8 +128,8 @@ static void tesla_rx_hook(const CANPacket_t *msg) {
     }
 
     // Brake pressed
-    if (msg->addr == 0x39dU) {
-      brake_pressed = (msg->data[2] & 0x03U) == 2U;
+    if (msg->addr == 0x145U) {
+      brake_pressed = ((msg->data[3] & 0x60U) >> 5) == 2U;
```
sshane:
```suggestion
      brake_pressed = ((msg->data[3] >> 5) & 0x03U) == 2U;
```

reads better to me

lukasloetkolben: I changed it in both places.
Is it good enough to merge?

sshane:
Can you write a quick check with test_models to ensure we don't miss any brake presses we used to catch? Can also use https://huggingface.co/datasets/commaai/commaCarSegments

sshane:
Mutation test is also failing

lukasloetkolben:
> Mutation test is also failing

Why did the mutation test pass with address 0x39dU but fail with 0x145U when the logical structure of the condition is identical?

adeebshihadeh:
> Why did the mutation test pass with address 0x39dU but fail with 0x145U when the logical structure of the condition is identical?

The mutation tests are so slow that we can only run on the `git diff` with respect to master, so this was likely always failing but will need to be fixed now that we have the mutation tests and it's been changed.

YJCITA:
image
I tryed this msg on my car, set breakPressed = BrakeValue > 3, is run OK.
But I donnot how to push code and data

> > Mutation test is also failing
> 
> Why did the mutation test pass with address 0x39dU but fail with 0x145U when the logical structure of the condition is identical?

AmyJeanes:
@sshane Are we able to get any movement on this? Lots of people getting stuck and having to use forks right now for Model Y Juniper. I've manually pulled this change into my own fork running on a 2019 Model 3 and it works exactly the same as before so backwards compat is good also.

sshane:
Checked for instances where the new brake pressed signal significantly lags behind the existing one >50ms, only found one case which is a false positive because of the noisy old signal here:

image

image

---

The rest of the checked cases (>50 mismatches (10 ms frames) or >10 mismatches per braking rising edge) look expected as well, falling edge is slightly later:

image

image

image

image

image

---

```
Checked routes: 65388, dongles: 106 platforms: {'TESLA_MODEL_3': 43907, 'TESLA_MODEL_Y': 21481}
```

sshane:
Thanks!

sshane:
The late falling and rising makes sense if we assume the `IBST_status`'s signals are a source of the ESP's message, specifically the mm distance brake rod distance signal `IBST_sInputRodDriver`. It looks like the IBST produces a very unfiltered higher threshold boolean we used to use, and the ESP uses a lower threshold near 0 and could also temporally filter it as well to produce a clean signal for TCS-related events.

So the IBST message is still likely readable by the ESP ECU, either through another bus or some other form of communication.

image


Both fall at the same time since the signal didn't go through the area where the thresholds would be different for long enough:

image

PR #2633, opened 2025-08-06
Revert "Refactor steering angle limit functions and add common application logic"

sshane:
Reverts commaai/opendbc#2503

### DIFF
--- a/opendbc/car/lateral.py
+++ b/opendbc/car/lateral.py
@@ -1,17 +1,8 @@
-import math
 import numpy as np
 from dataclasses import dataclass
-from opendbc.car import structs, rate_limit, DT_CTRL
-from opendbc.car.vehicle_model import VehicleModel
+from opendbc.car import structs
 
 FRICTION_THRESHOLD = 0.3
-
-# Add extra tolerance for average banked road since safety doesn't have the roll
-AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
-
-# ISO 11270
-ISO_LATERAL_ACCEL = 3.0  # m/s^2
-ISO_LATERAL_JERK = 5.0  # m/s^3
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


opendbc/car/lateral.py
```diff
@@ -128,42 +166,3 @@ def get_friction(lateral_accel_error: float, lateral_accel_deadzone: float, fric
...
-
-
-def get_max_angle_delta_vm(v_ego_raw: float, steer_step: int, VM: VehicleModel, max_lateral_jerk: float):
-  """ Calculate the maximum steering angle rate based on lateral jerk limits. """
-  max_curvature_rate_sec = max_lateral_jerk / (v_ego_raw ** 2)  # (1/m)/s
-  max_angle_rate_sec = math.degrees(VM.get_steer_from_curvature(max_curvature_rate_sec, v_ego_raw, 0))  # deg/s
-  return max_angle_rate_sec * (DT_CTRL * steer_step)
-
-
-def get_max_angle_vm(v_ego_raw: float, VM: VehicleModel, max_lateral_accel: float):
```
devtekve, line 140: Why remove the `max_lateral_accel` on this method? I had it because for the tests it's much easier when testing the safety limits to be able to provide the value aligned to the safety limits, even if the car itself is being driver with a far more conservative values on the `AngleSteeringLimits`defined. Otherwise, safety tests will fail as they can't properly test the boundaries (upper limit)

sshane: How are the tests passing now then?

devtekve: Because for tesla you have the safety.h code aligned with exactly the max jerk and accel. But for example for the angle steering port I’m working on , adeeb asked me to target below the limits to be a bit conservative, so the python code aims for lower than the max set in lateral.h 

when checking if going above the limit as you currently changed it will always pass because panda safety does indeed allow up to the actual real max accel and jerk

devtekve: You can see what I mean on this PR https://github.com/commaai/opendbc/pull/2421/files#diff-5218ada48260f71887a5938dc6fb7c6deb7acc196ded69a9cf1aba2b4a2facfb

PR #2331, opened 2025-05-30
Honda - Update CANFD Pilot

vanillagorillaa:
Re open after messed up rebase on last PR. https://github.com/commaai/opendbc/pull/2141

- New Test route
- Support for 2025 Pilot - e4d479cf4b22ba74/00000004--a3b44c6296
- LKAS_HUD (0x33D) on CANFD Cars is 8 bytes - Same as radarless cars
- Adds RX checks for CANFD

This PR is required before other CAN FD Honda PR's can/should be merged

Issue #2632, opened 2025-08-06
Honda: CAN-FD cars don't return FW or VIN over OBD


jyoung8607:
Bosch CAN-FD cars, e.g. #2610, are only getting back firmware for the camera and radar. The only successful queries are directly on the bus where those controllers live.

It appears that queries via comma power/OBD are dropped by the car's gateway before they get to the controller; we don't see the OBD queries for the camera and radar appear on that bus. It doesn't appear to be timing; it still happens with late startup. I also tried prefixing tester-present or a diagnostic session change without success.

Strangely, these queries work fine with query_fw_versions from the Panda repo. There may be some stateful filtering happening that could be getting swamped by openpilot's parallel query mechanism.

**Route:** `f39cf149898833ff/00000046--3298c279ff`

Manual query results: [query-fw-debug-output.txt](https://github.com/user-attachments/files/21634153/query-fw-debug-output.txt).

mvl-boston:
I could not find any routelogs where my Honda's returned the larger number of fingerprint ECU results.  I suspect it was early 2025 before I bought comma prime so deleted off.   I didn't use comma power but got way more results in the past.

I suggest trying your Accord on the 0.9.7 release3 diff, I recall it got more firmwares back then.   And then you can see which release dropped them to help narrow down root cause.

(I have the RDX only tonight so will try to get lockout testing done there today), but can similarly run this on my MDX tomorrow to see where all the results dropped.

jyoung8607:
- [x] Fix root cause: commaai/panda#2256
- [ ] Track dongles used to add UDS firmware for new Honda CAN-FD cars, refresh firmware after this is fixed
- [x] Re-check whether the EPS/VSA firmware exception is needed, think yes but make sure

jyoung8607:
> * [x]  Fix root cause: commaai/panda#2256

Fix for root cause is merged, however...

> * [ ]  Track dongles used to add UDS firmware for new Honda CAN-FD cars, refresh firmware after this is fixed

... leaving this open for now, because we'll need the fix to percolate out to end users and collect a batch of firmware before adding the new ECUs to fingerprints. Added to 0.10.2 milestone so that we'll have firmware gathered from 0.10.1-release.

jyoung8607:
> * [ ]  Track dongles used to add UDS firmware for new Honda CAN-FD cars, refresh firmware after this is fixed

I don't think we really need this. Can always circle back and add a batch later if desired.

PR #2636, opened 2025-08-07
Honda: Pilot 2023-2025 (canfd)

mvl-boston:
PR to support 2023-2025 Honda Pilot

- based on closed [PR2331](https://github.com/commaai/opendbc/pull/2331)
- changes aligned with [pr2610](https://github.com/commaai/opendbc/pull/2610)
- Parameters set via 75 minute log below.

Validation
* Dongle ID: ea1fc707f10a580a
* Route: 00000002--5d8a903b09

Dashcam route for steer calibraiton: 9662a1dab70c9dc3/000000af--9a141c1015
However tester said factory LKAS is steering way worse than my community build.   Possible the EPS allows more max than factory LKAS is willing to command.


mvl-boston:
Dashcam mode for Pilot is below for review.
9662a1dab70c9dc3/000000af--9a141c1015

mvl-boston:
However tester said factory LKAS is steering way worse than my community build. Possible the EPS allows more max than factory LKAS is willing to command.

adeebshihadeh:
> However tester said factory LKAS is steering way worse than my community build. Possible the EPS allows more max than factory LKAS is willing to command.

We generally don't want to do anything outside what the stock system does, both in terms of using signals they don't use and their ranges. For example, some TSS2 Toyota still allow the torque steering command found on older models, however no stock system uses it, so we also don't use it.

mvl-boston:
Received approval to publicly post the test log.   Revised in routes.py

jyoung8607:
> Received approval to publicly post the test log. Revised in routes.py

CI tests are choking on that route, and it's not clear why. Looks like we might be trying to send a message that upstream Panda safety doesn't expect. It's a route from a couple months ago on one of your forks, not sure what was in it. Would be nice to swap this out with a route from more modern openpilot, but not a blocker for merge.

jyoung8607:
LGTM. Notes for merge:

* There are confirmed examples of MY2023, MY2024 and MY2025
* The fingerprint list was recreated from five known dongles, notes in commit history/git blame
  * #2632

mvl-boston:
> > Received approval to publicly post the test log. Revised in routes.py
> 
> CI tests are choking on that route, and it's not clear why. Looks like we might be trying to send a message that upstream Panda safety doesn't expect. It's a route from a couple months ago on one of your forks, not sure what was in it. Would be nice to swap this out with a route from more modern openpilot, but not a blocker for merge.

Where can I see the choke to research?

mvl-boston:
> LGTM. Notes for merge:
> 
> * There are confirmed examples of MY2023, MY2024 and MY2025
> * The fingerprint list was recreated from five known dongles, notes in commit history/git blame
> * Startup values are near live-learned values for liveParameters, liveDelay, and liveTorqueParameters
> * Same caveats as the Accord 11G
>   * We can't control the lane-lines on the stock instrument cluster
>   * #2632

Thanks!

jyoung8607:
> Where can I see the choke to research?

If you click on the red X next to your commit that changed it, you should get [here](https://github.com/commaai/opendbc/actions/runs/16819293127/job/47642927811).

opendbc/car/honda/values.py
```diff
@@ -330,6 +329,7 @@ class CAR(Platforms):
   CAR.ACURA_RDX: 400,
   CAR.HONDA_CRV_EU: 400,
   CAR.HONDA_ACCORD_11G: 600,
+  CAR.HONDA_PILOT_4G: 600,
```
sshane, line 332: how did you find this? Seems like it activates commonly when disengaged

mvl-boston: This specific PR one preceeds me, but for the other cars the general approach was to find the value when the car is at the max steer command and set it just a bit higher to indicate the driver was overriding.

PR #2638, opened 2025-08-07
Honda: Safety support for driver regen paddle

### LINKED ISSUE #1100: Honda: check regen paddle

jyoung8607:
Resolves #1100.

Feedback *extremely* welcome on better ways to deal with the RX checks, within the confines of MISRA.

- [ ] Check commaCarSegments for potential false positives
- [ ] Revert commaai/openpilot#35929

### DIFF
--- a/opendbc/safety/modes/honda.h
+++ b/opendbc/safety/modes/honda.h
@@ -350,47 +350,6 @@ static safety_config honda_bosch_init(uint16_t param) {
   const uint16_t HONDA_PARAM_BOSCH_CANFD = 16;
   const uint16_t HONDA_PARAM_REGEN = 32;
 
-  // Bosch radarless has the powertrain bus on bus 0
-  static RxCheck honda_bosch_pt0_rx_checks[] = {
-    HONDA_COMMON_RX_CHECKS(0)
-  };
-
-  static RxCheck honda_bosch_pt0_regen_rx_checks[] = {
-    HONDA_COMMON_RX_CHECKS(0)
-    HONDA_REGEN_ADDR_CHECK(0)
-  };
-
-  static RxCheck honda_bosch_pt0_alt_brake_rx_checks[] = {
-    HONDA_COMMON_RX_CHECKS(0)
-    HONDA_ALT_BRAKE_ADDR_CHECK(0)
-  };
-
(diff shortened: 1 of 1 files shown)


opendbc/car/honda/values.py
```diff
@@ -55,6 +55,7 @@ class HondaSafetyFlags(IntFlag):
   NIDEC_ALT = 4
   RADARLESS = 8
   BOSCH_CANFD = 16
+  REGEN = 32
```
sshane, line 58:
```suggestion
  REGEN_PADDLE = 32
```

sshane:
I see no mention of regen or paddle in ISO 15622, perhaps it is not a requirement to enforce in our safety model? What do you think @adeebshihadeh? Otherwise, that refactor that was explored at the last hackathon to make the safety more dynamic and reduce added complexity like this would be nice to look into soon.

this refactor if it was not clear:

```c
if (regen_param) {
  add message to checks
}
```

jyoung8607:
Closing this one, it's a mess. The original issue remains open. As Shane mentioned, there has to be a better way.

PR #2647, opened 2025-08-11
HKG: Consolidate cluster lanelines platforms with int flag

sunnyhaibin:
No functional change, just making `hyundaican.py` less bulky. 28 platforms with the new int flag.


opendbc/car/hyundai/values.py
```diff
@@ -126,6 +126,8 @@ class HyundaiFlags(IntFlag):
 
   ALT_LIMITS_2 = 2 ** 26
 
+  HAS_CLUSTER_LANELINES = 2 ** 27
```
adeebshihadeh, line 129: is this something we can detect from the stock camera messages instead of hardcoding?

sunnyhaibin: Most CAN platforms don't have many stock camera messages like the CAN FD variants; I looked at a few platforms and they don't seem to have anything unique to use for detections.

adeebshihadeh: we can't just check the stock values for these?
```
    values["CF_Lkas_LdwsActivemode"] = int(left_lane) + (int(right_lane) << 1)
    values["CF_Lkas_LdwsOpt_USM"] = 2
```

sunnyhaibin: Not really, they change based on lanelines conditions, and I haven't seen any other messages that are specifically just for lanelines.

From a 2022 Hyundai Elantra on a ~10-minute route:
image

jyoung8607: Can support be inferred from the camera firmware from UDS?

sunnyhaibin: There isn't much of a pattern on just the camera FWs to distinguish them, as some camera FWs are shared across different models that don't share the same configs.

sshane: I also checked to do this last year and wasn't able to detect it easily

jyoung8607: Has anyone run panda/examples/query_fw_versions lately, against the ADAS stuff or the EPS rack? It didn't run brute force searches for all available data in the beginning days of FPv2. There might be actionable information we're just not sending queries for.

sunnyhaibin:
@sshane @jyoung8607 can we reopen this?

sunnyhaibin:
@sshane can we reopen this?

PR #1294, opened 2024-09-27
Hyundai: Refactor interface with int flags

sunnyhaibin:
**Description**

Refactor Hyundai interface to use `HyundaiFlags` int flags for feature detection, replacing predefined feature lists. Feature lists are retained for unit test for the time being.

**Offline Validation**
- [x] `pytest opendbc`
- [x] `pytest selfdrive/car/tests`
- [x] `selfdrive/test/process_replay/test_processes.py`

PR #2648, opened 2025-08-11
Hyundai: cleanup unsupported longitudinal cars list

sunnyhaibin:
- Continued effort from https://github.com/commaai/opendbc/pull/1294
- `CANFD_RADAR_SCC_CAR`:
  - Removed
  - No longer used as these variants now support openpilot longitudinal control
- `CANFD_UNSUPPORTED_LONGITUDINAL_CAR`:
  - Removed
  - Use int flag checks instead during assignments
- `UNSUPPORTED_LONGITUDINAL_CAR`
  - Retained for tests (could be removed, possibly in another PR)
  - Refactored into `dict` for maintainability
  - Updated tests with the new `dict`


sunnyhaibin:
@sshane @jyoung8607 can we reopen this?

PR #2665, opened 2025-08-15
CANParser: determine can_valid on read

adeebshihadeh:
Fixes https://github.com/commaai/openpilot/issues/35996

Before / After
image


sshane:
@YassineYousfi this may fix your case you dealt with in CarState

PR #2670, opened 2025-08-17
 Multiple changes: Updated Honda Values with new flags to support eps_modified and remove eps_modified variable in favor of HondaFlags-based approach, restored removed tuning for HONDA_CIVIC_BOSCH and HONDA_CIVIC_BOSCH_DIESEL and modify torque controller ki to recommended

roenthomas:
Can access the value of ret.flags outside of interface.py

Restored removed torque tuning for two models

As per @twilsonco
"If there's no oscillations, then it's not too high.

For the longest time, ki was kept very low due to oscillations. When the torque controller was introduced, for ki to be 10% of kp was a rather high value."

updated car/honda/values.py and interface.py
updated car/interfaces.py


adeebshihadeh:
See #2672

PR #2672, opened 2025-08-17
Honda: remove EPS mod support

adeebshihadeh:
We removed the mod from our Civic a while ago and are choosing to no longer support this in stock openpilot at this time.

Our [blog post](https://blog.comma.ai/safer-control-of-steering/) on this is still relevant.

### DIFF
--- a/opendbc/car/honda/fingerprints.py
+++ b/opendbc/car/honda/fingerprints.py
@@ -67,7 +67,6 @@ FW_VERSIONS = {
     ],
     (Ecu.eps, 0x18da30f1, None): [
       b'39990-TBX-H120\x00\x00',
-      b'39990-TVA,A150\x00\x00',
       b'39990-TVA-A140\x00\x00',
       b'39990-TVA-A150\x00\x00',
       b'39990-TVA-A160\x00\x00',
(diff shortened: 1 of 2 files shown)


ikichullo:
This really sucks.

dayofdoom:
> Even minor EPS firmware mods mostly retain the safety. If you insist on more torque, a small data only change to a table in the EPS firmware is likely much safer than the other two options.

This is an opaque change and doesn’t match with the blog post ethos.  I propose this be reverted. 

Perhaps there are updates coming that increase safety risk? Otherwise, this is an unpopular pr.

RiskyBiscuit-arc:
This change goes against what Comma was built on, being able to modify and improve what you already own. Everything I've seen the team talk about modifying EPS is safety, but all you’re doing is making it more unsafe as instead of having a strong community support to reach out to, users are resorting to underground chats and forms and are more likely to get false information or mess something up, which is more of a risk. You can't control what your users are going to do; you can only help them. Plus, as with running a torque-modded Honda myself, I feel safer with the mod than without it, as the driving was not "chill" as it couldn't handle even the simplest curves that I would expect an "autopilot”-like driver assist system to handle and would constantly be having to jerk to make corrections instead of slight adjustments occasionally with a torque-modded EPS.

PR #2674, opened 2025-08-18
Toyota Enhancements

Candy0707:
1. update HUD visual Alert
2. Add DriveMode

Validation
* Dongle ID: 3285d981f61036be
* Route: 3285d981f61036be/00000012--7c04ba33ca/0


opendbc/car/toyota/carstate.py
```diff
@@ -80,6 +82,13 @@ def update(self, can_parsers) -> structs.CarState:
       if not self.CP.enableDsu and not self.CP.flags & ToyotaFlags.DISABLE_RADAR.value:
         ret.stockAeb = bool(cp_acc.vl["PRE_COLLISION"]["PRECOLLISION_ACTIVE"] and cp_acc.vl["PRE_COLLISION"]["FORCE"] < -1e-5)
 
+    if cp.vl["GEAR_PACKET"]["SPORT_ON"]:
+      self.drive_Mode = DriveMode.sport
+    elif cp.vl["GEAR_PACKET"]['ECON_ON']:
+      self.drive_Mode = DriveMode.eco
+    else:
+      self.drive_Mode = DriveMode.normal
+
```
sshane: What is this for?

opendbc/car/toyota/toyotacan.py
```diff
@@ -103,18 +103,19 @@ def create_fcw_command(packer, fcw):
...
-def create_ui_command(packer, steer, chime, left_line, right_line, left_lane_depart, right_lane_depart, enabled, stock_lkas_hud):
+def create_ui_command(packer, steer, chime, left_line, right_line, left_lane_depart, right_lane_depart, enabled, stock_lkas_hud, main_enabled):
+  lane_depart = left_lane_depart or right_lane_depart
   values = {
     "TWO_BEEPS": chime,
     "LDA_ALERT": steer,
     "RIGHT_LINE": 3 if right_lane_depart else 1 if right_line else 2,
     "LEFT_LINE": 3 if left_lane_depart else 1 if left_line else 2,
     "BARRIERS": 1 if enabled else 0,
+    "LKAS_STATUS": 3 if lane_depart else 2 if enabled else 1 if main_enabled else 0,
```
sshane: What does this change?

Candy0707: Originally, the instrument would show that LTA was enabled even if it was not enabled. After the improvement, the instrument would show white when LTA was enabled, green when in motion, and orange and flashing when lane deviation occurred.

opendbc/car/toyota/carstate.py
Candy0707: Acceleration curve can be customized according to the driving mode set by the vehicle

opendbc/car/toyota/toyotacan.py
Candy0707: "LKAS_STATUS":0


"LKAS_STATUS": 1


"LKAS_STATUS": 2


"LKAS_STATUS": 3
It will turn orange and flash

sshane: Ah I have a LE Camry so I don't have this expanded display and never noticed. I don't think we will merge sport mode detection just yet, no need at the moment. That's perfectly reasonable for a fork, but not upstream. Can you remove the drive mode code and just leave the LTA change?

PR #2680, opened 2025-08-19
Honda - correcting Civic documentation for MT and Sensing

mvl-boston:
- Update documentation to require automatic transmission when ALLOW_MANUAL_TRANS is not set (eg: FK8 Civic Type R)
- Honda sensing became standard in 2019

### DIFF
--- a/opendbc/car/honda/values.py
+++ b/opendbc/car/honda/values.py
@@ -177,9 +177,10 @@ class CAR(Platforms):
   )
   HONDA_CIVIC_BOSCH = HondaBoschPlatformConfig(
     [
-      HondaCarDocs("Honda Civic 2019-21", "All", video="https://www.youtube.com/watch?v=4Iz1Mz5LGF8",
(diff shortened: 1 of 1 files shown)


jyoung8607 (changes requested):
Splitting up the model-years to document the packaging change is good.

There's no need to restrict documented support by transmission. The allow flag/filter will probably go away shortly after release and allow manual trans on any supported Honda/Acura. Until then, if someone shows up with a MT on a Honda/Acura that doesn't already allow it, I'm happy to add it.

mvl-boston:
> Splitting up the model-years to document the packaging change is good.
> 
> There's no need to restrict documented support by transmission. The allow flag/filter will probably go away shortly after release and allow manual trans on any supported Honda/Acura. Until then, if someone shows up with a MT on a Honda/Acura that doesn't already allow it, I'm happy to add it.

Done.

PR #2681, opened 2025-08-19
Honda: Update package availability for Civic, take 2

jyoung8607:
Triple checked, #2680 is almost correct but not quite.

https://www.hondainfocenter.com/2019/Civic-Sedan/Feature-Guide/Safety-Features/Honda-Sensing/

https://www.hondainfocenter.com/2020/Civic-Sedan/Feature-Guide/Safety-Features/Honda-Sensing/

### DIFF
--- a/opendbc/car/honda/values.py
+++ b/opendbc/car/honda/values.py
@@ -179,8 +179,8 @@ class CAR(Platforms):
     [
       HondaCarDocs("Honda Civic 2019-21", "All", video="https://www.youtube.com/watch?v=4Iz1Mz5LGF8",
                    footnotes=[Footnote.CIVIC_DIESEL], min_steer_speed=2. * CV.MPH_TO_MS),
-      HondaCarDocs("Honda Civic Hatchback 2017-18", min_steer_speed=12. * CV.MPH_TO_MS),
-      HondaCarDocs("Honda Civic Hatchback 2019-21", "All", min_steer_speed=12. * CV.MPH_TO_MS),
+      HondaCarDocs("Honda Civic Hatchback 2017-19", min_steer_speed=12. * CV.MPH_TO_MS),
+      HondaCarDocs("Honda Civic Hatchback 2020-21", "All", min_steer_speed=12. * CV.MPH_TO_MS),
     ],
     CarSpecs(mass=1326, wheelbase=2.7, steerRatio=15.38, centerToFrontRatio=0.4),  # steerRatio: 10.93 is end-to-end spec
     {Bus.pt: 'honda_civic_hatchback_ex_2017_can_generated'},


jyoung8607:
@mvl-boston Can you validate how the different Civic sub-models fit together here? The sedan vs hatch vs coupe language coupled with the Civic Si packaging difference is throwing me. Would it make sense to split Civic Si and/or Civic Type R off into its own CarDocs entry? As an example, we list the different Volkswagen Golf variants like Golf R separately.

mvl-boston:
Civic Type R and Civic Si are the two manual transmission models.   Per your prior comments you were going to keep silent on those and add them in as drivers raise concerns.   I know the 21 and earlier Type R (FK8) has been discussed recently in discord with pre-purchase questions, but no one has post-purchase with fingerprints.

I had included 2019 Hatchback as standard Honda sensing due to below:
https://hondanews.com/en-US/honda-automobiles/releases/release-266177b6f8334d40b52db962dd065fc2-2019-honda-civic-type-r-and-civic-hatchback-accelerate-into-dealerships

mvl-boston:
I researched and the MT models (Type R and SI) got standard Honda Sensing in 2020.   Honda Sensing was not available in 2019 for MT.

So if you list them separately, then they should start with 2020 (after Civic Bosch becomes MT ungated and fingerprinted).

mvl-boston:
Ugh - There is a Civic Hatchback Sport, which is an MT and does get Sensing in 2019.     The SI and Type R do not get Sensing until 2020.

Maybe worth separating SI and Type R after all?

mvl-boston:
Below is a consolidated summary based on the press releases, and determining Nidec vs Bosch from existing OP master code.
CivicModels

There was no Honda Sensing before 2016.

Note that the Hybrid is available for 2025+ in the US (earlier in Europe), but they don't brand it as a separate model, instead it is just a trim level that shares Bosch B.

mvl-boston:
I thought about this some more and would recommend collapsing into one "Civic" line, I think more than that would be confusing to potential buyers.

2016 - add requires Honda Sensing capabilities, point to Nidec featureset
2017-2018 - add requires Honda Sensing, point to Bosh reduced capabilties.  Add a footnote adding nidec capabilities for Coupe or Sedan
2019 - keep requires Honda Sensing note to exclude the SI/Type R, point to Bosch A capabilities, remove nidec language for coupe/sedan
2020 - 2021 - list as "all", point to Bosch A capabilities
22+ - list as "all" point to Bosch B capabilities, include footnote that Hybrid and Manual Transmission models are supported

Based on this, let me know if you prefer I write a PR for it.

sshane:
Why can't we merge this?

sshane:
Looks like it has it: https://www.hondainfocenter.com/-/media/Honda-Sales-Tool-Media-Folder/Images/2019-Civic-Hatch/PDF/2019-Civic-Hatchback-Trimwalk.ashx

PR #2685, opened 2025-08-21
Honda: Basic support for Bosch ALT_RADAR

mvl-boston:
Mitigation for models that lockout lkas after wheel jerks.
- Tested successfully on 5G Odyssey
- Test requested for 2019-2021 3G RDX, and 2022-2024 3G MMR RDX

Does NOT fix EPS shutdown experienced in Bosch A & Nidec Civics (https://github.com/commaai/opendbc/issues/1102) nor in 2025 (3G MMR2) RDX, which seem to have different behavior from above two LKAS lockout cars.  Continued attempts to iterate on an EPS shutdown solution will be addressed in https://github.com/commaai/opendbc/pull/2474 as test cars become available.

Baselined on https://github.com/commaai/opendbc/pull/2560
LKAS UI improvements moved to https://github.com/commaai/opendbc/pull/2473

### DIFF
--- a/opendbc/car/honda/values.py
+++ b/opendbc/car/honda/values.py
@@ -72,10 +72,9 @@ class HondaFlags(IntFlag):
   NIDEC_ALT_SCM_MESSAGES = 64
 
   BOSCH_CANFD = 128
-  BOSCH_ALT_RADAR = 2048
 
   HAS_ALL_DOOR_STATES = 256  # Some Hondas have all door states, others only driver door
-  # 512 is available for future use
+  BOSCH_ALT_RADAR = 512
   ALLOW_MANUAL_TRANS = 1024
   HYBRID = 2048
 


jyoung8607:
Notes for myself when test driving:

- [x] For EPS states 0 through 7, we need better understanding of states 1 through 5
- [ ] Cleanup and merge #2559
- [x] Verify efficacy of the mitigation
- [x] If possible (TBD) make the mitigation silent (no driver-visible fault)

mvl-boston:
Clarifying comments that EPS shutdown is not addressed here.

jyoung8607:
> * [ ]  For EPS states 0 through 7, we need better understanding of states 1 through 5

State 2 appears to be driver intervention, and state 4 appears to be an escalation of that state. All instances of state 4 I've seen thus far seem to be preceded by state 2. The stock camera immediately stops LKA in state 2. I think the EPS will tolerate steering in state 2 but maybe not in state 4, not fully understood yet.

State 3 LOW_SPEED_LOCKOUT only appears at very low speed. I think this is when the actual EPS is refusing to steer due to speed. In normal operation with stock ACC, some ALT_RADAR cars (Odyssey) additionally filter steering controls above a higher min steer speed, and some don't (MDX, TLX). This seems to be market segmentation for Traffic Jam Assist.

> * [ ]  Cleanup and merge [Honda: EPS fault value updates #2559](https://github.com/commaai/opendbc/pull/2559)

Going to ship this under ALT_RADAR only to start, and may expand to other cars later.

PR #2687, opened 2025-08-21
Honda: Interface refactor

jyoung8607:
Cleanup and reorganize. Not sure if final form yet.

EPS-modded Honda were the only users of non-linear legacy PIF control. After #2672 it can be deprecated. For forks that want to support those cars, a custom feedforward function for the lateral accel controller would be a better idea.

- [ ] Diff reduction PR for wheel speed factor
- [x] Diff reduction PR: #2694
- [ ] Mechanical validation of before/after lateral control configurations


jyoung8607:
Other Hondas beyond the new CAN-FD cars are having iffy results with the lateral accel torque controller, and it turns out we may still have applications that benefit from non-linear actuator output.

PR #2466, opened 2025-07-14
Toyota: compensate for changes in pitch

sshane:

PR #2690, opened 2025-08-22
Toyota: feedforward compensate laggy response

sshane:
Ideally will remove weird future aEgo code while being better. Split from https://github.com/commaai/opendbc/pull/2466


sshane:
Could revisit later, but control is now improved after https://github.com/commaai/opendbc/pull/2466

PR #2696, opened 2025-08-24
Honda: set manuals 25mph minimum

mvl-boston:
Honda manuals do not have low-speed-follow.  Thus they cannot do SNG and they disable cruise at 25mph.   Attempting to engage cruise below that speed results in a disruptive alarm from comma.

Lower speeds are available with alpha-long.

This pr makes comma aware of this limit and not attempt to engage cruise below 25mph with alphalong control off.

Test log: f9bcb4642e893c8b/00000001--fbc1c7303f


mvl-boston:
test was successful, ready for review

sshane:
How does openpilot engage if stock long at all? What alarm are you talking about?

mvl-boston:
The test drivers say that there is a cruise disengagement alarm.  Likely the alarm from unexpected disable in cruisestate.   This is very old now, will dig through DMs and see if the alarm route is still available from the test driver.

Edit: don't have alarm testroute.  Asked an MT driver for a testroute.

PR #2699, opened 2025-08-25
Improve debugging for safety

devtekve:
Validation
* Dongle ID: 
* Route:


opendbc/safety/tests/libsafety/SConscript
```diff
@@ -30,6 +30,17 @@ env = Environment(
 if system == "Darwin":
   env.PrependENVPath('PATH', '/opt/homebrew/bin')
 
+try:
+  if GetOption('unoptimized'):
+    flags = [
```
adeebshihadeh: can we just always use these options? does it make the tests meaningfully slower?

opendbc/safety/tests/libsafety/SConscript
```diff
@@ -30,6 +30,17 @@ env = Environment(
 if system == "Darwin":
   env.PrependENVPath('PATH', '/opt/homebrew/bin')
 
+try:
```
adeebshihadeh: if we do need the option, then remove the try/except

adeebshihadeh (changes requested):

opendbc/safety/tests/libsafety/SConscript
devtekve: Let me undo it and we see how it impacts the pipeline. As long as this is only done while on PC (not on tizi) we should be fine. Otherwise we'd appreciate the optimizations :D

opendbc/safety/tests/libsafety/SConscript
devtekve: FYI if we do need the option, this try except should be here until we merge the SConstruct change in OP side, otherwise the getOption fails if the option is not defined there

opendbc/safety/tests/libsafety/SConscript
```diff
@@ -22,6 +22,10 @@ env = Environment(
     '-std=gnu11',
     '-Wfatal-errors',
     '-Wno-pointer-to-int-cast',
+    '-DCANFD',
```
adeebshihadeh: this isn't a flag anymore

devtekve:
```suggestion
```

devtekve: Yeah you were quicker checking than me hahah I was caught up with other stuff. This is now good and didn’t seem to add much to the tests , still below 1m

PR #2700, opened 2025-08-26
Rivian: update mando radar interface

lukasloetkolben:
This corrects the radar interface for the Mando radar used on Rivians.
It’s likely that the same or a very similar interface applies to Hyundai equipped with mando radars as well.


lukasloetkolben:
Do you have time to review this, @sshane?

PR #2708, opened 2025-08-29
Added KIA Xceed phev 2021 (fingerprint and hybrid flag)

froddzei:
I added kia xceed fingerprint to the kia ceed port, since they have the same CD platform. everything works as expected, also added the "hybrid" flag in values.py in order to get it to work.

Used Harness B for my car.

Validation
* Dongle ID:  22703002ddbe2a08
* Route:  22703002ddbe2a08/0000000e--083b76c42c/0

### DIFF
--- a/opendbc/car/hyundai/values.py
+++ b/opendbc/car/hyundai/values.py
@@ -517,7 +517,7 @@ class CAR(Platforms):
   KIA_CEED = HyundaiPlatformConfig(
     [HyundaiCarDocs("Kia Ceed 2019-21", car_parts=CarParts.common([CarHarness.hyundai_e]))],
     CarSpecs(mass=1450, wheelbase=2.65, steerRatio=13.75, tireStiffnessFactor=0.5),
-    flags=HyundaiFlags.LEGACY,
+    flags=HyundaiFlags.LEGACY | HyundaiFlags.HYBRID,
   )
   KIA_EV6 = HyundaiCanFDPlatformConfig(
     [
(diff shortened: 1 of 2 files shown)


opendbc/car/hyundai/values.py
```diff
@@ -506,7 +506,7 @@ class CAR(Platforms):
   KIA_CEED = HyundaiPlatformConfig(
     [HyundaiCarDocs("Kia Ceed 2019-21", car_parts=CarParts.common([CarHarness.hyundai_e]))],
```
jyoung8607, line 507: Xceed will need an additional CarDocs entry at a minimum

opendbc/car/hyundai/values.py
```diff
@@ -506,7 +506,7 @@ class CAR(Platforms):
...
     CarSpecs(mass=1450, wheelbase=2.65, steerRatio=13.75, tireStiffnessFactor=0.5),
-    flags=HyundaiFlags.LEGACY,
+    flags=HyundaiFlags.LEGACY | HyundaiFlags.HYBRID,
```
jyoung8607, line 509: We can't just add the HYBRID flag for all Ceed, that will break the non-hybrid cars. We could potentially add XCeed/Hybrid permutations, but ideally HYBRID should be auto-detected instead.

froddzei:
Hello, do you need anything from me?

opendbc/car/hyundai/values.py
froddzei: i accidently deleted the repository do i need to re-create everything again ?

froddzei:
**Created a new PR :** https://github.com/commaai/opendbc/pull/2845

PR #2717, opened 2025-09-01
Honda: Acura TLX 2021 support

mvl-boston:
Support for 2021 Acura TLX.   Most likely the same logic will support 2022-2023 since they are the same Honda generation.

- Alt_radar car: dynamic units and non-normal steer status at standstill
- Car has traffic jam assist (eg: steers at all speeds, unlike other alt_radar)
- Included torque controller based on first testroute
- dashcam routes, showing standard 2560: 58c4a49c7e193080/0000000e--51d2c245d9

Todo:
- [x] See if torque controller results change after more driving
- [ ] Await testroute from driver with comma power hooked up


opendbc/car/honda/fingerprints.py
```diff
@@ -1034,4 +1034,12 @@
...
     ],
   },
+  CAR.ACURA_TLX_2G: {
+    (Ecu.fwdRadar, 0x18dab0f1, None): [
+      b'36802-TGV-A060\x00\x00',
+    ],
+     (Ecu.fwdCamera, 0x18dab5f1, None): [
+      b'36161-TGV-A030\x00\x00',
+    ],
+  },
```
jyoung8607, line 1061: Try to have someone test with comma power hooked up, even if only temporary.

mvl-boston: requested, awaiting log

mvl-boston:
>     * Does the ALT_RADAR steering fault mitigation work?
Driver has never experienced a lockout so I assume yes.  There were some non-normal statuses at red lights.

>     * Does #2773 look sensible on this car?
This is a low speed follow car so I don't think it is relevant.

jyoung8607:
> This is a low speed follow car so I don't think it is relevant.

We still have different behavior for when dashed/solid lines appear. We use the LKAS_PROBLEM state to indicate steering override, which is new and I just want to make sure it doesn't have dumb results on clusters other than Odyssey.

> * Car has traffic jam assist (eg: steers at all speeds, unlike other alt_radar)

Does it need that STEER_DOWN_TO_ZERO signal set in order to do so?

It's okay either way, I'll be merging support for that separately, just need to know for this PR.

mvl-boston:
For whatever strange reason, the Bosch A TLX supports steer to zero without
the bit set, and the Bosch A MDX needs the bit set.

jyoung8607:
@mvl-boston I got your note re: the broken comma power for this user. Since it happens that we'll be doing a round of firmware additions for all CAN-FD cars anyway, we can potentially handle TLX 2021 at the same time.

That said, I'm only aware of the one owner/one dongle ID, and that dongle hasn't phoned home since September 2, and it wasn't driving on the lateral accel torque controller for any of its prior routes. I already see SR will need to go up as well but it didn't drive long enough to settle. So, we don't have a good tune yet.

mvl-boston:
Will ask the owner to plug into home wifi for updates/phone home.  They switch the 3X between the TLX and a Prius if I remember correctly.

Do learned steer/lataccel values get confused when car-switching?

jyoung8607:
> Do learned steer/lataccel values get confused when car-switching?

They get reset, so learning starts over from the stock configured values.

Mounting calibration is *supposed* to reset but, long story short, may not. Good idea to manually reset it.

mvl-boston:
Lots of logs uploaded, slightly modified numbers to align with latest log:
58c4a49c7e193080/00000001--3e4a81ed74
58c4a49c7e193080/00000002--885cd4f5c3
58c4a49c7e193080/00000003--2416e3c2f7
58c4a49c7e193080/00000004--8516cae6bb
58c4a49c7e193080/00000006--d657cc984e
58c4a49c7e193080/00000005--e8cb6206a5
58c4a49c7e193080/00000008--aed64b5ba8

Per discussions, comma power was deferred.  Let me know if anything further needed to upstream.

jyoung8607:
Looks like SR and lat accel factors did converge near where they're currently set, so LGTM.

PR #2723, opened 2025-09-02
VW MQB: Add missing AWV (ACC_10) & LH_EPS_03 Signals

tealtwo:
This PR adds missing AWV signals for VW MQB in the ACC_10 message, these signals are used for the `AWA` function which is apart of the `AWV` function, and provides Evasive Steering Assistance as apart of the Front Assist package by using the signals added in this PR to command steering to the EPS, it is important to note this is only a DBC edit, as for the EPS to accept these signals it needs to be parameterized by ODIS to support `AWA`. 

`AWA` = `Ausweichassistent` = `Evasive Steering Assistant`


tealtwo:
ready to merge

PR #2727, opened 2025-09-03
VW PQ: add missing LDW_Status signal

tealtwo:


opendbc/car/volkswagen/pqcan.py
```diff
@@ -21,6 +21,7 @@ def create_lka_hud_control(packer, bus, ldw_stock_values, lat_active, steering_p
     ]}
 
   values.update({
+    "LDW_Kameratyp": 1,
```
jyoung8607, line 24: What does this accomplish? I'm open to merging if this brings us closer to stock behavior, but I don't remember this being an issue in the past, so I'd like to know why we're changing it now.

tealtwo: > What does this accomplish? I'm open to merging if this brings us closer to stock behavior, but I don't remember this being an issue in the past, so I'd like to know why we're changing it now.

This signal is sent from the A5 LKAS camera and doesn’t do anything but bring us closer to stock LKAS behavior. I can provide a route if needed.

tealtwo: this signal was also a “SET ME” bit in the DBC in previous version of openpilot and was just forgotten

jyoung8607: In the future yeah, I'd like to see routes for this sort of thing, but PQ being in dashcam mode we're a little more free to move quickly.

PR #2728, opened 2025-09-03
VW PQ: update Bremse_11 signal to add missing counter & checksum

tealtwo:


tealtwo:
marking this as a draft because PQ platform vehicles with MK100PQ do not have a Counter or Checksum in the Bremse_11 message and this will require further work.

tealtwo:
another note, it seems some VW PQ vehicles without ACC do not have the Bremse_11 message present at all

tealtwo:
okay after testing on MK100PQ ABS this PR is fine to merge since we do not parse B11 for anything useful in stock openpilot.

PR #2733, opened 2025-09-04
refactor: button event handling with ButtonEventConfig and ButtonMap

jakethesnake420:
This looks like a better button event interface. Should be able to simplify gm and toyota logic further.

### DIFF
--- a/opendbc/car/honda/carstate.py
+++ b/opendbc/car/honda/carstate.py
@@ -2,7 +2,7 @@ import numpy as np
 from collections import defaultdict
 
 from opendbc.can import CANDefine, CANParser
-from opendbc.car import Bus, create_button_events, structs
+from opendbc.car import Bus, create_button_events, structs, ButtonEvent, ButtonMap
 from opendbc.car.common.conversions import Conversions as CV
 from opendbc.car.honda.hondacan import CanBus
 from opendbc.car.honda.values import CAR, DBC, STEER_THRESHOLD, HONDA_BOSCH, HONDA_BOSCH_CANFD, \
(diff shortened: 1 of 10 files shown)


opendbc/car/honda/carstate.py
```diff
@@ -13,9 +13,16 @@
...
+BUTTON_MAPS: list[ButtonMap] = (
+  ButtonMap(ButtonType.accelCruise, CruiseButtons.RES_ACCEL),
+  ButtonMap(ButtonType.decelCruise, CruiseButtons.DECEL_SET),
+  ButtonMap(ButtonType.mainCruise, CruiseButtons.MAIN),
+  ButtonMap(ButtonType.cancel, CruiseButtons.CANCEL)
+)
+SETTINGS_BUTTON_MAPS: list[ButtonMap] = (
+  ButtonMap(ButtonType.gapAdjustCruise, CruiseSettings.DISTANCE),
+  ButtonMap(ButtonType.lkas, CruiseSettings.LKAS)
+)
```
sshane, line 25: why not a dict?

jakethesnake420: I changed it to a list so I could index into it without converting to a list because I had an idea but then I changed the create_button_events code so it was not necessary. I don't see why it couldn't be a dict now.

jakethesnake420: Actually, a dict would enforce that you can only have 1 button event for a button value which may be restrictive? So for example you could not have it so that pressing the Resume/cancel button creates 2 events, cancel and resume. Do you agree?

PR #2745, opened 2025-09-06
VM steering angle limits: clip to lateral accel while disengaged

sshane:


sshane:
@devtekve feel free to take this over, writing the test that catches this change first, then testing the new behavior

PR #2746, opened 2025-09-06
Vehicle model steering angle limiting: clip to lateral accel while disengaged

devtekve:
from https://github.com/commaai/opendbc/pull/2745

### DIFF
--- a/opendbc/car/lateral.py
+++ b/opendbc/car/lateral.py
@@ -116,17 +116,16 @@ def apply_steer_angle_limits_vm(apply_angle: float, apply_angle_last: float, v_e
   max_angle_delta = min(max_angle_delta, limits.ANGLE_LIMITS.MAX_ANGLE_RATE)
   new_apply_angle = rate_limit(apply_angle, apply_angle_last, -max_angle_delta, max_angle_delta)
 
-  # *** max lateral accel limit ***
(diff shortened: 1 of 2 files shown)


opendbc/car/lateral.py
```diff
@@ -120,9 +120,9 @@ def apply_steer_angle_limits_vm(apply_angle: float, apply_angle_last: float, v_e
...
   new_apply_angle = np.clip(new_apply_angle, -max_angle, max_angle)
 
-  # angle is current angle when inactive
+  # angle is current angle when inactive, but still limited to max lat accel to prevent faults.
   if not lat_active:
-    new_apply_angle = steering_angle
+    new_apply_angle = np.clip(steering_angle, -max_angle, max_angle)
 
   # prevent fault
   return float(np.clip(new_apply_angle, -limits.ANGLE_LIMITS.STEER_ANGLE_MAX, limits.ANGLE_LIMITS.STEER_ANGLE_MAX))
```
devtekve, line 128: Option B is:

```suggestion
  return float(np.clip(new_apply_angle, -max_angle, max_angle))
```

devtekve: if the max angle is bigger than the actual limits, it would fault, so this is not good

PR #2758, opened 2025-09-07
Honda - move brakehold_alt to honda_common

mvl-boston:
During a test run of bad code, it flagged in CI tests that the Honda Insight (Bosch A Hybrid) had the brakehold_alt message.   

Therefore, moving brakehold_alt to honda_common for use on the Insight.

Todo: check huggingface and see if there is a public log I can review.  Edit: confirmed message exists per private testlog.

CI discovery: https://github.com/mvl-boston/opendbc/actions/runs/17524028177/job/49772101396
Apparently the CI tests can see private testroutes where I can't.
So I now have a trick to search for messages in any testroute car by deliberately mis-coding and running a CI.


sshane:
good to merge?

mvl-boston:
> good to merge?

As luck has it, I just got an Insight route over DM today.   Will send you over discord as I'm not sure the driver intended it to go public.

It doesn't have rlogs but I can still see the message on Qlogs.   So good to merge from my perspective.

Edit: driver just uploaded rlogs, DM sent.

sshane:
We don't use this message so looks good

PR #2761, opened 2025-09-08
HKG: Add fingerprints and values to support Hyundai Sonata LF HEV (2018-2019)

hletrd:
Validation
* Dongle ID: 9f9b411a57b8ce21
* Route: 9f9b411a57b8ce21/00000005--af9e43ba56


hletrd:
## Current Status

* OP Steering and Safety Features
  * OpenPilot steering and all static safety features (brake pedal, gas pedal, cruise button mappings, door open/seatbelt latch checks) work well with the C-CAN bus. Basic car data and control signals are also provided via the C-CAN bus, allowing OpenPilot to receive all sensor states and gas pedal status.
 
* SCC Behavior
  * As with many HKG vehicles equipped with stock SCC, the stock ASCC module takes over when SCC is engaged. This behavior should be overridden by Comma in order to use Comma’s longitudinal control with experimental modes.
  * Radar and CAN2 Bus
    * The second CAN bus (which connects the ADAS module and the front radar) does not expose any raw radar tracks to the bus by default. The stock radar module can be activated to produce raw radar points for OpenPilot through the CAN2 bus by using a [dedicated script](https://github.com/commaai/openpilot/blob/master/selfdrive/debug/car/hyundai_enable_radar_points.py).

PR #2764, opened 2025-09-09
VW MQB: work around standstill limit

RJWadley:
depends on:
- https://github.com/commaai/openpilot/pull/37800
- https://github.com/commaai/opendbc/pull/3322

## goal
MQB cars without an EPB will fault if the ESP holds the brakes for too long at a standstill. This works around that fault to get full stop-n-go capability without an EPB.

If we transition from stopping to starting before the car has reached a standstill, we can ask for brake indefinitely without consequence. The car will happily comply with braking requests even in a starting state.

One side effect of leaving the stopping state early is that we are prone to rollback on very steep hills. The car takes about a second to build braking force, and if the car is traveling slowly this may be enough time for the car to start moving in reverse. To avoid rolling back, we estimate the slowest speed we can travel before rollback becomes a risk. If the car drops below this speed we force the car to stop.

ACC type 1 is also sensitive to control signals when the driver is pressing the brake (i.e. we are in a preEnabled state) so I am suppressing long_active in that case.

## validation route (outdated)
`9feef6a0323edf69/000002cb--53d4e34a83` driven on [cb2f550](https://github.com/commaai/opendbc/pull/2764/commits/cb2f55039243d4e7db34565faa58fef7a8f71d2e)


opendbc/car/volkswagen/mqbcan.py
```diff
@@ -96,8 +96,8 @@ def create_acc_accel_control(packer, bus, acc_type, acc_enabled, accel, acc_cont
     "ACC_Status_ACC": acc_control,
     "ACC_StartStopp_Info": acc_enabled,
     "ACC_Sollbeschleunigung_02": accel if acc_enabled else 3.01,
-    "ACC_zul_Regelabw_unten": 0.2,  # TODO: dynamic adjustment of comfort-band
-    "ACC_zul_Regelabw_oben": 0.2,  # TODO: dynamic adjustment of comfort-band
+    "ACC_zul_Regelabw_unten": 0 if starting else 0.2,  # TODO: dynamic adjustment of comfort-band
+    "ACC_zul_Regelabw_oben": 0 if starting else 0.2,  # TODO: dynamic adjustment of comfort-band
```
RJWadley: I experienced a few cases where openpilot got stuck at standstill because accel never exceeded the comfort band

opendbc/car/volkswagen/mqbcan.py
```diff
@@ -108,9 +108,9 @@ def create_acc_accel_control(packer, bus, acc_type, acc_enabled, accel, acc_cont
   if starting:
     acc_hold_type = 4  # hold release / startup
   elif esp_hold:
-    acc_hold_type = 3  # hold standby
+    acc_hold_type = 1  # hold request (at stop)
```
RJWadley: these were flipped from how stock ACC uses them. the car will still respect either value and stop, but sometimes the car is late to release ESP_confirmation (ESP_Haltebestaetigung) when transitioning directly from 3 -> 4.

on cars with EPB this is fine, but on cars without this can easily cause faults if the hold confirmation is retained for too long.

opendbc/car/volkswagen/carcontroller.py
```diff
@@ -79,11 +80,29 @@ def update(self, CC, CS, now_nanos):
...
-        stopping = actuators.longControlState == LongCtrlState.stopping
-        starting = actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or CS.out.vEgo < self.CP.vEgoStopping)
-        can_sends.extend(self.CCS.create_acc_accel_control(self.packer_pt, self.CAN.pt, CS.acc_type, CC.longActive, accel,
+        # acc is picky about what signals we can send while brake is pressed (e.g. preEnabled at standstill)
+        longActive = CC.longActive and not CS.out.brakePressed
+
+        acc_control = self.CCS.acc_control_value(CS.out.cruiseState.available, CS.out.accFaulted, longActive)
+        accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if longActive else 0)
+        stopping = longActive and actuators.longControlState == LongCtrlState.stopping
+        rollout_in_progress = CS.out.vEgo < self.CP.vEgoStopping and accel > 0
```
RJWadley: the previous logic would often flip 'starting' to true when slowing to a stop, which is not what the car expects

I'm restricting 'starting' to only positive acceleration here

opendbc/car/volkswagen/carcontroller.py
```diff
@@ -79,11 +80,29 @@ def update(self, CC, CS, now_nanos):
...
+        starting = longActive and actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or rollout_in_progress)
+
+        # reset standstill timer for MQB w/out EPB while braking
+        # this mimics what would happen if the stock ACC intentionally rolled off after holding for too long
+        # the key difference is that the stock ACC also enables freewheel (ACC_Freilauf_Info) during this time
+        # but we leave it disabled to reset the timer without actually rolling off
+        if longActive and stopping and CS.esp_hold_confirmation and not (self.CP.flags & VolkswagenFlags.PQ) and CS.acc_type == 1:
```
RJWadley: Is `CS.acc_type == 1` the correct way to detect ACC Low cars that can't hold at stop?

```
VAL_ 290 ACC_Typ 
0 "Basis_ACC" 
1 "ACC_mit_FollowToStop" 
2 "ACC_mit_StopAndGo" 
3 "ACC_nicht_codiert" ;
```

RJWadley: I've seen from discord logs that ACC High disengages and enables parking brake after three minutes at standstill. If this is still the case in 2025, then applying this workaround to all MQB (instead of just ACC Low) might remove that time limit and keep OP enabled indefinitely.

I don't have an ACC high car to test with though, so I can't verify.

tealtwo: > I've seen from discord logs that ACC High disengages and enables parking brake after three minutes at standstill. If this is still the case in 2025, then applying this workaround to all MQB (instead of just ACC Low) might remove that time limit and keep OP enabled indefinitely.
> 
> I don't have an ACC high car to test with though, so I can't verify.

I have a ACC High VW Atlas, it holds indefinitely at a stop from my use of stock VW ACC, the max time I have tested is 5 minutes without EPB engaging, but it may be longer

tealtwo:
before merge the openpilot long controller will have to be adapted to work with PQ as well, the current implementation may not function properly with PQ vehicles running stock openpilot

opendbc/car/volkswagen/carcontroller.py
jyoung8607: > Is `CS.acc_type == 1` the correct way to detect ACC Low cars that can't hold at stop?

We've seen conflicting data in the past, but I *believe* that ACC Low/High are actually references to max engagement speed. ACC High means you have enough "look-ahead" to track lead cars under German autobahn speeds. I believe that can be obtained either with a LRR, or a MRR combined with a camera.

FtS vs SnG capability are separate from the "High/Low" capability.

opendbc/car/volkswagen/mqbcan.py
jyoung8607: I'm sure there's tuning gains to be had, but file that in a separate PR first so I can validate.

opendbc/car/volkswagen/carcontroller.py
```diff
@@ -79,11 +82,38 @@ def update(self, CC, CS, now_nanos):
...
+
+        accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if longActive else 0)
+        stopping = longActive and actuators.longControlState == LongCtrlState.stopping
+        rollout_in_progress = CS.out.vEgo < self.CP.vEgoStopping and accel > 0
+        starting = longActive and actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or rollout_in_progress)
+
+        # reset standstill timer for MQB w/out EPB while braking
+        # this mimics what would happen if the stock ACC intentionally rolled off after holding for too long
+        # the key difference is that the stock ACC also enables freewheel (ACC_Freilauf_Info) during this time
+        # but we leave it disabled to reset the timer without actually rolling off
```
jyoung8607: Are you sure you have a car that actually supports freewheeling? This seems to be rare, potentially dry-clutch DSGs only. IIRC we've only had one show up, and IIRC they might be broken without passing through another flag.

opendbc/car/volkswagen/mqbcan.py
jyoung8607: I'll take a look when you're out-of-draft and ready for real review, but IIRC I got those field descriptions from an authoritative source back in the day. That's not to say we're for-sure using them correctly.

jyoung8607 (commented):
Nice work finding a way to hold longer! I've never been happy with our (lack of) proper FtS support, and this might be an even better solution.

RJWadley:
I've been doing some fairly extensive tests on this over the last month or so to try and improve the behavior and reduce faults. The biggest problem right now is hills.

If we're on a hill (specifically uphill), we have to send a longer 'start request' before the car will acknowledge.
If the signal is too long, the car may press the gas and the brakes at the same time, which is a tad uncomfy.
If we're on a very steep hill, the request might be impossible to send as we don't have enough time before faulting.

I'm looking for some kind of workaround, but I'm leaning toward just doing the best we can while also asking the user to press brake on steeper hills. (The dashboard has a 'press brake' alert we can use for that)

opendbc/car/volkswagen/mqbcan.py
RJWadley: after further testing, I don't think it matters like I thought it did. I'll probably leave it this way since its more "correct" but I am not entirely sure if the ESP on my car even uses this value

opendbc/car/volkswagen/mqbcan.py
RJWadley, line 120: not really related, but `ACC_Folgebeschl` = "acceleration of the car we're following relative to our car"

opendbc/car/volkswagen/mqbcan.py
RJWadley: Yeah there's a couple weird (tune-related) behaviors at standstill that I haven't quite figured out yet.

I think I'll focus the work in this PR on standstill fault prevention and may circle back to tuning stop behavior later

jyoung8607: If you get into stop tuning, this is key: https://github.com/commaai/opendbc/blob/8ac2dc9566cb06ca098a2734af34cb79c45691c0/opendbc/car/volkswagen/mqbcan.py#L118

That's always been just a placeholder. It's known to be a distance in meters. If you want to figure out the math/code necessary to compute the distance to a nice smooth stop based on the car's current speed and acceleration at the time we transition to stopping, that would help a lot. I think your FtS car will use this as well.

opendbc/car/volkswagen/carcontroller.py
RJWadley: > Are you sure you have a car that actually supports freewheeling?

I have no idea, really. But sending different values for `ACC_Freilauf_Info` does have a measurable effect.

If I send 0 (freewheel allowed) the car will immediately start rolling. If we're facing uphill, the car will start rolling backward. If I send 2 (freewheel not allowed) the car will remain at a stop for a moment before rolling forward (assuming no brake request, ofc). It won't ever roll backwards, even if I send ACC disabled.

I haven't experimented much with 1 (No transition to coasting allowed) and 3 (Coasting requested) but plan to soon. The stock system sends 0 when forcing FTS disengagement. EDIT: after testing neither value was useful to me but they did also have an impact

opendbc/car/volkswagen/mqbcan.py
jyoung8607: If needs must, you're welcome to add FtS-specific conditional behavior here.

opendbc/car/volkswagen/carcontroller.py
jyoung8607: Re: standstill time, I concur with @tealtwo, I tested this a long time ago on my cars. There is indeed a three minute timer, but the stock radar is running it. The car is happy to sit longer for openpilot.

jyoung8607:
> I've been doing some fairly extensive tests on this over the last month or so to try and improve the behavior and reduce faults. The biggest problem right now is hills.

You may find this challenging. One of the differences between FtS and SnG cars is bidirectional rear wheel speed sensors, for detecting when the wheel is rolling backwards, to protect against rollback on hill starts. Essentially the same problem you're having. With that info, I suspect the ESP is able to keep the car still until the powertrain develops enough force to overcome the hill. Same basic thing happens with manual trans hill-hold. I don't know how well FtS cars deal with this.
 
> I'm looking for some kind of workaround, but I'm leaning toward just doing the best we can while also asking the user to press brake on steeper hills. (The dashboard has a 'press brake' alert we can use for that)

Not exactly Plan A, but we do have longitudinal pitch data available in carcontroller. I wouldn't trust the car's road-grade signal for this.

opendbc/car/volkswagen/mqbcan.py
jyoung8607: Where are you getting that? That's not reflected in docs I've seen (not that they said much useful) and it doesn't match observations from my car's stock ACC.

RJWadley: At first I thought it was an accel command to our car (and was wondering if I might be able to reset standstill somehow using it - that's the main reason I was investigating in the first place)

Keep in mind I have a sample size of one car, but in testing I made a few key observations of the stock behavior:
- This signal correlates with `ACC_Abstandsindex`. When that signal is neutral, so is this one.
- This signal is sent even when cruise is not engaged and the car isn't moving. If I'm behind a car at a light, this signal will increase as they drive away, even if I never move and I don't have cruise engaged.
- If I'm following a car moving at a steady pace and I slam on my breaks, the signal goes very negative. 

happy to share a few segments if you'd like to investigate/compare

jyoung8607: If you graph it in PJ against the normal accel control signal, such that they're scaled properly, I suspect you'll find they're identical or almost identical while ACC is engaged. However, as you say, it keeps sending projected accelerations even when ACC is not engaged, still tracking the lead car. I have a suspicion it might be used to improve DSG shift predictions. Not sure what else.

RJWadley: will experiment and follow up in discord

opendbc/car/volkswagen/mqbcan.py
RJWadley: removed in favor of future PR

opendbc/car/volkswagen/mqbcan.py
RJWadley: after further testing it doesn't seem to matter either way

opendbc/car/volkswagen/carcontroller.py
RJWadley: removed for brevity, may restore later for correctness

sshane:
@jyoung8607 do you expect we can merge this soon?

sshane:
This is a lot of code. I think we should consider we may not want to support SnG on cars without EPB, especially because all the auto manufacturers don't support it. If you open the door the car will roll forward vs. cars with EPB safely putting on the parking brake

RJWadley:
I kept this as one large PR for the purposes of testing and validation. Now that the approach has been proven, I would be comfortable splitting it into smaller PRs that are easier to merge if necessary. Even in its full state, this should only touch alpha long on type 1 cars.

---

FtS vs SnG capability is by far the most confusing compatibility issue for VW, and eliminating it would be a massive win for the car compatibility docs. jyoung said it best:
"[VW FtS vs SnG is the worst potential inaccuracy in the compatibility guide of which I'm aware.](https://discord.com/channels/469524606043160576/534359517836607488/1073037604007727115)"


A few things to consider with respect to support in general:
<details>
<summary><b>Should EPB be a general requirement for stop and go?</b></summary>
<br>
There are other cars that already support SnG without an EPB that behave just like these cars do, so it's not currently a hard requirement. I don't think openpilot should require EPB, but a case could be made either way.
</details>

<details>
<summary><b>What happens if I open the door?</b></summary>
<br>
This is actually already handled pretty well. The car <em>and</em> openpilot will yell at you, and openpilot will keep the brake held for 3 seconds to give the driver time to take over. 
<br>
<br>
Here's a segment, if you'd like to see the behavior: <code>9feef6a0323edf69/0000030e--8257e4f64a/9</code>
</details>

<details>
<summary><b>Does this hurt the brakes?</b></summary>
<br>
Nope. We are operating well within the manufacturer's limits using a braking command that VW already uses in stock.
</details>

RJWadley:
I've updated the branch, but don't have permission to reopen this PR. I'll create a new PR soon unless a maintainer can reopen this one.

PR #2765, opened 2025-09-09
CANParser: Improved error logging

jyoung8607:
After the recent CANParser rewrite, we lost logging of missing messages that cause canValid == False, which makes debugging not very fun.

Added that logging back, did some other minor cleanup, and rate-limited error logging to once per second per message so we don't massively inflate the logs.

Example:

```
CANParser: 0x296 SCM_BUTTONS not valid (timeout or missing)
CANParser: 0x296 SCM_BUTTONS not valid (timeout or missing)
CANParser: 0x309 CAR_SPEED not valid (timeout or missing)
CANParser: 0x309 CAR_SPEED not valid (timeout or missing)
CANParser: 0x158 ENGINE_DATA not valid (timeout or missing)
CANParser: 0x158 ENGINE_DATA not valid (timeout or missing)
```

### DIFF
--- a/opendbc/can/parser.py
+++ b/opendbc/can/parser.py
@@ -204,7 +204,7 @@ class CANParser:
     for state in self.message_states.values():
       if state.counter_fail >= MAX_BAD_COUNTER:
         counters_valid = False
-        state.rate_limited_log(self._last_update_nanos, f"counter_fail >= {MAX_BAD_COUNTER}")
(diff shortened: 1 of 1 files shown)


jyoung8607:
@adeebshihadeh new hotness CANParser is your baby, thought you might have an opinion.

`last_warning_log_nanos` doesn't handle the case of time going backwards (replay wrap-around or rewind), but neither do `first_seen_nanos` or `timestamps`. If you consider that a problem, we can check for time warps in another PR.

PR #2767, opened 2025-09-09
Honda: Pitch compensation for Bosch throttle

jyoung8607:
Picked in concept from #2165, with filtering inspired by Shane's work on Toyota.


jyoung8607:
Doesn't work quite right yet, will need to switch the gas actuator from accel-based to torque-based first.

PR #2772, opened 2025-09-10
Honda - add pitch compensation to Nidec

mvl-boston:
Honda Nidecs benefit from pitch compensation for gas and brake.   Have been testing this for a month on an openpilot fork.

Adding this change in a similar manner as the Bosch change in earlier versions of https://github.com/commaai/opendbc/pull/2488

Todo:
- [x] Await upstream of https://github.com/commaai/opendbc/pull/2488
- [x] Rebaseline with any further changes in https://github.com/commaai/opendbc/pull/2488
- [ ] Separate PR needed to fix gas control, to make the impact of this fix more obvious in a test route
- [ ] Create pre and post testroutes with this fix

### DIFF
--- a/opendbc/car/honda/carcontroller.py
+++ b/opendbc/car/honda/carcontroller.py
@@ -12,8 +12,8 @@ from opendbc.car.interfaces import CarControllerBase
 VisualAlert = structs.CarControl.HUDControl.VisualAlert
 LongCtrlState = structs.CarControl.Actuators.LongControlState
 
-
 MAX_PITCH_COMPENSATION = 1.5  # m/s^2
+
 
 def compute_gb_honda_bosch(accel, speed):
   # TODO returns 0s, is unused


mvl-boston:
Putting this on hold as test route is not obviously better.    Need to move from speed control to gas pedal control to make impact more visible, staying as draft until a PR for pedal control is working.

sshane:
I had tried to add pitch compensation to Toyota when trying to understand and undo the internal PCM compensations it does to our acceleration request, but was never able to get it working reliably in all scenarios.

Highpassing (~derivative) the pitch, and then allowing the integral controller to wind up after pitch changes worked much more reliably and more responsively. It removed any steady state error if the pitch estimation was wrong, or if the factor to go from pitch -> accel (gas/brake in your case) is totally unknown.

Unclear if it would benefit you here, but you can copy what I did for Toyota and report back: https://github.com/commaai/opendbc/blob/fce60b416389faf8d37c685eec4b34e5c85767bf/opendbc/car/toyota/carcontroller.py#L231-L234

sshane:
I found straight pitch compensation caused strange effects leaving/entering standstill (not enough brake or too much and that was a pure accel request), so I worry about the same kinds of edge cases here. Closing unless you want to try the derivative of pitch solution I merged for Toyota

PR #2773, opened 2025-09-11
Honda: New HUD experience for ALT_RADAR cars

jyoung8607:
Different HUD experience I've been driving with on the 2024 Odyssey. Think this is more faithful to stock than the current openpilot experience, and gives the driver more useful information. It's not verified on any other cars yet, so applying to ALT_RADAR cars only.

Spiritual successor to #2473.

### DIFF
--- a/opendbc/car/honda/hondacan.py
+++ b/opendbc/car/honda/hondacan.py
@@ -161,7 +161,7 @@ def create_acc_hud(packer, bus, CP, enabled, pcm_speed, pcm_accel, hud_control, 
   return packer.make_can_msg("ACC_HUD", bus, acc_hud_values)
 
 
-def create_lkas_hud(packer, bus, CP, hud_control, main_on, lat_active, reduced_steering, alert_steer_required, lkas_hud):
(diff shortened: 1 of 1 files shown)


jyoung8607:
@mvl-boston Feedback welcome on this, I think it accomplishes what you needed.

opendbc/car/honda/carcontroller.py
```diff
@@ -223,7 +223,10 @@ def update(self, CC, CS, now_nanos):
         can_sends.append(hondacan.create_acc_hud(self.packer, self.CAN.pt, self.CP, CC.enabled, pcm_speed, pcm_accel,
                                                  hud_control, hud_v_cruise, CS.is_metric, CS.acc_hud))
 
-      can_sends.extend(hondacan.create_lkas_hud(self.packer, self.CAN.lkas, self.CP, hud_control, alert_steer_required, CS.lkas_hud))
+      steering_available = CS.out.cruiseState.available and CS.out.vEgo > self.CP.minSteerSpeed
+      reduced_steering = CS.out.steeringPressed
```
jyoung8607, line 227: @mvl-boston I didn't end up including the early steer-saturated indicator, but if you wanted to play with it in your fork, you would just add the condition to `reduced_steering`.

PR #2774, opened 2025-09-11
Nissan: Use vehicle model for lateral

gncnpk:


gncnpk:
Needs tests for lateral acceleration and jerk, still figuring out how safety testing works.

PR #2777, opened 2025-09-11
Honda: Additional ALT_RADAR platform support

jyoung8607:
* EPS and VSA firmware are only returned via bus 1 in multiplexing mode, exceptions added
* Improved EPS fault mitigation
* Improved below-steer-speed alert handling

TODO: delay exiting the below-steer-speed state for a few seconds on ALT_RADAR cars


jyoung8607:
This was pretty successful in reacting to EPS fault states due to driver overrides. We raise `steerFaultTemporary` and release `CC.latActive`, but openpilot knows to [skip the user-facing alert if the driver is overriding](https://github.com/commaai/openpilot/blob/10580aca922789c58fae1a6922193df9b2fdf311/selfdrive/car/car_specific.py#L191-L202). The EPS and ALT_RADAR filter seem to come back available right away once the EPS fault state clears, which happens quickly.

image

PR #2786, opened 2025-09-12
(Bolt) Copy longitudinal tuning from Starpilot

simontheflutist:
Stock tuning for experimental long is nauseating. While I haven't fine tuned at all, the [numbers from Starpilot](https://github.com/firestar5683/StarPilot/blob/TorqueTune/selfdrive/car/gm/interface.py#L116) are a night-and-day vibes improvement.

Validation
* Dongle ID: 
* Route:

### DIFF
--- a/opendbc/car/gm/interface.py
+++ b/opendbc/car/gm/interface.py
@@ -111,7 +111,7 @@ class CarInterface(CarInterfaceBase):
       ret.vEgoStopping = 0.25
       ret.vEgoStarting = 0.25
 
-      if candidate in CAMERA_ACC_CAR:~
+      if candidate in CAMERA_ACC_CAR:
         ret.longitudinalTuning.kiV = [0.5] * len(ret.longitudinalTuning.kiBP)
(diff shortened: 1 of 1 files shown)


jyoung8607:
Happy to merge if the data show it's an improvement, but we will need that data, before and after. I'm assuming that's why this is marked draft.

Reports from the longitudinal test maneuver tool would be ideal, this kind of PR is what it's there for.

simontheflutist:
Playing around with this fork (openpilot `simontheflutist:long-like-starpilot`) and curious about an unexpected behavior.

- When alpha long is enabled, engaging ACC at <15mph sets the ACC setpoint to 25mph, whereas
- When alpha long is disabled, engaging ACC at <15mph sets the ACC setpoint to 15mph (minimum on the Bolt).

However, despite the higher minimum when engaging with alpha long, I can still use the +/- keys to lower the ACC setpoint down to 15mph after engaging.

jyoung8607:
Per our conversation from last month, feel free to reopen this with before-and-after test data.

PR #2787, opened 2025-09-12
HKG: Add FW for 2024 Genesis GV70

idy0204:
Validation
* Dongle ID: 1e68c8305f7c8201
* Route: 1e68c8305f7c8201/00000000--c96be7ee8a


idy0204:
Openpilot Comma X3 has issue with recognizing 2024 GV70 and found out that the  fwdcamera is not list on the fingerprint list.
Please check the route and push it to master.

PR #2788, opened 2025-09-13
Honda - add back removed 2020 RDX fingerprint

mvl-boston:
add back 2020 RDX fingerprint removed in https://github.com/commaai/opendbc/pull/2778

source: https://github.com/commaai/openpilot/commit/8c2bec03af58add66dd2ab72ca852d30879cb06f


jyoung8607:
Thanks for the fix! This was indeed a regression on my part, I misread commaai/openpilot#27812 as adding two EPS firmware entries, but it was really just adding one and appending a comma to the other.

PR #2789, opened 2025-09-14
Update CARS.md. combine Niro EV years

EfrB:


sshane:
You shouldn't edit the generated file. You can edit https://github.com/commaai/opendbc/blob/master/opendbc/car/hyundai/values.py, or make a PR to https://github.com/commaai/opendbc/blob/master/opendbc/car/docs.py to automatically combine model year entries

PR #2790, opened 2025-09-14
CI: bugfix of tests by pinning pycapnp to 2.1.0

devtekve:
Pipelines broke since last night https://github.com/commaai/opendbc/actions/runs/17701582104/job/50307962453 

(also seen in sunnypilot) due to a pycapnp update from `2.1.0` to `2.2.0`, this pins so we are back to "working order" but probably we should embrace the change? i'm unfamiliar with the change tho.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
   "numpy",
   "crcmod",
   "tqdm",
-  "pycapnp",
+  "pycapnp==2.1.0",
   "pycryptodome",
 ]
 


jyoung8607:
@devtekve Thanks for tracking down the issue. Looks like it stems from capnproto/pycapnp#380.

@adeebshihadeh This might be a blessing in disguise; take a look at that pycapnp PR. There might be some openpilot messaging performance gains to be found. It might make sense to look at deleting VisionIPC again.

PR #2793, opened 2025-09-16
Capnp memoryview

sshane:
@jyoung8607 interested in finishing this?


sshane:
actually that was easier than I thought. @jyoung8607 any other concerns of yours?

jyoung8607:
I didn't think the byte casting would be hard, I just didn't have time to think through whether all the usages would be okay with the extra copy to bytes, or making sure that AGNOS is synced and that we can upgrade back and forth.

sshane:
Have to merge one or the other with a failing test first since test_models is split

PR #2671, opened 2025-08-17
Honda - Nidec Hybrid brake control

mvl-boston:
Support for Honda separate brake control messages used by Nidec hybrids
- Needed for expected-upstreamable 2020 Acura RLX
- Set as a generic flag because this is shared by many other non-upstreamable cars (with non-standard steer wiring)
- update brakehold as some (not all) hybrids use a different message
- prevents fault when attempting to resume cruise directly from brakehold, requires companion [openpilot PR36121](https://github.com/commaai/openpilot/pull/36121)
- removes brake hysteresis for hybrids as it causes braking vibration 
- Pre-refactor was tested successfully on non-upstreamed Hybrids: Nidec MDX, RLX, Clarity
- Post-refactor tested successfully on Nidec MDX Hybrid: ad9840558640c31d/0000007a--789d1eda01
- Additional brakehold fault logic retested: ad9840558640c31d/00000002--531e9d28b4

PR #2795, opened 2025-09-17
 Honda: Dashcam support for Accord 9G - serial steering

mvl-boston:
Part of a series of PRs to support serial steering cars in dashcam mode. These cars use linbus for steering and thus require a community translation board to convert canbus signals to linbus for the car to understand.

In total, this series would allow dashcam support for 1.2 million additional cars per North American sales statistics.

This PR supports:
- 16-17 Accord (ACCORD_9G)

Other PRs will support (pending route logs to open):
- 14-17 MDX (ACURA_MDX_3G)
- 18-20 MDX (ACURA_MDX_3G_MMR - different dbc / newer steer status)
- 15-20 TLX (ACURA_TLX_1G)

Note: Hybrid versions of the Accord and MDX additionally require https://github.com/commaai/opendbc/pull/2671, which is baselined into this PR.

Notes:
- Used legacy pid, because torque controller did not drive as well.
- Created new steering control d.   steering control c reserved for Singapore Odyssey


opendbc/car/extra_cars.py
```diff
@@ -45,7 +52,8 @@ class CAR(Platforms):
       CommunityCarDocs("Acura TLX 2015-17", "Advance Package"),
       CommunityCarDocs("Acura TLX 2018-20", "All"),
       GMSecurityCarDocs("Acura ZDX 2024", "All"),
-      CommunityCarDocs("Honda Accord 2016-17", "Honda Sensing"),
+      HondaSerialSteeringCarDocs("Honda Accord 2016-17", "Honda Sensing"),
+      HondaSerialSteeringCarDocs("Honda Accord Hybrid 2017", "All"),
```
jyoung8607: You can put these in the car port directly. Look at the Toyota SecOC cars as an example.

jyoung8607 (changes requested):

adeebshihadeh:
We're not really interested in merging support for cars that don't have a path to being upstreamed.

jyoung8607:
> We're not really interested in merging support for cars that don't have a path to being upstreamed.

Can we reserve judgment on that for now? This PR looks like a Whole Big Thing, but it won't merge that way:

* The diff is mostly #2671, which will merge first, for supported cars
* Separately, a minor generator refactor to promote the no-longer-snowflake [Acura RDX gear message](https://github.com/commaai/opendbc/blob/8d1fbe01bfba3f7fd25d3d950876884565f024c4/opendbc/dbc/generator/honda/acura_rdx_2018_can.dbc#L8-L15)

Once the pre-refactors (prefactors?) are done, the extra carrying-cost of a dashcam-mode serial steering car should be down to a single DBC generator template with a single message in it.

We've told community we'll be more permissive about dashcam merges. Let's take another look once the diff is reduced.

adeebshihadeh:
I'd like to get to a point where carrying these are free, but we are very much not there, as evidenced by the 144 open pull requests in this repo.

Until we get to that point, I absolutely do not want anything in this repo that makes it harder to get upstream-able cars upstreamed.

opendbc/car/extra_cars.py
mvl-boston: Let me know if this change is correct/complete.   Not clear on all the aspects of cardocs.

mvl-boston:
Closing per adeeb's request.   Not willing to prioritize 3rd party hardware for dashcam mode at this time.

PR #2794, opened 2025-09-17
Honda DBC updates

vanillagorillaa:
Nerd sniped myself looking for a different signal. 

39563a18fe07169f/00000028--d3800a1d4b

PR #2797, opened 2025-09-17
Honda: Merge Kinematics

vanillagorillaa:
Merging as discussed in https://github.com/commaai/opendbc/pull/2794

Will make this a simple merge PR, and add other signals and fixes in the other PR.

Bosch Kinematics seems to have a bit different signal scaling and offset than nidec. Leaving as draft for the moment to verify.


vanillagorillaa:
I think the bosch kinematics has just been wrong since it was added. 

will put the other signal cleanups in my other PR.

PR #2798, opened 2025-09-18
Honda: fcw_distance fix

vanillagorillaa:
Not sure this may be correct


sshane:
What does this do?

vanillagorillaa:
The single bit `FCM_OFF` and `FCM_OFF_2` signals are actually one two bit signal. This fixes the dbc definition, and fixes the use in hondacan

PR #2799, opened 2025-09-19
try pypy

adeebshihadeh:


adeebshihadeh:
Seems this won't work with pycapnp due to https://github.com/cython/cython/issues/1818

PR #2800, opened 2025-09-20
try capnpy

adeebshihadeh:


adeebshihadeh:
I think we're just gonna have to write our own thing.

PR #2802, opened 2025-09-22
Nissan: Add FW for 2018 Nissan Leaf

Sathans:
Added Nissan leaf IC updated firmware

Validation
* Dongle ID: 44211d9f2879657d
* Route: 44211d9f2879657d/0000005b--2db356e40d

### DIFF
--- a/docs/CARS.md
+++ b/docs/CARS.md
@@ -242,7 +242,7 @@
 |Mazda|CX-5 2022-25|All|[Upstream](#upstream)|
 |Mazda|CX-9 2016-20|All|[Dashcam mode](#dashcam)|
 |Mazda|CX-9 2021-23|All|[Upstream](#upstream)|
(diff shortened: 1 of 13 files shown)


jyoung8607:
What model-year is this? VIN decoder came back with MY2014, which was the previous generation Leaf and I'm not sure it even has ADAS, but your car seems to be engaging successfully.

sshane:
VPIC can't decode it, it's an EU car. I wouldn't rely on it. Global model year character is often wrong

Sathans:
It's an EU car, Nissan Leaf 40kWh Tekna (2018).
A few months ago I had uninstalled frogpilot and installed sunny, then it didn't work anymore. I assume that I had always set the fingerprint manually, or there was some update while I had frog installed. But here are the new fingerprints

PR #2804, opened 2025-09-24
VW PQ: fix incorrect GRA_Zeitluecke value

tealtwo:
this PR fixes the incorrect value for the GRA_Zeitluecke signal in the GRA_Neu message for the multifunction steering wheel buttons to allow for the ACC distance button to be used to change openpilot longitudinal control's personality setting, route will be provided after fixes have been tested


tealtwo:
why did only the macOS build test fail???

sshane:
You can ignore that. Is this good to go? Route?

tealtwo:
> You can ignore that. Is this good to go? Route?

it's been tested & confirmed working, grabbing route today

tealtwo:
ready to merge `0f53129ed44f6920/00000049--a89e8c8360`

PR #2805, opened 2025-09-26
KIA FORTE 24 ADD

shurtadohwnew:
Validation
* Dongle ID: 
* Route:


sshane:
Can you provide a dongle ID?

shurtadohwnew:
Dongle ID: 850678021a7879cc

Route ID using Kia Forte 2023 car paramters on sunnypilot
850678021a7879cc/00000001--deed8039bc

The latest commits are modifications from the previous ones, just the latest ones needed

opendbc/car/hyundai/hyundaican.py
```diff
@@ -39,7 +39,7 @@ def create_lkas11(packer, frame, CP, apply_torque, steer_req,
                            CAR.HYUNDAI_ELANTRA_HEV_2021, CAR.HYUNDAI_SONATA_HYBRID, CAR.HYUNDAI_KONA_EV, CAR.HYUNDAI_KONA_HEV, CAR.HYUNDAI_KONA_EV_2022,
                            CAR.HYUNDAI_SANTA_FE_2022, CAR.KIA_K5_2021, CAR.HYUNDAI_IONIQ_HEV_2022, CAR.HYUNDAI_SANTA_FE_HEV_2022,
                            CAR.HYUNDAI_SANTA_FE_PHEV_2022, CAR.KIA_STINGER_2022, CAR.KIA_K5_HEV_2020, CAR.KIA_CEED,
-                           CAR.HYUNDAI_AZERA_6TH_GEN, CAR.HYUNDAI_AZERA_HEV_6TH_GEN, CAR.HYUNDAI_CUSTIN_1ST_GEN, CAR.HYUNDAI_KONA_2022):
+                           CAR.HYUNDAI_AZERA_6TH_GEN, CAR.HYUNDAI_AZERA_HEV_6TH_GEN, CAR.HYUNDAI_CUSTIN_1ST_GEN, CAR.HYUNDAI_KONA_2022, CAR.KIA_FORTE):
```
jyoung8607, line 42: Was it necessary to add this in order for your car to be usable? This is a substantial change for all other model-years of Kia Forte. Even if it's right for your car, we'd have to see if it breaks older Forte.

@sunnyhaibin this is my least favorite line of code in all of opendbc. Is there no way to auto-detect if we have a more modern instrument cluster?

sshane: I tried a while ago but couldn't find any pattern in this message alone. That's not to say it's not possible though

shurtadohwnew: Not necesary, this change should be omitted

PR #2807, opened 2025-09-27
safety: cleanup remaining panda bits

adeebshihadeh:


opendbc/safety/safety.h
```diff
@@ -335,7 +335,6 @@ void safety_tick(const safety_config *cfg) {
 
 static void relay_malfunction_set(void) {
   relay_malfunction = true;
-  fault_occurred(FAULT_RELAY_MALFUNCTION);
```
adeebshihadeh, line 338: moved to panda: https://github.com/commaai/panda/pull/2281

PR #2818, opened 2025-09-29
Honda: Support for Traffic Jam Assist steering control bit

jyoung8607:
Resurrect the support pieces of #2104, which was backed out in #2240.

The change itself works great, on the cars where it works, what remains is to establish the set of cars it works on.


jyoung8607:
@vanillagorillaa @mvl-boston This makes it so you just need to add a single flag to a model to enable the steer-to-zero bit. The first model to use it will be #2410. If you're aware of other models this works on, happy to review PRs to add them individually.

Issue #2781, opened 2025-09-11
Kia Niro EV 2024 (non HDA II) not recognized


EfrB:
Should be supported but says dashcam mode 
Route with stock openpilot:
https://connect.comma.ai/674d529ed17f9d41/00000002--d709d7694b

Sathans:
Not sure if this is the reason, but in from your comma there is also another fwdRadar with different fwVersion': 
``` 
{'address': 2000,
'brand': 'hyundai',
'bus': 0,
'ecu': 'fwdRadar',
'fwVersion': b'\xf1\x8b #\x12\x06',
'logging': True,
'obdMultiplexing': True,
'request': [b'"\xf1\x8b'],
'responseAddress': 2008,
'subAddress': 0}
```

Also for fwdCamera: 
```   
{'address': 1988,
'brand': 'hyundai',
'bus': 0,
'ecu': 'fwdCamera',
'fwVersion': b'\xf1\x8b #\x12\x04',
'logging': True,
'obdMultiplexing': True,
'request': [b'"\xf1\x8b'],
'responseAddress': 1996,
'subAddress': 0}```

EfrB:
Added
#2825

sshane:
openpilot recognizes your vehicle and there is no dashcam only flag in that route. What happens when you engage?

PR #2825, opened 2025-10-05
Fix for Kia Niro EV 2024 (non HDA II)

EfrB:
Validation
* Dongle ID: 
* Route: 
#2781


sunnyhaibin:
- This is currently failing CI.
- These are not valid fingerprints for any ECUs of Hyundai/Kia/Genesis, so your car isn't actually being fingerprinted properly with the UDS responses. Do you have a route?

PR #2826, opened 2025-10-05
cleanup red panda kit leftovers

adeebshihadeh:
Might need to handle this in test_models...


adeebshihadeh:
Yeah, will revisit later. Merged the docs cleanup separately.

PR #2829, opened 2025-10-07
Honda: exclude TJA message for Nidec

mvl-boston:
Nidec DBC does not have steer_to_zero message, so logic in [pr 2818](https://github.com/commaai/opendbc/pull/2818) should be restricted to Bosch.

Otherwise errors like errorLogMessage within  [this log](https://useradmin.comma.ai/?onebox=c3570664c1015c08%2F00000007--ea01281334) occur:


jyoung8607:
Thanks for calling this out, it's a bug. That said, I'm going to simplify this a bit before merging.

jyoung8607:
Ended up being simpler to start over, fixed in #2832.

PR #2830, opened 2025-10-07
Update Nissan Leaf Instrument Cluster for alignment

Sathans:
Update Nissan Leaf Instrument Cluster for alignment with default styling


Sathans:
This is important for Nissan Leaf Instrument Cluster users on Sunnypilot with its new vehicle selector. Currently we can't select Nissan Leaf Instrument Cluster manually and the fingerprint also awaits approval.

jyoung8607:
Your immediate issue should be resolved now that #2802 is merged. I understand what you were trying to accomplish with this additional change, but I don't think it's the right thing for upstream openpilot. We hope to spend some time improving fuzzy fingerprinting instead.

PR #2831, opened 2025-10-07
Nissan: Add 2024 Nissan Altima SL Ecu.engine fingerprint

gncnpk:


jyoung8607:
Historically, when encountering a gap in known supported model-years, we've allowed bridging one-year gaps but this is a very big jump from 2020 to 2024. It does all look to be sixth-gen according to Wikipedia, and the firmware is quite similar, but it's hard to be sure given there's a 2022 facelift in North America and this car isn't in America.

I've adjusted the CarDocs to leave 2021-2023 as unknown. Happy to fill that in as other people try it.

PR #2833, opened 2025-10-08
Lateral torque controller capnp: add derivative coefficient field in LateralTorqueTuning

felsager:
Add derivative coefficient field, to enable Openpilot's lateral torque controller to use error derivative compensation


haraschax:
LGTM

PR #2836, opened 2025-10-10
Honda: additional 2025 Accord Hybrid fingerprints

mvl-boston:
validation: 2c7d10e75bf841a1/00000000--25d9a008d2


mvl-boston:
Added validation from second driver

jyoung8607:
We're going to do these later, in a batch, as part of resolving #2632.

PR #2840, opened 2025-10-15
Replace approximation of gravity acceleration with standard value

roenthomas:
https://physics.nist.gov/cgi-bin/cuu/Value?gn


jyoung8607:
It's true that value is a common convention, and therefore more technically correct. However, the change would need to be matched in openpilot at a minimum, and probably various back-end model training and validation code as well. If this constant needs to change, it's probably best that the change comes from the comma autonomy team.

PR #2841, opened 2025-10-15
doco/website: Add footnote to VW-group and Tesla cars to prevent user confusion about where harness connects (option 1 - brand-agnostic footnote)

borhub:
Follow-up from https://github.com/commaai/openpilot/pull/36325

The Comma website is missing information for VW car users informing them that the J533 harness doesn't connect to the usual location on the windshield camera as advertised on the Comma website. This omission caused users to be confused and their Comma installation delayed when the harness connector didn't match the camera connector which they expected to plug into. People think they got the wrong connector, have to go to Discord etc.

This change fixes the customer experience by adding a footnote that directs the users to connect to the correct module under the dash. The copy of the note is generic so that it can potentially be reused for other cars whose harness doesn't plug into the windshield camera, starting with Tesla already in this PR.

---

The generic, brand-agnostic copy solution option here has the benefit of not creating too many footnotes in CARS.md, but is slightly less specific which may lead to continued user confusion queries in brand-specific Discords. But perhaps a lot of footnotes in CARS.md is a non-issue (no-one reads that file?), and having more specific instructions about the harness location on the website is more important (that is what the alternative option PR https://github.com/commaai/opendbc/pull/2842 does).

Thoughts?

### DIFF
--- a/opendbc/car/docs_definitions.py
+++ b/opendbc/car/docs_definitions.py
@@ -189,6 +189,10 @@ CarFootnote = namedtuple("CarFootnote", ["text", "column", "docs_only", "setup_n
 
 
 class CommonFootnote(Enum):
+  NON_CAMERA_HARNESS = CarFootnote(
+    "The harness used in this model is not a windshield camera harness but one that plugs" +
+    "into a car gateway box, usually located in the driver footwell area.",
+    Column.MODEL)
   EXP_LONG_AVAIL = CarFootnote(
     "openpilot Longitudinal Control (Alpha) is available behind a toggle; " +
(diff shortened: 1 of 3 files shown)


sshane:
The setup guide docs should just be updated to include an install video I think

borhub:
Replied in the related https://github.com/commaai/opendbc/pull/2842#issuecomment-3550051574 to keep the discussion in one place

sshane:
@jyoung8607 are you happy with this regarding https://github.com/commaai/opendbc/pull/2842#pullrequestreview-3483809921?

PR #2842, opened 2025-10-15
doco/website: Add footnote to VW-group cars to prevent user confusion about where harness connects (option 2 - brand-specific footnote)

borhub:
Follow-up from https://github.com/commaai/openpilot/pull/36325

The Comma website is missing information for VW car users informing them that the J533 harness doesn't connect to the usual location on the windshield camera as advertised on the Comma website. This omission caused users to be confused and their Comma installation delayed when the harness connector didn't match the camera connector which they expected to plug into. People think they got the wrong connector, have to go to Discord etc.

This change fixes the customer experience by adding a footnote that directs the users to connect to the correct module under the dash.

---

The brand-specific copy solution option here leads to creation of more footnotes in CARS.md as more footnotes are added into respective brands' `values.py` files, but it gives the customer a clearer guidance about where to plug the harness, likely reducing some Discord back-and-forths, consumers sending devices back etc. The CARS.md would end up with one footnote for each brand, but this might be a good tradeoff for getting greater end-customer clarity on the website.
The brand-agnostic, more generic copy alternative is in PR https://github.com/commaai/opendbc/pull/2841

### DIFF
--- a/opendbc/car/volkswagen/values.py
+++ b/opendbc/car/volkswagen/values.py
@@ -214,20 +214,31 @@ class Footnote(Enum):
     "Model-years 2022 and beyond may have a combined CAN gateway and BCM, which is supported by openpilot " +
     "in software, but doesn't yet have a harness available from the comma store.",
     Column.HARDWARE)
+  VW_J533_HARNESS_CONNECTION_LOCATION = CarFootnote(
+    "The J533 harness used in this model is not a windshield camera harness but one that plugs" +
+    "into the J533 gateway box, usually located in the driver footwell area.",
+    Column.MODEL)
 
 
 @dataclass
 class VWCarDocs(CarDocs):
   package: str = "Adaptive Cruise Control (ACC) & Lane Assist"
-  car_parts: CarParts = field(default_factory=CarParts.common([CarHarness.vw_j533]))
 
   def init_make(self, CP: structs.CarParams):
-    self.footnotes.append(Footnote.VW_EXP_LONG)
-    if "SKODA" in CP.carFingerprint:
-      self.footnotes.append(Footnote.SKODA_HEATED_WINDSHIELD)
+    harness = CarHarness.vw_j533          # Var used in preparation for multi-harness future (e.g. MEB platform)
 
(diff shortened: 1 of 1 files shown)


speedyconzales:
a good idea! thought of sending it back because of this. limited in tools, I am not able to install it currently. already spent time on this. honestly I wouldn't have bought it in the first place if I knew that beforehand. definitely needs to be added on the homepage

borhub:
@sshane @adeebshihadeh From a product owner lens, are you happy to merge this in and propagate the resulting CARS.md into the openpilot repo and then (the ultimate goal) to the website repo, so that the website has the additional footnote?

Note there's a 2nd variant of this PR that has a more generic implementation that could be used for multiple brands (but we'd need a more generic text) https://github.com/commaai/opendbc/pull/2841 I think either would improve the customer experience a lot and avoid surprises and confusion.

(Note: I haven't actually built the CARS.md as I don't know the process - being a Python newb - but static analysis seems happy, so hopefully the code will work as is)

sshane:
We have a few harnesses like this, if you think this disproportionally affects VW for some reason, we can merge this now. Just needs some clean up

borhub:
@sshane  It keeps coming up in the VW Discord so that's where I've encountered it, after my own experience (see also Speedygonzales' comment above).
As I don't have the visibility of how often the confusion comes up with other brands, I thought it's best to leave the tradeoff call up to you/Product folks/Adeeb about whether to go with a more brand-specific text and solution (this PR) or a more generic one but with also more generic wording (PR https://github.com/commaai/opendbc/pull/2841). The generic one is somewhat cleaner in that it doesn't multiply the footnotes as much and more reusable but slightly less specific about the location and name of the module.

Regarding your comment in the generic PR about adding a video, I agree and this could be _in addition_ to this PR. There's a great install video by Konik on Youtube we could use for VW cars: https://www.youtube.com/watch?v=__RwCHBcLVg (it shows the J533 module really well but also shows a non-Comma harness so not sure how sensitive you are to that. I previously mentioned this idea in my comment a month ago https://discordapp.com/channels/469524606043160576/534359517836607488/1427237094782140507  ).

I could add the video to this PR (if that's the pattern you prefer, i.e. brand-specific) together with the merge conflicts resolution if you're happy to have the video. Or if you prefer the generic solution, I could repurpose this PR to just add the video which would already independently help with mitigating the confusion.

PR #2845, opened 2025-10-18
Re-open: HKG: 2021 Kia Xceed PHEV

froddzei:
**From:** https://github.com/commaai/opendbc/pull/2708
_______________________________________________________________________________________________________________________
### Closed by accident

**Earlier comments:**
[jyoung8607](https://github.com/jyoung8607): _Kia Ceed and XCeed are indeed very close: https://en.wikipedia.org/wiki/Kia_Ceed#XCeed_

_We need to rework how the HYBRID flag is set for HKG in general first, otherwise we'll end up having to add a bunch of unnecessary platform permutations.__ (https://github.com/commaai/opendbc/pull/2708#pullrequestreview-3323517469)

[jyoung8607](https://github.com/jyoung8607): _Xceed will need an additional CarDocs entry at a minimum_ (https://github.com/commaai/opendbc/pull/2708#discussion_r2420205063)

[jyoung8607](https://github.com/jyoung8607): _We can't just add the HYBRID flag for all Ceed, that will break the non-hybrid cars. We could potentially add XCeed/Hybrid permutations, but ideally HYBRID should be auto-detected instead._ (https://github.com/commaai/opendbc/pull/2708#discussion_r2420210041)

_______________________________________________________________________________________________________________________

I added kia xceed fingerprint to the kia ceed port, since they have the same CD platform. everything works as expected, also added the "hybrid" flag in values.py in order to get it to work.

Used Harness B for my car.

Validation

Dongle ID: 22703002ddbe2a08
Route: 22703002ddbe2a08/0000000e--083b76c42c/0

if you need anything from me, tell me

### DIFF
--- a/opendbc/car/hyundai/fingerprints.py
+++ b/opendbc/car/hyundai/fingerprints.py
@@ -565,14 +565,17 @@ FW_VERSIONS = {
     (Ecu.fwdRadar, 0x7d0, None): [
       b'\xf1\x00CD__ SCC F-CUP      1.00 1.00 99110-J7500         ',
       b'\xf1\x00CD__ SCC F-CUP      1.00 1.02 99110-J7000         ',
+      b'\xf1\x00CDph SCC F-CUP      1.00 1.01 99110-CR100         ',
     ],
     (Ecu.eps, 0x7d4, None): [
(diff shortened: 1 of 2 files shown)


froddzei:
SO, What am i supposed to do to get help here?

royjr:
https://github.com/commaai/opendbc/pull/3144

PR #2847, opened 2025-10-18
HKG: Add FW for 2022 Hyundai Ioniq HEV

royjr:
Hyundai Ioniq Hybrid 2022 (non plug-in)
Hyundai H
4a975fc1d9e71801/00000005--50d176008a


rmozer1:
Yes, I have the Hyundai H harness.   I would love to test it out when the pull is available.  I'm missing something when I try to use putty to log into my device.  It really doesn't seem to matter what the user name is (comma, commai, or my rmozer). Forgive me if the question is stupid, or please point me to some information that I can read to get up to speed.
[
putty failure
](url)

royjr:
user confirmed successful fingerprint and drive

PR #2848, opened 2025-10-18
Honda - Bump crv to 2026

mvl-boston:
validation route: 178ba560c7b224e0/00000000--96e6dad8db


jyoung8607:
If you believe the fwdRadar and fwdCamera firmware are the same across ICE and hybrid CR-V (I haven't checked) you can go ahead and extend the model-years for hybrid. Otherwise, I'll merge as-is later this weekend.

mvl-boston:
Only ICE log I was sent was the 24 CRV and it did not match the 24 CRV Hybrid.

PR #2118, opened 2025-04-12
Honda Bosch: CAN FD Pilot 2023

vanillagorillaa:
**Checklist**
- [ ] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [ ] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: f9c43864cf057d05|2024-01-15--23-01-20
- [X] route with stock system: f9c43864cf057d05|2024-01-13--14-08-31
- [X] Harness : Bosch C

First port for a new CAN-FD Honda

DBC Notes: 
Using accord DBC, these messages are missing. Will have to check if they are on different ID's now
229 BOSCH_SUPPLEMENTAL_1
232 BRAKE_HOLD
316 GAS_PEDAL
490 VEHICLE_DYNAMICS
545 XXX_16
597 ROUGH_WHEEL_SPEED
892 CRUISE_PARAMS
927 RADAR_HUD
1029 DOOR_STATUS
13274 LKAS_HUD_A
13275 LKAS_HUD_B

PR #2852, opened 2025-10-21
Honda: Fix HONDA_PILOT specs

roenthomas:
In April, the original PR for the HONDA_PILOT_4G was merged.

Within that PR, a commit (https://github.com/commaai/opendbc/pull/2118/commits/7dbc832e536e7841c4a3ccb41e2b9b828246ffc1) changed HONDA_PILOT specs to be linked to HONDA_PILOT_4G, specifically, HONDA_PILOT_4G.specs

Then, in August, another commit was merged (https://github.com/commaai/opendbc/commit/d01d4a32a79b8a83c8fa5515f1d705840d033ab6#diff-17c61409dcbc81cf28dfe470a447966a238a5b641910726d04993c4a5b16debe) to updated the specs of HONDA_PILOT_4G. Since HONDA_PILOT specs are linked, its specs are also changed.

This seems odd that a car, which had long been discontinued, would have its specs changed.

This commit is to restore the HONDA_PILOT specs to what it originally was prior to the updates to HONDA_PILOT_4G, while retaining the new specs for HONDA_PILOT_4G

### DIFF
--- a/opendbc/car/honda/values.py
+++ b/opendbc/car/honda/values.py
@@ -337,7 +337,7 @@ class CAR(Platforms):
       HondaCarDocs("Honda Pilot 2016-22", min_steer_speed=12. * CV.MPH_TO_MS),
       HondaCarDocs("Honda Passport 2019-25", "All", min_steer_speed=12. * CV.MPH_TO_MS),
     ],
-    HONDA_PILOT_4G.specs,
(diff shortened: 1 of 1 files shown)


sshane:
The steering ratio is learned, and the initial value is not as important. Have you noticed any noticeable difference in steering? Going to close as this doesn't appear critical

PR #2856, opened 2025-10-25
2025 Subaru Crosstrek Lateral Support + Angle Based Subarus

jacobwaller:
Validation
* Dongle ID: 38b065e31c0a9ed7
* Route: 38b065e31c0a9ed7/00000001--5656806473

This corrects the existing support for angle based lkas subarus (Afaik, this would be the first fully supported one, so no breaking changes). But I've looked into it and these changes should allow easy porting of:
* 2024+ Crosstrek
* 2023 Ascent (#2217)
* 2022+ Outback 

They should only require a fingerprint + values added in values.py

### DIFF
--- a/opendbc/car/lateral.py
+++ b/opendbc/car/lateral.py
@@ -23,6 +23,23 @@ class AngleSteeringLimits:
   MAX_LATERAL_JERK: float = 0
   MAX_ANGLE_RATE: float = math.inf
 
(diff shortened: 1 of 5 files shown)


jacobwaller:
There is a failing test, which seems to be related to a route from some Acura not being available:
```
TestCarModel_1_ACURA_TLX_2G
```

Happy to remove that route, but figured it makes better sense to not touch stuff that isn't involved in this port

PR #2857, opened 2025-10-26
VW MQB: Update fingerprints.py for 2017 audi s3 with acc, without lkas

adnanklink:
Validation
* Dongle ID: fdc944a2993049de
* Route: fdc944a2993049de/00000003--bc88293668/0


adnanklink:
Been using a build with the fingerprints added for the last 2 weeks. Works exactly as expected!

PR #2859, opened 2025-10-27
Toyota: earlier high angle rate EPS fault prevention 

wtogami:
* 2021-2023 RAV4 Prime `TOYOTA_RAV4_PRIME` (logs below)
* 2022 RAV4 `TOYOTA_RAV4_TSS2_2022` (logs below)
* 2021 RAV4 `TOYOTA_RAV4_TSS2` (3 reports, no logs)
* 2021-2023 Sienna `TOYOTA_SIENNA_4TH_GEN` (2 reports, no logs)

These Toyota models can experience a rare EPS fault.

After a sharp, low-speed manual turn the human lets go of the wheel. During or shortly after the unwind the EPS faults. It can be difficult to reproduce.

[PR #24067](https://github.com/commaai/openpilot/pull/24067) "Toyota: remove 100°/sec steering lockout" attempted to prevent this high angle rate EPS fault by cutting torque if more than 18 frames. I found reducing this threshold by just one frame fixes the problem.

RAV4 Prime Routes
```
acd9ae4383134531/000000eb--b5fef978c7/488/503
acd9ae4383134531/000000d1--e4cb3cf682/484/498
acd9ae4383134531/000000d3--8495477bc1/567/584
```

RAV4 2022 Routes
```
8c3afba96b7a5db6/00000592--d968bfd1bf/11
timestamp: 19:22:40
EPS fault during manual right turn
acd9ae4383134531/00000155--d6bbd238ba/18
timestamp: 13:54:45
EPS fault during manual left turn
```


wtogami:
@sshane  back when you analyzed [PR #24067](https://github.com/commaai/openpilot/pull/24067) you had a lot more data to check against to find the ideal threshold. I'm curious if the Connect/Firehose corpus from recent months shows similar EPS faults happening on Toyotas at 18 frames?

wtogami:
I'm convinced this is correct. Totally solves the problem for my Toyota. But it can be exceedingly difficult to reproduce.

I think you folks have access to all the logs to be able to verify this EPS fault is happening to other Toyotas and what threshold they are experiencing?

Please let me know if you really need me to get logs from the other Toyota models who reported the same issue. It's been difficult to get them to cooperate because they don't currently understand Connect, are not subscribed so they lack history older than three days, or it's difficult to reproduce the fault. The combination of these factors has made it difficult to get logs from the other users.

wtogami:
Update: `TOYOTA_RAV4_TSS2_2022` now has logs of the same EPS fault

RAV4 2022 Routes
```
8c3afba96b7a5db6/00000592--d968bfd1bf/11
timestamp: 19:22:40
EPS fault during manual right turn
acd9ae4383134531/00000155--d6bbd238ba/18
timestamp: 13:54:45
EPS fault during manual left turn
```

sshane:
Looks like I found 21 to be the minimum in https://github.com/commaai/openpilot/pull/24067, but it's possible a few car variants fault slightly faster, `acd9ae4383134531/000000eb--b5fef978c7/488/503` was 20 frames. Bumping down by 1 seems safe to me in regards to control, it's not like openpilot torque control is very nuanced in high rate situations.

Though I don't believe the `card` split from `controlsd` should have added any latency by design since everything fault-prevention related is still in `card`. It blocks on reading `can`, then as fast as possible generates `carState` and sends it to `controlsd`, receives previous frame `carControl` from `controlsd`,  then runs carcontroller which has the counting logic contained. The actuation messages should be +10ms delayed after the split, but not the counting logic if I'm reading this correctly, though there may be additional delays from all the messages we send between can and sendcan.

PR #2864, opened 2025-10-30
2025 Subaru Crosstrek Lateral Support + Angle Based LKAS Support for Subarus

jacobwaller:
Validation
* Dongle ID:  38b065e31c0a9ed7
* Route: 38b065e31c0a9ed7/0000000b--eab0d07145

This corrects the existing support for angle based lkas subarus (Afaik, this would be the first fully supported one, so no breaking changes). But I've looked into it and these changes should allow easy porting of:

2024+ Crosstrek
2023+ Ascent (https://github.com/commaai/opendbc/pull/2217)
2022+ Outback
2025 Impreza
Those vehicles _should_ only require a fingerprint + values added in values.py


opendbc/car/subaru/interface.py
```diff
@@ -65,6 +68,10 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
     elif candidate == CAR.SUBARU_CROSSTREK_HYBRID:
       ret.steerActuatorDelay = 0.1
 
+    elif candidate == CAR.SUBARU_CROSSTREK_2025:
+      ret.dashcamOnly = is_release
```
adeebshihadeh: let's move this to the safety mode

opendbc/car/subaru/carstate.py
```diff
@@ -60,11 +60,17 @@ def update(self, can_parsers) -> structs.CarState:
     can_gear = int(cp_transmission.vl["Transmission"]["Gear"])
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
-    ret.steeringAngleDeg = cp.vl["Steering_Torque"]["Steering_Angle"]
+    if self.CP.flags & SubaruFlags.LKAS_ANGLE:
+      ret.steeringAngleDeg = cp.vl["Steering_2"]["Steering_Angle"]
```
adeebshihadeh: do we know if there's a common message for all subaru?

adeebshihadeh (commented):
Mostly LGTM, @sshane can you give the safety here a good review for a dashcam merge?

opendbc/car/subaru/carcontroller.py
```diff
@@ -32,30 +33,47 @@ def update(self, CC, CS, now_nanos):
...
-      # limits due to driver torque
+      if self.CP.flags & SubaruFlags.LKAS_ANGLE:
+        actual_steering_angle_deg = CS.out.steeringAngleDeg
+        desired_steering_angle_deg = actuators.steeringAngleDeg
+
+        apply_steer = apply_std_steer_angle_limits(
+          desired_steering_angle_deg,
+          self.apply_steer_last,
+          CS.out.vEgoRaw,
+          actual_steering_angle_deg,
```
sshane:
```suggestion
          CS.out.steeringAngleDeg,
```

easier to read to not have intermediary variables around here

opendbc/car/subaru/carcontroller.py
```diff
@@ -137,8 +155,11 @@ def update(self, CC, CS, now_nanos):
           can_sends.append(subarucan.create_es_static_2(self.packer))
 
     new_actuators = actuators.as_builder()
-    new_actuators.torque = self.apply_torque_last / self.p.STEER_MAX
-    new_actuators.torqueOutputCan = self.apply_torque_last
+    if self.CP.flags & SubaruFlags.LKAS_ANGLE:
+      new_actuators.steeringAngleDeg = self.apply_steer_last
+    else:
+      new_actuators.torque = self.apply_torque_last / self.p.STEER_MAX
+      new_actuators.torqueOutputCan = self.apply_torque_last
```
sshane: no need for this if statement, `apply_torque_last` or `apply_steer_last` will just be zero

opendbc/car/subaru/carcontroller.py
```diff
@@ -16,6 +16,7 @@ class CarController(CarControllerBase):
   def __init__(self, dbc_names, CP):
     super().__init__(dbc_names, CP)
     self.apply_torque_last = 0
+    self.apply_steer_last = 0
```
sshane:
```suggestion
    self.apply_angle_last = 0
```

matches other brands

opendbc/car/subaru/fingerprints.py
```diff
@@ -243,6 +243,20 @@
       b'\xd7!`p\x07',
       b'\xf4!`0\x07',
     ],
+  },
+    CAR.SUBARU_CROSSTREK_2025: {
```
sshane:
```suggestion
  CAR.SUBARU_CROSSTREK_2025: {
```

opendbc/car/subaru/interface.py
```diff
@@ -42,6 +42,12 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
     else:
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
+    if ret.flags & SubaruFlags.LKAS_ANGLE:
+      ret.steerActuatorDelay = 0.3
+      ret.safetyConfigs[0].safetyParam |= SubaruSafetyFlags.LKAS_ANGLE.value
+      ret.dashcamOnly = is_release
+
```
sshane:
```suggestion
```

opendbc/car/subaru/values.py
```diff
@@ -18,6 +19,12 @@ def __init__(self, CP):
     self.STEER_DRIVER_MULTIPLIER = 50  # weight driver torque heavily
     self.STEER_DRIVER_FACTOR = 1       # from dbc
 
+    self.ANGLE_LIMITS: AngleSteeringLimits = AngleSteeringLimits(
+        545,
+        ([0., 5., 35.], [5., .8, .15,]),
+        ([0., 5., 35.], [5., .8, .15,]),
```
sshane: two spaces for indent

opendbc/safety/modes/subaru.h
```diff
@@ -91,8 +102,18 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
 }
 
 static void subaru_rx_hook(const CANPacket_t *msg) {
-  const unsigned int alt_main_bus = subaru_gen2 ? SUBARU_ALT_BUS : SUBARU_MAIN_BUS;
+  const unsigned int alt_main_bus = (subaru_gen2 || subaru_lkas_angle) ? SUBARU_ALT_BUS : SUBARU_MAIN_BUS;
```
sshane: are these cars not gen 2?

opendbc/safety/modes/subaru.h
```diff
@@ -91,8 +102,18 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
...
 
+  #ifdef ALLOW_DEBUG
+  if (subaru_lkas_angle && (msg->addr == MSG_SUBARU_Steering_2) && (msg->bus == SUBARU_MAIN_BUS)) {
+    uint32_t raw = GET_BYTES(msg, 3, 3);
+    raw &= 0x1FFFFU;
+    int angle_meas_new = ROUND(to_signed(raw, 17) * -1);
```
sshane: you shouldn't need to round here, why?

opendbc/safety/modes/subaru.h
```diff
@@ -91,8 +102,18 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
 }
 
 static void subaru_rx_hook(const CANPacket_t *msg) {
-  const unsigned int alt_main_bus = subaru_gen2 ? SUBARU_ALT_BUS : SUBARU_MAIN_BUS;
+  const unsigned int alt_main_bus = (subaru_gen2 || subaru_lkas_angle) ? SUBARU_ALT_BUS : SUBARU_MAIN_BUS;
 
+  #ifdef ALLOW_DEBUG
+  if (subaru_lkas_angle && (msg->addr == MSG_SUBARU_Steering_2) && (msg->bus == SUBARU_MAIN_BUS)) {
```
sshane: I think you can remove ALLOW_DEBUG here, `subaru_lkas_angle` is already gated on it

opendbc/safety/modes/subaru.h
```diff
@@ -104,12 +125,27 @@ static void subaru_rx_hook(const CANPacket_t *msg) {
     angle_meas_new = ROUND(to_signed(angle_meas_new, 16) * -2.17);
     update_sample(&angle_meas, angle_meas_new);
   }
-
+  #ifdef ALLOW_DEBUG
+  }
+  #endif
```
sshane: Plus this looks messy, we don't do this for any other safety mode

opendbc/safety/modes/subaru.h
sshane: make this two if statements: if (subaru_lkas_angle) { if (msg==...

opendbc/safety/modes/subaru.h
```diff
@@ -91,8 +102,18 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
...
+  #ifdef ALLOW_DEBUG
+  if (subaru_lkas_angle && (msg->addr == MSG_SUBARU_Steering_2) && (msg->bus == SUBARU_MAIN_BUS)) {
+    uint32_t raw = GET_BYTES(msg, 3, 3);
+    raw &= 0x1FFFFU;
+    int angle_meas_new = ROUND(to_signed(raw, 17) * -1);
+    update_sample(&angle_meas, angle_meas_new);
+  }
+  
+  if (!subaru_lkas_angle) {
+  #endif // I don't like this, but I want to leave the following block completely unchanged until the comma team can get ahold of a subaru with LKAS_ANGLE
```
sshane: then make this else {

opendbc/safety/modes/subaru.h
```diff
@@ -104,12 +125,27 @@ static void subaru_rx_hook(const CANPacket_t *msg) {
...
     update_sample(&angle_meas, angle_meas_new);
   }
-
+  #ifdef ALLOW_DEBUG
+  }
+  #endif
+  
+
+  #ifdef ALLOW_DEBUG
+  if (subaru_lkas_angle) {
```
sshane: Same, remove `ALLOW_DEBUG` here and below

opendbc/safety/modes/subaru.h
```diff
@@ -233,17 +306,26 @@ static safety_config subaru_init(uint16_t param) {
...
+  const uint16_t SUBARU_PARAM_GEN2 = 1;
   subaru_gen2 = GET_FLAG(param, SUBARU_PARAM_GEN2);
-
+  
 #ifdef ALLOW_DEBUG
   const uint16_t SUBARU_PARAM_LONGITUDINAL = 2;
   subaru_longitudinal = GET_FLAG(param, SUBARU_PARAM_LONGITUDINAL);
+
+  const uint16_t SUBARU_PARAM_LKAS_ANGLE = 8;
+  subaru_lkas_angle = GET_FLAG(param, SUBARU_PARAM_LKAS_ANGLE);
```
sshane: this is what gates subaru_lkas_angle for everything above

opendbc/car/subaru/carcontroller.py
jacobwaller: resolved locally

opendbc/car/subaru/carcontroller.py
jacobwaller: resolved locally

opendbc/car/subaru/fingerprints.py
jacobwaller: resolved locally

opendbc/car/subaru/values.py
jacobwaller: resolved locally

opendbc/safety/modes/subaru.h
jacobwaller: They are -- good catch. Resolved locally

opendbc/car/subaru/carstate.py
jacobwaller: @adeebshihadeh From what I can see (by just using the new message and re-running the tests), it seems like it is common for all Subarus *except* preglobal.

A few options:
1. Leave it how it is in the PR
2. Change the condition to `if not self.CP.flags & SubaruFlags.PREGLOBAL`
3. Make it a one-liner:
`ret.steeringAngleDeg = cp.vl["Steering_2"]["Steering_Angle"] if not self.CP.flags & SubaruFlags.PREGLOBAL else cp.vl["Steering_Torque"]["Steering_Angle"]`

Do you have a preference between those?

opendbc/car/subaru/interface.py
```diff
@@ -42,6 +42,11 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
     else:
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
+    if ret.flags & SubaruFlags.LKAS_ANGLE:
+      ret.steerActuatorDelay = 0.3
+      ret.safetyConfigs[0].safetyParam |= SubaruSafetyFlags.LKAS_ANGLE.value
+      ret.dashcamOnly = is_release
```
jacobwaller: @adeebshihadeh is this what you meant by
> let's move this to the safety mode

comment from [here](https://github.com/commaai/opendbc/pull/2864#discussion_r2607768704)

opendbc/safety/modes/subaru.h
```diff
@@ -93,22 +104,39 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
 static void subaru_rx_hook(const CANPacket_t *msg) {
   const unsigned int alt_main_bus = subaru_gen2 ? SUBARU_ALT_BUS : SUBARU_MAIN_BUS;
 
-  if ((msg->addr == MSG_SUBARU_Steering_Torque) && (msg->bus == SUBARU_MAIN_BUS)) {
```
sshane: Which cars can we not use Steering_Torque? The angle meas mutation test is failing because we only test angle meas for angle LKAS cars you added here, not the unused message that Justin added originally.

jacobwaller: None as far as I know, all the angle based cars I've tested have been able to use `Steering_2`

We can probably remove sourcing angle_meas from `Steering_Torque` if it's not needed for torque based cars

opendbc/car/subaru/carstate.py
sshane: What about `Steering`? 
image

jacobwaller: @sshane I believe I initiallytried `Steering` and it led to an extremely strong right-hugging tendency that was fixed when I switched to Steering_2. It's weirdly off from Steering_2

Similar to the issue Sunny had here:
https://discord.com/channels/469524606043160576/525718620517564446/1423050384343175229

sshane: Hmm, well the current Steering_Torque starts at 40 degrees offset on pre-global Outback:

image

And Steering is also generally lower:

image

---

Do we have any user with a pre-global to try and get a route where they turn the wheel to 90, 180, 270, 360, etc?

sshane:
@elkoled if the diff is exceptionally long, can you add a `<detail>` tag to hide it until expanded?

opendbc/car/subaru/carstate.py
jacobwaller: Yeah we switched from the Steering_Torque because it's 0 on my car and offset a crazy amount on the 2024 crosstrek :)

I'm not sure on pre-global routes, but shouldn't all pre-globals be torque based and therefore wouldn't need angle_meas? Might be mistaken there

sshane: `Steering` matches on a few gen2 cars I've checked, assuming it works on gen1 non-pre-global we can use either `Steering` (100Hz), `Steering_Torque` (50Hz), or `Steering_2` (50Hz). The angle signal from `Brake_Pressure_L_R` (likely the TCS module) seems offset and generally lags behind when wheel moves, so shouldn't use that.

However after second glance on Subaru Outback 2020, the higher freq `Steering` signal has less precision even though it's technically broadcast at a higher rate. Few cases below:

image

image

---

If we can prove `Steering_2` is valid on all global cars (gen1 and gen2 + hybrid and angle LKAS), we can switch to it when not-preglobal. That can be a separate PR we can merge first. I would prefer 2 angle signals over 3 any day

sshane: > Yeah we switched from the Steering_Torque because it's 0 on my car and offset a crazy amount on the 2024 crosstrek :)
> 
> I'm not sure on pre-global routes, but shouldn't all pre-globals be torque based and therefore wouldn't need angle_meas? Might be mistaken there

After this, pre-global and global would use `Steering_Torque`, and angle global would use `Steering_Torque_2`. I think it makes more sense to have pre-global use `Steering_Torque` and global and angle global to both use `Steering_Torque_2`. They look pretty much identical on a global Outback 2020. It's just the grouping of which cars use which signals for clarity/simplicity's sake

jacobwaller: Are you using `subaru_global_2017_generated` for the DBC you're referring to?

Steering_Torque_2 doesn't populate anything with an angle for me on my route (`38b065e31c0a9ed7/0000000b--eab0d07145`)

sshane: Sorry meant `Steering_2`

opendbc/safety/modes/subaru.h
sshane: I'll have to see why @jnewb1 used that signal if it is zero on your car. Maybe the existing angle cars are different

sshane: Okay the three angle LKAS cars Justin merged all have valid steering angles on our existing message, so that makes sense. 

Subaru Forester 2022 ( which is not gen 2?!) looks good. Idk what's going on with the noise, but `Steering_2` matches perfectly here:

image

---

`Steering_2` also looks good on Outback 2023:

image


---

`Steering_2` also looks good on Ascent 2023:

image

opendbc/car/subaru/carstate.py
sshane: Merged in https://github.com/commaai/opendbc/pull/3094

opendbc/car/subaru/interface.py
sshane: I think he means add a ifdef to safety for release builds

opendbc/car/subaru/carcontroller.py
```diff
@@ -32,30 +33,44 @@ def update(self, CC, CS, now_nanos):
 
     # *** steering ***
     if (self.frame % self.p.STEER_STEP) == 0:
-      apply_torque = int(round(actuators.torque * self.p.STEER_MAX))
-
-      # limits due to driver torque
+      if self.CP.flags & SubaruFlags.LKAS_ANGLE:
```
sshane, line 65: should explore splitting this out into method to see if it simplifies the main update function

sshane:
@jacobwaller I think we can merge a PR that fixes the current angle cars by adding the new angle measurement + tests + angle safety, then we can just add your fingerprints. Wanna split that out and keep this PR as is for now? It's pretty much just this but without adding your platform

opendbc/safety/modes/subaru.h
```diff
@@ -93,17 +104,34 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
...
-    torque_driver_new = -1 * to_signed(torque_driver_new, 11);
-    update_sample(&torque_driver, torque_driver_new);
+  if (subaru_lkas_angle) {
+    if ((msg->addr == MSG_SUBARU_Steering_2) && (msg->bus == SUBARU_MAIN_BUS)) {
+      uint32_t raw = GET_BYTES(msg, 3, 3);
+      raw &= 0x1FFFFU;
+      int angle_meas_new = -to_signed(raw, 17);
+      update_sample(&angle_meas, angle_meas_new);
+    }
+  } else {
```
sshane: I think we can always measure driver torque, it just won't be used (should check if it's zero or not on these cars though)

jacobwaller: Yeah, it's non-zero on my car. I pulled that out of the else

opendbc/car/subaru/carcontroller.py
jacobwaller: Done, I did the same for the torque-based steering. lmk what you think

opendbc/safety/modes/subaru.h
```diff
@@ -35,9 +37,9 @@
 
 #define MSG_SUBARU_ES_UDS_Request        0x787U
 
-#define MSG_SUBARU_ES_HighBeamAssist     0x121U
-#define MSG_SUBARU_ES_STATIC_1           0x22aU
-#define MSG_SUBARU_ES_STATIC_2           0x325U
+#define MSG_SUBARU_ES_HighBeamAssist     0x22AU
+#define MSG_SUBARU_ES_STATIC_1           0x325U
+#define MSG_SUBARU_ES_STATIC_2           0x121U
```
jacobwaller: Insanely, I think these were wrong previously, but we never actually checked them in the tests

Image

Image

opendbc/safety/modes/subaru.h
```diff
@@ -93,17 +105,34 @@ static uint32_t subaru_compute_checksum(const CANPacket_t *msg) {
...
     torque_driver_new = -1 * to_signed(torque_driver_new, 11);
     update_sample(&torque_driver, torque_driver_new);
   }
 
-  // enter controls on rising edge of ACC, exit controls on ACC off
-  if ((msg->addr == MSG_SUBARU_CruiseControl) && (msg->bus == alt_main_bus)) {
-    bool cruise_engaged = (msg->data[5] >> 1) & 1U;
-    pcm_cruise_check(cruise_engaged);
+  if (subaru_lkas_angle) {
+    if ((msg->addr == MSG_SUBARU_ES_Brake) && (msg->bus == SUBARU_ALT_BUS)) {
```
jacobwaller: @sshane you had this as CAM, but I changed to alt because it's on bus 1 in my car

opendbc/safety/tests/test_subaru.py
```diff
@@ -25,9 +25,9 @@ class SubaruMsg(enum.IntEnum):
   ES_LKAS_State     = 0x322
   ES_Infotainment   = 0x323
   ES_UDS_Request    = 0x787
-  ES_HighBeamAssist = 0x121
-  ES_STATIC_1       = 0x22a
-  ES_STATIC_2       = 0x325
+  ES_HighBeamAssist = 0x22A
+  ES_STATIC_1       = 0x325
+  ES_STATIC_2       = 0x121
```
jacobwaller: same as https://github.com/commaai/opendbc/pull/2864#discussion_r2749835146

opendbc/safety/modes/subaru.h
sshane: This is great as a separate PR. Yeah IDK why we merged so much unused code :p

HerbBoy:
Hi - is there plans to implement this in the near future? and if so what harness is required for a crosstrek wilderness 2024?

qinghao1:
Hey @sshane what needs to be done to merge this PR? Myself and many others are eagerly awaiting support for the latest Subaru cars

NylonDiamond:
Is this still getting merged? Coming up on 1 year in limbo..

PR #2886, opened 2025-11-07
Honda: Disable more CI test routes

jyoung8607:
Pending commaai/openpilot#36098


jyoung8607:
There's something more going on here than expiration, I spot-checked a couple and I can still get the rlogs manually. Also, HRV was failing at one point and now it isn't. It would be nice if we could restore the existing test routes for CI use, but will have to look into it later.

PR #2888, opened 2025-11-07
can: cache DBC parsing to speed up car interface tests

logesh45:
## Performance
Cuts `test_car_interfaces.py` runtime from ~38.84s to ~15.98s (≈2.4× faster). The slowest test drops from ~6.52s to ~0.81s (8× faster), meeting the target of ≤0.2s avg and <1s max per car.

## Changes
- **DBC caching**: Module-level cache with explicit [get_dbc()](cci:1://file:///Users/logeshr/Documents/VSCodeProjects/opendbc/opendbc/can/dbc.py:89:0-101:29) factory function. Each DBC file is parsed only once per process, with subsequent calls reusing the cached instance.
- **CANParser fast path**: Early-return in [CANParser.update](cci:1://file:///Users/logeshr/Documents/VSCodeProjects/opendbc/opendbc/can/parser.py:215:2-255:24) when called with an empty frame list, avoiding needless buffer clearing.

## Verification
Ran the test suite before and after the change:
```bash
pytest --durations=10 selfdrive/car/tests/test_car_interfaces.py
```

Before: 

image

After:
image

### DIFF
--- a/opendbc/can/dbc.py
+++ b/opendbc/can/dbc.py
@@ -1,7 +1,8 @@
 import re
 import os
-from dataclasses import dataclass
 from collections.abc import Callable
+from dataclasses import dataclass
+from functools import cache
 
 from opendbc import DBC_PATH
 
@@ -73,14 +74,8 @@ VAL_RE = re.compile(r"^VAL_ (\w+) (\w+) (.*);")
 VAL_SPLIT_RE = re.compile(r'["]+')
 
 
-@dataclass
+@cache
 class DBC:
-  name: str
-  msgs: dict[int, Msg]
-  addr_to_msg: dict[int, Msg]
-  name_to_msg: dict[str, Msg]
-  vals: list[Val]
-
   def __init__(self, name: str):
     dbc_path = name
     if not os.path.exists(dbc_path):


opendbc/can/dbc.py
```diff
@@ -79,12 +80,41 @@ class DBC:
...
+    if cached is not None:
+      return cached
+
+    self = super().__new__(cls)
+    self._dbc_path = dbc_path
+    return self
+
+  def __init__(self, name: str):
+    if getattr(self, "_initialized", False):
+      return
```
KRRT7: I'd avoid implementing the singleton pattern in this manner as it can have hidden side effects and makes reasoning about the logic a bit harder; you can read this article on why singletons are a bad idea: https://nedbatchelder.com/blog/202204/singleton_is_a_bad_idea.html

KRRT7 (changes requested):

opendbc/can/dbc.py
KRRT7: I opened a PR with a cleaner implementation, hope you don't mind

logesh45: Let me make these changes.

KRRT7: Feel free to merge my PR

TheSecurityDev: @KRRT7 I don't see your PR, where is it?

KRRT7: I opened it against their fork & branch https://github.com/logesh45/opendbc/pull/1

logesh45: @KRRT7 Thank you for your inputs. I refactored further to use module level factory to keep DBC a pure data class and make Cache independent and easier to test.

logesh45:
@maxime-desroches This is the PR I was talking about today. I cant run the main openpilot repo GitHub tests with these changes because this is a submodule.

opendbc/can/parser.py
```diff
@@ -214,6 +214,9 @@ def can_valid(self) -> bool:
     return self.can_invalid_cnt < CAN_INVALID_CNT and counters_valid
 
   def update(self, strings, sendcan: bool = False):
+    if not strings:
+      return set()
+
```
jyoung8607: Why this change? Is it necessary?

jyoung8607 (approved):
This is a meaningful improvement! I started looking at the same problem. I'm not sure this is the best or prettiest way to solve it, but it seems to be the simplest.

@adeebshihadeh This sidesteps a great deal of duplicate parsing work. Right now for every single car port we have to parse the full text of the same DBC file several times: two or three CANParsers, a CANDefine, and a CANPacker or two. That's a problem for ports with big DBCs, like Ford with a 12,700 line DBC. Then multiply all that for hypothesis.

Before, on my workstation, pretty consistent:
```
(opendbc) jyoung@jy-workstation-ubuntu:~/openpilot/opendbc_repo$ pytest -n1 opendbc/car/tests/test_car_interfaces.py -k test_car_interfaces[FORD_MAVERICK_MK1]
Cannot read termcap database;
using dumb terminal settings.
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.12.0, pytest-8.4.2, pluggy-1.6.0
Using --randomly-seed=2812254386
rootdir: /home/jyoung/openpilot/opendbc_repo
configfile: pyproject.toml
plugins: randomly-4.0.1, cov-7.0.0, xdist-3.7.1.dev24+g2b4372b, subtests-0.15.0, mock-3.15.1, hypothesis-6.47.5
initialized: 1/1 workerCannot read termcap database;
using dumb terminal settings.
1 worker [1 item]      
.                                                                                                                                                                                [100%]
================================================================================= slowest 10 durations =================================================================================
1.01s call     opendbc/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces[FORD_MAVERICK_MK1]

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================================== 1 passed in 1.59s ===================================================================================
```

After, on my workstation, also pretty consistent:
```
(opendbc) jyoung@jy-workstation-ubuntu:~/openpilot/opendbc_repo$ pytest -n1 opendbc/car/tests/test_car_interfaces.py -k test_car_interfaces[FORD_MAVERICK_MK1]
Cannot read termcap database;
using dumb terminal settings.
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.12.0, pytest-8.4.2, pluggy-1.6.0
Using --randomly-seed=133558739
rootdir: /home/jyoung/openpilot/opendbc_repo
configfile: pyproject.toml
plugins: randomly-4.0.1, cov-7.0.0, xdist-3.7.1.dev24+g2b4372b, subtests-0.15.0, mock-3.15.1, hypothesis-6.47.5
initialized: 1/1 workerCannot read termcap database;
using dumb terminal settings.
1 worker [1 item]      
.                                                                                                                                                                                [100%]
================================================================================= slowest 10 durations =================================================================================
0.14s call     opendbc/car/tests/test_car_interfaces.py::TestCarInterfaces::test_car_interfaces[FORD_MAVERICK_MK1]

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================================================== 1 passed in 0.68s ===================================================================================

```

opendbc/can/parser.py
logesh45: The change handles the case where strings could be `None`, preventing a potential `TypeError` when trying to check `strings[0]` on line 220. This makes the code more robust against unexpected inputs.

Current tests pass empty lists, other code paths might pass `None` at a later time, and this prevents runtime errors

sshane: let's type hint strings to make it clear then

sshane:
You can just decorate with functools.cache, no? https://github.com/commaai/opendbc/pull/2922

opendbc/can/parser.py
sshane: and this is a separate, unrelated change to this PR? you can open a new PR for it

sshane:
you usually get better results if you first try to simplify, then optimize! Firstly, we weren't using any of the benefits of dataclass here, re-implementing `__init__`.  I also think `DBC.from_file` which would have handled the memoization in 5 lines or so would have been a better pattern than making a function separate to the class. See how https://github.com/commaai/openpilot/blob/85a162dd4320f14cb8aba1bcd7ff427d9b753ce5/system/ui/lib/wifi_manager.py#L80-L93 works since it does some preprocessing before storing in the frozen dataclass

But to make it super simple, just going to memoize it like this for now

PR #2889, opened 2025-11-07
development template car port

jyoung8607:


jyoung8607:
This was a valuable exercise, but I'm not going to merge its current form. What's needed is a construct that enables tests when working with the template car port, for offline route/test-driven development, but in a form that passes tests when not using it for development AND doesn't bit-rot during opendbc refactors. This PR can still be used as a reference.

PR #2897, opened 2025-11-10
Skoda: rename Škoda to Skoda to fix the supported vehicle search on the new website

borhub:
It is practically impossible to search for the brand name of Skoda cars on the new website due to most users not having their keyboards set up to type [š/Š]koda. Nothing comes up in the new typeahead-based search when a user types in "skoda".

This PR changes the name of the cars to the more widely used spelling without the accent on the S, so that (once propagated to the CARS.md file and to the OP and website repos) results come up when searching by the brand name.

I am a native Czech speaker (where Skoda is from) and even we regularly don't use the accented form of characters such as the š/Š when typing emails, using Google search etc., and even Czech websites are always set up to work with the unaccented words in their search forms. So this aligns with that style.

### DIFF
--- a/opendbc/car/volkswagen/values.py
+++ b/opendbc/car/volkswagen/values.py
@@ -203,7 +203,7 @@ class Footnote(Enum):
     "Refers only to the MQB-based European B8 Passat, not the NMS Passat in the USA/China/Mideast markets.",
     Column.MODEL)
   SKODA_HEATED_WINDSHIELD = CarFootnote(
-    "Some Škoda vehicles are equipped with heated windshields, which are known " +
+    "Some Skoda vehicles are equipped with heated windshields, which are known " +
     "to block GPS signal needed for some comma 3X functionality.",
     Column.MODEL)
   VW_EXP_LONG = CarFootnote(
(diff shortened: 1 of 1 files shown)


borhub:
While we're here, would appreciate if you could look at:
https://github.com/commaai/opendbc/pull/2842
and/or the more generic alternative https://github.com/commaai/opendbc/pull/2841
which I think would also help with the customer experience. The new website continues to give an impression that the VW J533 harness is a rearview mirror cam harness, making it a disappointing setup experience for new users who may think they got the wrong harness and won't be able to drive in 15 minutes or so (when they actually can). Goodwill is hard to win back :)

sshane:
should fix on website

borhub:
@sshane @jyoung8607 Hope you don't mind the pushback but if I were you or Adeeb I wouldn't bother adding special handling on the website for this, Imho that's unnecessary complexity and against the core Comma principle to keep things simple.

As you can see in the text of the 3 reviews from the UK and Australia below, everyone in the English-speaking real world refers to the cars as Skoda without the accent over the s.
https://www.carwow.co.uk/skoda/superb
https://www.carbuyer.co.uk/skoda/superb
https://www.carexpert.com.au/car-reviews/2025-skoda-superb-review

German:
https://www.autobild.de/artikel/skoda-superb-gebrauchtwagen-test-13897291.html

[Edit: I see the Vehicles page is now back and it replaced the typeahead search, which makes this change less immediately valuable, but imho still worth considering due to the simplification and avoidance of potential future issues with search]

jyoung8607:
I agree that search for "skoda" should find the car. It looks like that works in the current version of the comma shop, and works when searching docs/CARS.md in Chrome. Subject to that, I'd like to keep the official brand name intact.

PR #2899, opened 2025-11-11
Update to new four car parts

sshane:
look at the simplicity!


opendbc/car/volkswagen/values.py
```diff
@@ -235,7 +235,7 @@ class Footnote(Enum):
     Column.MODEL)
   SKODA_HEATED_WINDSHIELD = CarFootnote(
     "Some Škoda vehicles are equipped with heated windshields, which are known " +
-    "to block GPS signal needed for some comma 3X functionality.",
+    "to block GPS signal needed for some comma four functionality.",
```
sshane, line 237: @adeebshihadeh have any suspicion this is different now?

adeebshihadeh: no

PR #2902, opened 2025-11-13
Tesla: Fix Autopark

AmyJeanes:
Validation
* Dongle ID: 4b3cbd6038d07ca6
* Route: TBD

## TODO: Add route(s)
- Need to test:
  - [ ] Summon from off-road
  - [ ] Summon from on-road
  - [ ] Autopark
  - [ ] Stock LKAS
  - [ ] Stock emergency LKAS
  - [ ] UI behaviour when engaging autopark
    - [ ] Should not allow openpilot to engage
    - [ ] Should not show any errors or warnings (only generic unavailable one if attempting to engage)

TODO: Consider adding new event / car state for autoparking

### Fixed Autopark failing in Tesla vehicles

Note that Autopark is not normally accessible with openpilot is enabled because it requires TACC to be set instead of Autopilot/FSD which disables autopark for some reason in the car - so for testing purposes in the routes `invalidLkasSetting` was forcibly disabled but this change is not included in this PR as that is a separate change and comes with some interesting implications so keeping them separate. That will come in a future PR.

Previously in the code we had `autopark` defined from the `DI_autoparkState` message, which as it turned out after extensive testing, is actually not used by autopark at all but instead summon / smart summon - so I have renamed related variables to reflect this. I also clarified in a comment that it only works when the car was off-road when starting summon, which is the current behaviour in master not new in my branch - and added a TODO to fix this later.

We use the new `ACTIVE_AUTOPARK` state in the `DAS_autopilotState` message to determine whether or not autopark is active, and, like summon, we block messages and enable pass-through when it is. We also ensure that `cruiseState.enabled = false` in the car state also matches this new behaviour.

### DIFF
--- a/opendbc/safety/modes/tesla.h
+++ b/opendbc/safety/modes/tesla.h
@@ -19,6 +19,7 @@ static bool tesla_summon_prev = false;
 // Autopark
 static bool tesla_autopark = false;
 static bool tesla_autopark_prev = false;
+static bool tesla_autopark_blocked = false;
 
 static uint8_t tesla_get_counter(const CANPacket_t *msg) {
 
@@ -38,6 +39,9 @@ static uint8_t tesla_get_counter(const CANPacket_t *msg) {
   } else if (msg->addr == 0x370U) {
     // Signal: EPAS3S_sysStatusCounter
     cnt = msg->data[6] & 0x0FU;
+  } else if (msg->addr == 0x39bU) {
+    // Signal: DAS_statusCounter
+    cnt = msg->data[6] >> 4;
   } else {
   }
   return cnt;
@@ -51,6 +55,9 @@ static int _tesla_get_checksum_byte(const int addr) {
   } else if (addr == 0x488) {
     // Signal: DAS_steeringControlChecksum
     checksum_byte = 3;
+  } else if (addr == 0x39b) {
+    // Signal: DAS_statusChecksum
+    checksum_byte = 7;
   } else if ((addr == 0x257) || (addr == 0x118) || (addr == 0x39d) || (addr == 0x286) || (addr == 0x311)) {
     // Signal: DI_speedChecksum, DI_systemStatusChecksum, IBST_statusChecksum, DI_locStatusChecksum, UI_warningChecksum
     checksum_byte = 0;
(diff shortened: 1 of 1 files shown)


sshane:
I recall testing Autopark with Autopilot/FSD on and it worked, does it not? I also recall both systems using similar APIs, so I just picked one when naming the variables

AmyJeanes:
@sshane from what I saw in my testing and cabana scanning the "autopark" as currently implemented in opendbc is actually only relevant to Summon and Smart Summon. Actual Autopark now uses DAS_autopilotState ACTIVE_AUTOPARK (maybe new since vision based autopark was introduced?)

This PR is a (currently untested) "proper" implementation of it, but I am dailying these changes over in my fork at the moment until I get round to doing this properly: https://github.com/sunnypilot/sunnypilot/commit/78bc354af9c9697e7d342d5a1ce868bda0900b13

It's a little bit of a hack in there, but it does work properly, the only weird bit really is that it gives a invalid lkas mode warning on screen while autoparking.

If you've got any feedback for me on anything that I can use while doing this fix properly let me know I will keep it in mind!

One of the challenges I think will be that some newer cars don't have the double press for TACC vs AP so they will be effectively unable to use fixed autopark unless we make OP work with AP engaged, curious if you have any thoughts around that situation.

opendbc/car/tesla/carstate.py
```diff
@@ -69,12 +79,17 @@ def update(self, can_parsers) -> structs.CarState:
     cruise_state = self.can_define.dv["DI_state"]["DI_cruiseState"].get(int(cp_party.vl["DI_state"]["DI_cruiseState"]), None)
     speed_units = self.can_define.dv["DI_state"]["DI_speedUnits"].get(int(cp_party.vl["DI_state"]["DI_speedUnits"]), None)
 
-    autopark_state = self.can_define.dv["DI_state"]["DI_autoparkState"].get(int(cp_party.vl["DI_state"]["DI_autoparkState"]), None)
+    # DI_autoparkState is used by Summon, not autopark (which uses DAS_autopilotState = ACTIVE_AUTOPARK)
```
dzid26: So `DI_autoparkState` and known enums probably should be then renamed in the DBC to reflect that it refers to summon.

AmyJeanes: No longer using this signal, it reported too low (2Hz) for safe use anyway.

PR #2457, opened 2025-07-11
Honda: Gear position message refactor

jyoung8607:
The Honda gearbox message situation needed some help. Four different messages defined with an array of overlapping and conflicting names, incorrect message frequencies, different fields used from the same message in different DBCs, different signal bit widths and endianness settings, several ways to pick a message, lots of duplication. Other than that it was great.

This PR migrates the Honda port to gearbox message auto-detection with a common gearbox DBC. Virtually all cars are covered by two messages: one for traditional CVTs, and another for traditional autos, eCVTs and direct-drive EVs. The oldest ACURA_RDX is a snowflake that has to stay in its custom DBC.

**Validation**

Verified against several data sources using commaai/openpilot#35682:
* commaCarSegments v2
* upstream comma CI test routes
* pending CI test routes from the Q3 CY2025 Honda car port backlog

**TODO**

- [x] Add back a manual trans reverse signal, don't think the current signal is good
- [x] Final validation
- [x] Monitor that auto-detect works in the real world, that we don't have startup timing issues

PR #2933, opened 2025-12-07
Honda: 2015 CRV South Africa

mvl-boston:
Car has different (legacy) gearbox, from USA and EU CRV.  Otherwise aligned with code structure for EU CRV.

Testroute: 4991f6e1afbe5adb/00000073--d04999333c


jyoung8607 (commented):
At the time I consolidated gearbox messages (#2457) Acura RDX was the only weird outlier, and leaving the DBC separate was an expedient hack.

Now that we're merging more cars that reference that legacy message (and I think there were others coming?) I want to revisit that abstraction. That way we get rid of some more openpilot car model sprawl, collapsing the UDS firmware for this (and HONDA_CRV_EU) into HONDA_CRV.

I can work on that soon, leave this open in the meantime.

mvl-boston:
I believe I did that in the serial steering Accord PR if you want to grab those commits.

mvl-boston:
I think I get what you are saying: a 3 step autodetect in interface and merging of dbc.

That makes more sense.  Let me know if you prefer I draft one.

jyoung8607:
@mvl-boston the test route doesn't have uploaded rlogs (legitimately, I checked). CI was passing before due to a separate bug which is now fixed. Otherwise LGTM and will merge as soon as we have a test route with rlogs.

mvl-boston:
> @mvl-boston the test route doesn't have uploaded rlogs (legitimately, I checked). CI was passing before due to a separate bug which is now fixed. Otherwise LGTM and will merge as soon as we have a test route with rlogs.

New log added.

PR #2936, opened 2025-12-15
PSA: add FW query

elkoled:
This PR adds a minimal FW query for the ABS ECU which works on all tested PSA models.
Additionally, a second fingerprint is added that covers these models:
- Peugeot 508
- DS DS7
- Citroen Berlingo

Fingerprint distinction:
**PSA_PEUGEOT_208:** STEERING_ALT msg on BUS 0
**PSA_PEUGEOT_508:** STEERING_ALT msg on BUS 2


adeebshihadeh:
@sshane is this safe to merge or do we need to test on other brands?

opendbc/car/psa/carstate.py
```diff
@@ -33,8 +33,13 @@ def update(self, can_parsers) -> structs.CarState:
...
     # steering wheel
-    ret.steeringAngleDeg = cp.vl['STEERING_ALT']['ANGLE'] # EPS
-    ret.steeringRateDeg = cp.vl['STEERING_ALT']['RATE'] * (2 * cp.vl['STEERING_ALT']['RATE_SIGN'] - 1) # convert [0,1] to [-1,1] EPS: rot. speed * rot. sign
+    STEERING_ALT_BUS = {
+      CAR.PSA_PEUGEOT_208: cp.vl,
+      CAR.PSA_PEUGEOT_508: cp_cam.vl,
+    }
+    bus = STEERING_ALT_BUS[self.CP.carFingerprint]
+    ret.steeringAngleDeg = bus['STEERING_ALT']['ANGLE'] # EPS
+    ret.steeringRateDeg  = bus['STEERING_ALT']['RATE'] * (1 - 2 * bus['STEERING_ALT']['RATE_SIGN']) # convert [0,1] to [1,-1] EPS: rot. speed * rot. sign
```
sshane, line 42: Why is this in this PR?

sshane:
Should be safe, tx addr `0x6ad` is unique, and rx addr `0x68d` is also unique. The fingerprinting first queries for existing ECUs to do the full FW query

opendbc/car/psa/carstate.py
elkoled: This is required for the new PSA_PEUGEOT_508 models to work.
I thought it was related enough to include in this PR.

PR #2940, opened 2025-12-15
Revert "Tesla: Add fingerprint for Model Y"

sshane:
Reverts commaai/opendbc#2924


sshane:
@jyoung8607 see https://github.com/commaai/opendbc/pull/2934

PR #2946, opened 2025-12-18
Tesla: fuzzy fingerprinting

### LINKED ISSUE #1917: [$500 bounty] Tesla: fuzzy fingerprinting

rosacry:
Closes #1917

## Summary
Implements fuzzy fingerprinting for Tesla vehicles (Model 3, Model Y, Model X) by parsing the model code from EPS firmware versions.

## Implementation Details
- Extracts model code from Tesla EPS firmware format: `<project>,<model_code><version>`
- Maps first character of model code to platform:
  - `E*` → Model 3
  - `Y*` → Model Y
  - `X*` → Model X
- Uses regex pattern for robust parsing (handles garbage input gracefully)

### Has Tesla ever changed the APIs we use?
**Yes.** Tesla has changed CAN-level control behavior in the past.  
For example, FSD 14 swapped lateral control bits relative to FSD 13. Detecting this type of change requires CAN signal analysis (e.g., comparing FSD 13 vs FSD 14 routes) rather than firmware string matching, and is therefore outside the scope of firmware-based fuzzy fingerprinting. This is discussed in the PR comments.

### Are there any variants of our supported cars that won't work?
Vehicles running FSD 14 may require additional CAN-level detection beyond fuzzy fingerprinting. The EPS firmware string does not encode FSD version information, so distinguishing these variants would need to be handled via CAN signal detection in carstate rather than fingerprinting.

## Testing
- 14 comprehensive unit tests covering:
  - Model 3/Y/X firmware parsing
  - Invalid firmware handling
  - Unseen firmware versions (future-proofing)
  - Unknown models (Model S, etc.) correctly rejected
  - Edge cases (empty FW, wrong ECU address)
- All existing firmware versions in FW_VERSIONS database verified to parse correctly

## Files Changed
- `opendbc/car/tesla/values.py` - Added `match_fw_to_car_fuzzy()` and supporting functions
- `opendbc/car/tesla/tests/test_tesla.py` - New test file with comprehensive coverage
- `opendbc/car/tesla/tests/__init__.py` - Test package init

### DIFF
--- a/opendbc/car/tesla/tests/test_tesla.py
+++ b/opendbc/car/tesla/tests/test_tesla.py
@@ -9,12 +9,6 @@ from opendbc.car.tesla.fingerprints import FW_VERSIONS
 Ecu = CarParams.Ecu
 
 
-# Known ECU address for Tesla EPS
-ECU_ADDRESSES = {
-  Ecu.eps: 0x730,
-}
-
-
 class TestTeslaFW:
   def test_fw_pattern_model3(self):
     """Test that Model 3 firmware patterns are correctly parsed."""
@@ -87,7 +81,6 @@ class TestTeslaFW:
     ]
     for fw in invalid_fws:
       model_code = get_model_code(fw)
-      # Empty model code after comma should still fail platform mapping
       if model_code:
         platform = get_platform_from_model_code(model_code)
         if platform:
--- a/opendbc/car/tesla/values.py
+++ b/opendbc/car/tesla/values.py
@@ -65,16 +65,6 @@ class CAR(Platforms):
 
 # Tesla EPS firmware: "<project>,<model_code><version>"
 # Model codes: E* = Model 3, Y* = Model Y, X* = Model X
-#
-# API Consistency Notes (bounty requirement):
-# - Steering API is consistent across HW2.5, HW3, and HW4 for Model 3/Y
-# - HW2.5 sends longitudinal control at 40Hz instead of 25Hz
-#
-# Known Limitations:
-# - Some HW2.5 vehicles without ACC capability cannot utilize longitudinal control
-#   due to different Drive Interface firmware versions (these still fingerprint but
-#   will be detected at runtime)
-# - Model S is not currently supported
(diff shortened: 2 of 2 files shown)


sshane:
> Has Tesla ever changed the APIs we use?
No. The steering API is consistent across HW2.5, HW3, and HW4 for Model 3/Y. HW2.5 sends the same signal for longitudinal control but at a different frequency (40Hz instead of 25Hz).

This is false, FSD 14 swapped around the lateral control bits we'd need to detect with this, or in carstate after we fingerprint

rosacry:
Thanks for the correction - I've updated the PR description to remove that claim.

For detecting FSD 14's lateral control changes, would you like me to extend this fuzzy fingerprinting to detect FSD version from the firmware? Or is that better handled in carstate?

sshane:
I'm not sure there's enough information in the FW in https://github.com/commaai/opendbc/pull/2934 to know what changed with the API. We'd have to compare FSD 13 and 14 routes to find a reliable CAN signal to detect.

rosacry:
Got it, makes sense - FSD 14 detection is a separate problem that needs CAN signal analysis.

Is this PR otherwise good to go for the basic fuzzy fingerprinting (E→Model 3, Y→Model Y, X→Model X)?

rosacry:
@sshane (and anyone else reviewing) quick ping — I updated the PR description to address the bounty questions (Tesla API changes / FSD14 variants) and clarified scope. All checks still pass; ready for re-review when you have a moment.

sshane:
I don't think we understand how Tesla can issue FW updates to change behavior enough yet to support fuzzy fingerprinting right now. Thanks for the PR though.

rosacry:
Understood, that's a fair concern. Tesla's OTA behavior is still a bit of a black box. I'll keep an eye on the issue and can revisit this if there's more data on how their FW updates affect the CAN interface in practice. Thanks for taking the time to review!

PR #2944, opened 2025-12-17
safety: Enforce 100% branch coverage

mpurnell1:
## Description
This PR makes progress towards #2557 by rewriting production code to eliminate structurally dead branches from coverage. Combined with the test additions in #3148, branch coverage goes from 92.5% on master to 98.0%.

Changes by commit:
1. **Add `ADDR_BUS_MATCH` macro** — combines `(msg->addr == X) && (msg->bus == Y)` into a single comparison, eliminating short-circuit branches that are always true (bus is pre-validated by `rx_checks`)
2. **Flatten nested bus wrappers** — uses `ADDR_BUS_MATCH` to replace `if (bus == X) { if (addr == Y) ...}` patterns across gm, mazda, subaru_preglobal, toyota, volkswagen_mqb, volkswagen_pq
3. **Rewrite compound conditions** — reorder `&&` operands so config flags (which vary across test configs) are checked first, convert `||` to `|` where all operands are bool, flatten redundant bus wrappers in hyundai/hyundai_canfd
4. **Rewrite dispatchers and curvature clamp** — use ternary/early-return in 2-address dispatcher functions (rivian, psa, hyundai_common) to eliminate dead `else if` false branches; merge lateral.h curvature CLAMP calls to eliminate structurally dead `x > high` branches

Related PRs:
- [x] #2948
- [x] #2949
- [ ] #3148

## Verification

`opendbc/safety/tests/test.sh` on master with branch coverage temporarily enabled
```
TOTAL                                       1890     1748    92%
```

`opendbc/safety/tests/test.sh` on this branch with branch coverage temporarily enabled
```
TOTAL                                       1810     1773    98%
```

37 remaining uncovered branches are in: safety.h framework internals, ford/tesla/hyundai dispatcher `else if` chains (3+ addresses), volkswagen_common/volkswagen_pq, helpers.h interpolation edge case, and body.h/gm.h/psa.h single-address guards.

PR #2952, opened 2025-12-19
tests: add CAN flag and rx/tx hook test helpers

mpurnell1:
## Description
This makes incremental progress towards #2944. In this PR, I create 6 test helper functions in C to directly call rx hooks, bypassing the whitelists which would usually prevent these invalid calls from going through. Subsequent PRs will use these helpers to test full branch coverage of rx and tx hooks in all modes. I'm opening this PR to begin the discussion as to whether or not this is how we want to solve the difficult-to-cover empty `else` branches which are found in many mode header files. [example 1](https://github.com/commaai/opendbc/blob/3ff537155e91ba8bf8df66d8e62dfafbed9f6390/opendbc/safety/modes/ford.h#L32) [example 2](https://github.com/commaai/opendbc/blob/3ff537155e91ba8bf8df66d8e62dfafbed9f6390/opendbc/safety/modes/hyundai.h#L74) [example 3](https://github.com/commaai/opendbc/blob/3ff537155e91ba8bf8df66d8e62dfafbed9f6390/opendbc/safety/modes/psa.h#L24)

From what I can tell, we only include safety.c (where the helpers are defined) during tests, and never during production builds for panda. Thus, it should be safe to create these functions which bypass the safety functionality of the whitelists. However, please let me know if you think tests for these helper functions should be added, or if you would like any checks to be performed inside these functions as sanity checks, such as hiding the definitions behind `#ifdef ALLOW_DEBUG` or something similar.

## Verification:
No major changes, 6 new APIs exposed but never used. Coverage is decreased but will recover as the APIs start getting used in other PRs (see #2953).
`opendbc/safety/tests/test.sh --report` on master:
image

`opendbc/safety/tests/test.sh --report` on this branch:
image


mpurnell1:
@sshane this PR is ready for review

PR #2970, opened 2025-12-19
Move CAN Ignition Hooks from panda to opendbc

### LINKED ISSUE #1834: Move CAN ignition hooks and add test

rosacry:
Closes #1834

Ports CAN ignition detection from panda's `board/drivers/can_common.h` to opendbc.

### Detection Logic

| Vehicle | Address | Logic |
|---------|---------|-------|
| GM | 0x1F1 | `data[0] & 0x2` |
| Mazda | 0x9E | `data[0] >> 5 == 6` |
| Rivian | 0x152 | `data[7] >> 4 & 0x3 == 1` with counter mod 15 |
| Tesla | 0x221 | `data[0] >> 5 & 0x3 == 3` with counter mod 16 |

### Files Changed
- `opendbc/safety/ignition.h` - ignition detection logic
- `opendbc/safety/safety.h` - include, call from rx hook, init on mode change
- `opendbc/safety/declarations.h` - extern declarations


sshane:
Ready for review?

rosacry:
Yes, I believe so. Let me know if i'm missing anything

opendbc/safety/tests/test_can_ignition.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python3
+import unittest
+
+from opendbc.car.structs import CarParams
+from opendbc.safety.tests.libsafety import libsafety_py
+from opendbc.safety.tests.common import make_msg
+
+
+class TestCANIgnition(unittest.TestCase):
```
adeebshihadeh: these tests are pretty useless, let's just remove them. a better test, for example, would be to assert in test_models that `self.safety.get_ignition_can()` matches the CAN ignition value at runtime

opendbc/safety/ignition.h
```diff
@@ -0,0 +1,56 @@
+// CAN ignition detection ported from panda/board/drivers/can_common.h
```
adeebshihadeh: useless comment

looks like it's from an LLM (as does this PR's description). please clean up any LLM junk before asking for a review again

opendbc/safety/safety.h
```diff
@@ -443,6 +447,9 @@ int set_safety_hooks(uint16_t mode, uint16_t param) {
   reset_sample(&torque_driver);
   reset_sample(&angle_meas);
 
+  // reset CAN ignition state
+  ignition_can_init();
```
adeebshihadeh, line 449: this clearly doesn't work. how did you test this code?

opendbc/safety/tests/test_can_ignition.py
rosacry: removed the standalone test file and added ignition tests to the vehicle-specific test files (test_gm.py, test_mazda.py, test_rivian.py, test_tesla.py). all tests pass with 100% coverage.

opendbc/safety/ignition.h
rosacry: removed

opendbc/safety/safety.h
rosacry: it works, all safety tests pass including the new ignition tests. The function is static in ignition.h, which is included before the call site in safety.h.

sshane: this function is only called after fingerprinting, correct? see the mode and param at the top of the function. what happens from before fingerprinting to setting the safety mode? `ignition_can` is reset back to 0, along with the counters?

opendbc/safety/ignition.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include "opendbc/safety/declarations.h"
+
+bool ignition_can = false;
+uint32_t ignition_can_cnt = 0U;
+static int prev_rivian_cnt = -1;
+static int prev_tesla_cnt = -1;
+
+static void ignition_can_hook(const CANPacket_t *to_push) {
```
sshane, line 10: the style varies from the panda ignition function (yours left, panda right), I would prefer a clean copy/paste rather than subtle rewrite

image

sshane:
With these issues plus no panda PR to finish this project, I don't fully trust the work done here, so I'm going to close

opendbc/safety/safety.h
```diff
@@ -211,6 +212,8 @@ bool safety_rx_hook(const CANPacket_t *msg) {
     heartbeat_engaged_mismatches = 0;
   }
 
+  ignition_can_hook(msg);
+
```
sshane, line 216: Forgot if it builds this way, but without removing this from fdcan.h, we call this twice:

```
    safety_rx_invalid += safety_rx_hook(&to_push) ? 0U : 1U;
    ignition_can_hook(&to_push);
```

rosacry:
@sshane I've created the panda PR you mentioned: commaai/panda#2323 (checks passing)

It removes the ignition_can_hook call from fdcan.h and the function definition, so there's no duplicate call anymore. The variables stay in panda since main.c uses them, and opendbc just externs them.

Would you be open to reopening this? Both PRs need to merge together for it to work.

PR #2971, opened 2025-12-19
fix(SConscript): use build_env instead of env in loop

rosacry:
## Summary
Fixes a bug in the libsafety SConscript where the loop variable `build_env` is ignored, causing both regular and mutation testing builds to use the same environment.

## The Bug
The loop iterates with `build_env` alternating between `env` and `menv`, but the body always uses `env`:
```python
for build_env, suffix in ((env, ""), (menv, "_mutation")):
    safety = env.SharedObject(...)    # Always uses `env`
    libsafety = env.SharedLibrary(...)
```

This means `libsafety_mutation.so` should use `menv` (clang-17, Mull plugin, coverage flags) but incorrectly uses `env`.

## The Fix
Use the loop variable `build_env` instead of hardcoded `env`:
```python
safety = build_env.SharedObject(f"safety{suffix}.os", "safety.c")
libsafety = build_env.SharedLibrary(f"libsafety{suffix}", [safety])
```

## Files Changed
- `opendbc/safety/tests/libsafety/SConscript` - Use `build_env` instead of `env` in loop body


rosacry:
What looked like a bug was actually working code with implicit CI dependencies. My bad.

PR #2919, opened 2025-11-29
Honda - electric brake booster refactor

mvl-boston:
Refactor of electric brake booster
- Renaming from alt_brake to electric_brake_booster, message correlates with presence of electric brake booster in fingerprint
- Moving from manual flagging to autodetect, to make it easier to merge new cars.

Todo:
- [x] Validate in test routes that there is no model with a 0x1BE messages that was not previously mapped to ALT_BRAKE - confirmed see [here](https://github.com/mvl-boston/opendbc/actions/runs/26000378654/job/76422362378?pr=415)
- [x] Validate in huggingface that there is no model with a 0x1BE messages that was not previously mapped to ALT_BRAKE - confirmed see [code](https://github.com/mvl-boston/opendbc/blob/datafile/docs/scan_1be.py) and [output](https://github.com/mvl-boston/opendbc/blob/datafile/docs/1be_output)
- [x] Validate no 0x1BE in Nidec CIVIC_BOSCH route in huggingface - despite electric brake booster in fingerprints.py - see above

PR #2973, opened 2025-12-20
Honda - support RDX 3G variants with brake booster (alt_brake)

mvl-boston:
Some 3rd gen RDX models have a brake booster and thus need the alt-brake flag autodetected.

alt-brake causes the correct brakepressed carstate to be set.

d9ac1a7b6755fff9/0000001f--8e7db88e99


BaharathBathula (commented):
Nice contribution. You might consider separating this logic into a smaller
helper function to improve readability and testability.

mvl-boston:
> Nice contribution. You might consider separating this logic into a smaller helper function to improve readability and testability.

Not sure what you are recommending?  Long term https://github.com/commaai/opendbc/pull/2919 is cleaner, but if comma prefers less of a change than this PR will solve the short term issue.

sshane:
I need to be caught up with the Honda brake pressed bits, the dongle you linked has both BRAKE_MODULE and POWERTRAIN_DATA, why use BRAKE_MODULE over the existing signal? On other cars, do they lack the POWERTRAIN_DATA brake pressed bit and this PR is only for correctness? If so, I think we should default to POWERTRAIN_DATA unless needed.

Also here the existing signal catches brake held at a stop and new does not. Is the user braking some other way and existing signal is wrong?:

image

mvl-boston:
POWERTRAIN_DATA is the proper user brake signal if there is no brake booster (eg: non_alt_brake).

If there is a brake booster installed (eg: presence of BRAKE_MODULE message), then POWERTRAIN_DATA represents either user_brake or ACC brake.  And user brake signal is in BRAKE_MODULE instead.

If brake booster cars are left pointing carstate brakepressed to POWERTRAIN_DATA, then comma will improperly disengage (mis-assuming user brake) any time ACC attempts to engage brakes.

opendbc/car/honda/interface.py
```diff
@@ -197,7 +197,7 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
     # These cars use alternate user brake msg (0x1BE)
-    if 0x1BE in fingerprint[CAN.pt] and candidate in (CAR.HONDA_ACCORD, CAR.HONDA_HRV_3G, *HONDA_BOSCH_CANFD):
+    if 0x1BE in fingerprint[CAN.pt] and candidate in (CAR.HONDA_ACCORD, CAR.HONDA_HRV_3G, CAR.ACURA_RDX_3G, *HONDA_BOSCH_CANFD):
```
sshane, line 200: Wonder why we have this set if we do live detection, can any car have 0x1BE and we don't want to use it? Removing would simplify things

mvl-boston: I agree.   PR2919 does generic live detection.  I had wanted to learn huggingface scripting to do a global regression test before opening gates broadly.  Hadn't had a chance to try yet which is why 2919 is in draft.

PR #2776, opened 2025-09-11
Honda - Odyssey Singapore

mvl-boston:
Add support for 2021 Odyssey Singapore

- 2021 is Nidec in Singapore, unlike Bosch in USA
- Same controls as 2019 Taiwan Odyssey

Todo:
- [x] Testroute with new/cleaned up logic (see routes.py), merged with Taiwan Odyssey

PR #2985, opened 2025-12-23
Honda: Taiwan Odyssey 2019

eFiniLan:
Hi team,

I attempted to port the 2018 model [2 years ago](https://github.com/eFiniLan/openpilot/tree/odyssey_2018_twn) but lacked logs; we have now confirmed with a MY2019 that it uses the same fingerprint and architecture.

This model requires a different Nidec configuration than the honda_odyssey_exl_2018 DBC.

Validation
* Dongle ID: 7817fe954aff07b8
* Route: 7817fe954aff07b8/00000001--fdaaf36c4f


mvl-boston:
You can point to honda_crv_touring_2016_can.dbc instead of making a new dbc.

Seems like many Asian vehicles don't have a carspeed message.   I wrote an autodetect in https://github.com/commaai/opendbc/pull/2776 that could be shared here.

eFiniLan:
> You can point to honda_crv_touring_2016_can.dbc instead of making a new dbc.
> 
> Seems like many Asian vehicles don't have a carspeed message. I wrote an autodetect in #2776 that could be shared here.

thanks! that approach looks nice, I checked your code and it seems nearly identical :joy:
perhaps I should just submit the fingerprint :joy: 

edit:

actually we tested your branch and the lateral didnt work, we need to create another steering sensor dbc for it. :/

mvl-boston:
> > You can point to honda_crv_touring_2016_can.dbc instead of making a new dbc.
> > Seems like many Asian vehicles don't have a carspeed message. I wrote an autodetect in #2776 that could be shared here.
> 
> thanks! that approach looks nice, I checked your code and it seems nearly identical 😂 perhaps I should just submit the fingerprint 😂

Taiwan seems to use a different steering status from Singapore.  Taiwan can point to honda_crv_touring_2016_can_generated.dbc, but Singapore needs a new DBC since it uses the 9G Accord steer status.  So the fingerprints will need to be pointed to different models.   Feel free to DM me at MVL on discord if you need any assistance.

mvl-boston:
> 
> actually we tested your branch and the lateral didnt work, we need to create another steering sensor dbc for it. :/

Yes are you stuck mapping any messages?   Most of the cars in that time fell into a few families and I may have already seen them.   Feel free to DM me a log in discord and I can connect anything you're missing.

opendbc/car/honda/values.py
```diff
@@ -324,6 +324,12 @@ class CAR(Platforms):
     radar_dbc_dict('honda_odyssey_exl_2018_generated'),
     flags=HondaFlags.NIDEC_ALT_PCM_ACCEL | HondaFlags.HAS_ALL_DOOR_STATES,
   )
+  HONDA_ODYSSEY_TWN = HondaNidecPlatformConfig(
+    [],
```
sshane: Why no car docs?

opendbc/car/honda/interface.py
```diff
@@ -215,7 +219,7 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
     # to a negative value, so it won't matter. Otherwise, add 0.5 mph margin to not
     # conflict with PCM acc
     ret.autoResumeSng = candidate in (HONDA_BOSCH | {CAR.HONDA_CIVIC})
-    ret.minEnableSpeed = -1. if ret.autoResumeSng else 25.51 * CV.MPH_TO_MS
+    ret.minEnableSpeed = -1. if ret.autoResumeSng else (19. if candidate == CAR.HONDA_ODYSSEY_TWN else 25.51) * CV.MPH_TO_MS
```
sshane: Let's expand this into a real if else statement for clarity

opendbc/car/honda/carstate.py
```diff
@@ -68,7 +70,7 @@ def update(self, can_parsers) -> structs.CarState:
     self.cruise_buttons = cp.vl["SCM_BUTTONS"]["CRUISE_BUTTONS"]
 
     # used for car hud message
-    self.is_metric = not cp.vl["CAR_SPEED"]["IMPERIAL_UNIT"]
+    self.is_metric = True if self.CP.carFingerprint in (CAR.HONDA_ODYSSEY_TWN) else not cp.vl["CAR_SPEED"]["IMPERIAL_UNIT"]
```
sshane:
```suggestion
    self.is_metric = self.CP.carFingerprint in (CAR.HONDA_ODYSSEY_TWN,) or not cp.vl["CAR_SPEED"]["IMPERIAL_UNIT"]
```

sshane:
Why did the diff bot report changes?

eFiniLan:
Thanks @sshane ! Updated the PR per your comments.

sshane:
Oh https://github.com/commaai/opendbc/pull/2776 is a good reason to autodetect CAR_SPEED. @mvl-boston know if we can use ACC_HUD for the imperial unit for all Nidecs? Also wonder about the cluster speed, but not needed on comma four

sshane:
@elkoled why did the bot fail with a huge output here?

elkoled:
That were the updated refs of my CarParams test.
They are back to normal again.

PR #2986, opened 2025-12-24
Honda - Euro Civic Type R FK8 support

mvl-boston:
Similar to Bosch_Civic, needs added manual transmission flag and fingerprints

8305f0cbb669656c|00000007--7570283ee1

Thanks to atmezferix.


sshane:
thanks and sorry for the delay

PR #2996, opened 2025-12-29
Hyundai: document SCC_CONTROL signal preservation for ACC cancel

nikitap066:
Document why SCC_CONTROL signals must be preserved when cancelling ACC on Hyundai CAN FD camera-based platforms.

No functional changes.

* Dongle ID: N/A
* Route: N/A


adeebshihadeh:
Great PR, thanks!

PR #3009, opened 2025-12-30
Toyota: whitelist hybrids for standstill resume behaviour

nikitap066:
Implements TODO to whitelist Toyota hybrid vehicles, which do not exhibit the standstill resume lurch seen on non-hybrids.

UPDATED: better hybrid check logic to ensure correct standstill behaviour

No functional change for non-hybrid vehicles.

Dongle ID: N/A
Route: N/A


sshane:
thanks!

PR #3012, opened 2025-12-31
FSD 14: support

sshane:


sshane:
found no reliable bits split by FW version yet. continuing in https://github.com/commaai/opendbc/pull/2934

PR #3014, opened 2026-01-01
Honda Bosch - live-learning gas pedal tuning

mvl-boston:
This PR updates Honda Bosch gas control.

Honda Bosch A & C ACC uses a m/s2 deceleration command for brake, but uses gas-pedal-position command for acceleration.  

Three different adjustments are needed to properly calculate what gas-pedal-position will result in a target acceleration.

1) Pitch/hill adjustment - This is a trigonometry calculation, mimicking the calculation used by Ford/Toyota

2) Wind drag adjustment - Wind drag for an average car is calculated [here](https://github.com/mvl-boston/opendbc/raw/refs/heads/datafile/docs/drag_calculation.xlsx).   However this can vary based on car shape, elevation, air density, and wind.  Thus a live-learning factor is introduced to update the value used in the calculation.

3) Pedal-to-force mapping - 80% of Honda's use a linear mapping of pedal position to force.  However, certain cars have physical or artificial reasons where this was not programmed linearly by Honda.  For example, the CVT Honda Accord creates simulated gearshifts to "feel" like a regular automatic transmission and the factor changes per simulated gear, and the Acura RDX has increased force when the turbocharger spools up.    Thus a live-learning factor quickly reacts to changing mapping ratios, the factor was tuned to react in approximately 1 second to changes, to behave like a human driver would react.

This logic has been added to all my custom forks that are widely used by many Honda drivers, and has been in place since December 12.   The actuators.accel vs aEgo match is near perfect in plotjuggler curves for all tested Bosch A and C cars, after a few minutes of learning.   Driver feedback is that this is a night and day improvement vs existing Honda Bosch long-alpha control.

I'm happy to ask testdrivers to provide whatever testlogs are needed by comma to validate this change.  Please advise which logs to request.


sshane:
This kind of improvement should be done such that it's easy to extend to other brands, GM for example. I don't think we have high priority plans on adding manually written live learning at the moment (NN would be ideal but is hard). If that's not possible, we can close for now and revisit.

mvl-boston:
Not familiar with NN, will study and figure out if this can be moved to it

sshane:
I'm just saying that most large sweeping tuning changes usually come from comma since it requires a lot of data to validate that control is better for all of our users with varying cars and responses. A generic learning approach live in the car is the best approach, but no one has done this reliably or low computationally yet for robotics. The next best approach is something like torqued that is generic and works for many of our supported cars, but still makes some hand-written assumptions that break down for some cars like a linear torque response or exponential speed curve (GM).

It's also important to try and attempt the smaller thing first, like a PR that just adds pitch compensation, which was proven to be a big improvement for Toyota. Or a PR that makes that pitch compensation generic so it's just a ~one line change for Honda. That makes the entire set of desired changes easier to validate and merge.

Please feel free to ask more questions on how to merge improvements like this, I'm happy to answer. But this is too complex I think to merge as-is, and partially requires autonomy team to validate in simulation (? @felsager).

PR #3018, opened 2026-01-03
chrysler: add fuzzy fingerprinting using platform part numbers

### LINKED ISSUE #1092: Chrysler: reliable fuzzy fingerprinting

AgentsLogic:
Implements fuzzy fingerprinting for Chrysler/Dodge/Jeep/Ram vehicles by matching on ECU part numbers while ignoring the 2-character revision suffix.

## Firmware Format

Chrysler firmware versions follow the format: `XXXXXXXXYY` where:
- `XXXXXXXX` = 8-digit part number (platform identifier)
- `YY` = 2-character revision code (changes with software updates)

For example: `68227902AF` → part number `68227902`, revision `AF`

## Implementation

- `get_platform_codes()`: Extracts part numbers by stripping the 2-char revision suffix
- `match_fw_to_car_fuzzy()`: Matches live firmware to platforms using part numbers only
- ECUs used for matching: `combinationMeter`, `abs`, `fwdRadar`, `eps`

## Tests

- Unit tests for `get_platform_codes()` function
- Parametrized tests verifying all platforms can be matched with mutated revisions
- Negative tests for wrong part numbers and empty input

Closes #1092


sshane:
Can you run this on a few recent Chrysler FP PRs and report back if this now fingerprints them? (remove all their fingerprints locally when testing)

Pa-yfi:
I'm not the author of this PR, but re: sshane's request above to run this against recent Chrysler FP PRs — I scripted that check and figured the numbers were worth posting here rather than sitting in a local checkout.

Instead of picking a few FP PRs by hand, I ran it over the whole history of opendbc/car/chrysler/fingerprints.py (19 commits). For every FW version ever added to that file, the database is rebuilt as it stood immediately before that commit, the newly-added FW is presented as if it came off a live car, and the fuzzy matcher is asked to identify it.

Against this PR's matcher:

replayed = 70
  correctly matched               = 52
  safely unmatched (no candidate) = 18
  WRONG / ambiguous match         = 0

So 52 of 70 would have fingerprinted with no false positives. Every one of the 18 misses is a commit that introduced a new part number (new trim or model year), not a new revision of a known part — staying unmatched there is the correct behaviour. Unseen revisions of known part numbers all match.

One thing that fell out of the same data: how well each ECU type actually discriminates platforms, measured as how many of its part numbers appear on more than one platform.

  combinationMeter    0 / 103
  engine              0 / 125
  transmission        0 / 73
  eps                 1 / 33
  abs                 2 / 32
  srs                 3 / 58
  fwdRadar            6 / 24

fwdRadar is in this PR's PLATFORM_CODE_ECUS and is the weakest of the group — 04672758, for example, appears on four different Pacifica platforms. Substituting srs for it gives identical backtest numbers, but should be more robust on a car that isn't in the database yet.

Separately, worth flagging a gap against the Ford spec this bounty inherits (openpilot 31052): that asks for the platform code to be parsed with a way to split API/MY differences, so that an unseen model year still fingerprints from minimal FW. Matching on the full 8-character part number doesn't do that, which is precisely why those 18 cases miss. The safety half of the requirement looks met here; the generalization half isn't yet.

Script used, if you want to re-run it or point it at a different ECU set (needs full history, so git fetch --unshallow first):

```python
import re, subprocess, types
from opendbc.car.structs import CarParams
from opendbc.car.chrysler.values import match_fw_to_car_fuzzy
E = CarParams.Ecu
FP = "opendbc/car/chrysler/fingerprints.py"
class P:
  def __init__(s, f): s.f = f
  def __getattr__(s, k): return s.f(k)

def load(rev):
  try: src = subprocess.check_output(["git","show",f"{rev}:{FP}"], text=True, stderr=subprocess.DEVNULL)
  except subprocess.CalledProcessError: return None
  src = re.sub(r'^\s*(from|import) .*$', '', src, flags=re.M)
  ns = {'Ecu': P(lambda k: getattr(E,k)), 'CAR': P(lambda k: k),
        'CarParams': types.SimpleNamespace(Ecu=P(lambda k: getattr(E,k)))}
  try: exec(src, ns)
  except Exception: return None
  return ns.get('FW_VERSIONS')

log = subprocess.check_output(["git","log","--format=%H","--follow","--reverse","--",FP], text=True).split()
head = load('HEAD')
for p, e in head.items():   # sanity: a car built from its own DB rows must match only itself
  assert match_fw_to_car_fuzzy({(a,s):{v[0]} for (_,a,s),v in e.items()}, '', head) == {p}, p

ok = miss = wrong = 0; prev = None
for sha in log:
  cur = load(sha)
  if cur is None: continue
  if prev is None: prev = cur; continue
  for plat, ecus in cur.items():
    if plat not in prev: continue
    for ecu, vers in ecus.items():
      for nv in [v for v in vers if v not in prev.get(plat,{}).get(ecu,[])]:
        live = {}
        for o, ov in ecus.items():
          known = prev.get(plat,{}).get(o,[]) or ov
          live[(o[1],o[2])] = {nv} if o == ecu else {known[0]}
        r = match_fw_to_car_fuzzy(live, '', prev)
        if r == {plat}: ok += 1
        elif not r: miss += 1
        else: wrong += 1
  prev = cur
print(f"replayed={ok+miss+wrong} matched={ok} safely_unmatched={miss} WRONG={wrong}")
```

The sanity assert matters — if the Ecu values don't resolve to the real enum the matcher trivially returns every platform and the whole run looks perfect for the wrong reason.

PR #3022, opened 2026-01-05
VW: Allow PQ/MLB cars to drive on dev builds

jyoung8607:
Per recent discussions on Discord, allow PQ and MLB platform cars to drive on master and nightly-dev.

While here, fix a docs bug with overlapping model-years that tests exposed.


luckyducky8676:
Hi, saw your thread on AudiWorld looking for testers. I have a b8.5 allroad prestige with the EPS rack and ACC and would be a good candidate. Can we please get lateral steering going? I saw that there was an SBieger implementation too

PR #3026, opened 2026-01-07
Add a "Heavy" flag to Hyundai for heavier vehicles that could use more torque

SpysyWeeb:
Validation
* Dongle ID: ff9dba54b42b3afa
* Route: https://connect.comma.ai/ff9dba54b42b3afa/0000000a--33ba3cf345

This lays the ground work for heavier Hyundai vehicles that could use more torque in curves to simply add the "Heavy" flag to the specific model. The Hyundai Palisade has been used to test, and has shown improvement with this change.


sshane:
The problem with this is that openpilot can set a safety parameter and now gain access to extra torque, possibly violating ISO11270 limits. We want to move away from `ALT_LIMITS`-style params

PR #3027, opened 2026-01-07
Mazda: Add cruise button events for accel and decel

nikitap066:
Implements TODO to add cruise control button events for accel and decel on Mazda platforms.

No functional behavior change; exposes existing CAN signals to CarState.

Validation
Dongle ID: N/A
Route: N/A

### DIFF
--- a/opendbc/car/mazda/carstate.py
+++ b/opendbc/car/mazda/carstate.py
@@ -28,13 +28,6 @@ class CarState(CarStateBase):
 
     ret = structs.CarState()
 
-    prev_distance_button = self.distance_button
-    prev_accel_button = self.accel_button
-    prev_decel_button = self.decel_button
-    self.distance_button = cp.vl["CRZ_BTNS"]["DISTANCE_LESS"]
-    self.accel_button = cp.vl["CRZ_BTNS"]["RES"]
-    self.decel_button = cp.vl["CRZ_BTNS"]["SET_M"]
-
     self.parse_wheel_speeds(ret,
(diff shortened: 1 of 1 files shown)


opendbc/car/mazda/carstate.py
```diff
@@ -112,8 +118,12 @@ def update(self, can_parsers) -> structs.CarState:
...
     ret.steerFaultPermanent = cp_cam.vl["CAM_LKAS"]["ERR_BIT_1"] == 1
 
-    # TODO: add button types for inc and dec
-    ret.buttonEvents = create_button_events(self.distance_button, prev_distance_button, {1: ButtonType.gapAdjustCruise})
+    # cruise control button events: distance, inc, and dec
+    ret.buttonEvents = [
+      *create_button_events(self.distance_button, prev_distance_button, {1: ButtonType.gapAdjustCruise}),
+      *create_button_events(self.accel_button, prev_accel_button, {1: ButtonType.accelCruise}),
+      *create_button_events(self.decel_button, prev_decel_button, {1: ButtonType.decelCruise}),
+    ]
```
sshane, line 126: I don't believe openpilot uses these buttons for anything functional yet since this car is `pcmCruise`, is that correct?

nikitap066: Yes, that’s correct. This change just exposes the existing INC/DEC button signals in CarState, and resolves the TODO without altering behavior. It should make the signals available if they’re needed in the future.

jixiexiaoge:
a feedback  to you , carrotpilot use this to CLSC , and works good on my CX-5

PR #3024, opened 2026-01-07
Car diff

elkoled:
## Car diff - CarState regression testing
This PR implements a diff method to catch behavioral changes in CarState by replaying a subset of carCommaSegments through CarInterface and comparing their CarState output against master refs in ci-artifacts.

Example:
https://github.com/elkoled/opendbc/pull/245

### CI
- **PR:** Detects changed platforms, downloads master refs as zst from ci-artifacts, replays segments from commaCarSegments, reports any behavior differences as bot comment
- **master:** Generates and uploads new references for changed platforms

### Usage
**auto-detect platform**
`source setup.sh && python opendbc/car/tests/car_diff.py`

**test specific platform**
`source setup.sh && python opendbc/car/tests/car_diff.py --platform TOYOTA_COROLLA`

### CI Benchmark
Measured on standard gh actions runner

| Brand   | Platforms                | Segs | CI time |
|-----------------|----------------------------------|------|---------|
| hyundai    | 56 (Hyundai/Kia/Genesis)         | 260  |  8:41   |
| toyota    | 36 (Toyota/Lexus)                | 190  |  7:09   |
| honda     | 21 (Honda/Acura)                 | 154  |  5:30   |
| volkswagen | 21 (VW/Audi/Seat/Skoda)          | 130  |  4:41   |
| gm         | 12 (Chevy/Buick/Cadillac/GMC)    |  97  |  4:11     |
| ford      | 10                               | 91   |  4:00   |
| chrysler  | 9 (Chrysler/Jeep/Dodge/Ram)      | 90   |  2:32   |
| subaru    | 9                                | 69   |  3:56   |
| nissan     | 5                                | 50   |  3:16   |
| tesla      | 2                                | 20   |  2:16   |
| mazda      | 3                                | 17   |  2:12   |

PR #3028, opened 2026-01-08
body: optimize CRC-8 checksum with lookup table

KRRT7:
Replace bit-by-bit CRC-8 calculation with precomputed lookup table for improved performance.

### DIFF
--- a/.github/workflows/car_diff.yml
+++ b/.github/workflows/car_diff.yml
@@ -1,0 +1,42 @@
+name: car diff
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  comment:
+    name: comment
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    continue-on-error: true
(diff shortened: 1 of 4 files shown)


sshane:
can also use a @cache instead of globals to cache the generation if you go this route

KRRT7:
went with the CRC8J1850 example

KRRT7:
@sshane @jyoung8607 are you guys interested in performance optimizations for opendbc?

adeebshihadeh:
> @sshane @jyoung8607 are you guys interested in performance optimizations for opendbc?

Yes, but they need to be visible changes that someone will notice, not micro-optimizations. Making tests faster is a good place to start, e.g. https://github.com/commaai/opendbc/pull/3024.

KRRT7:
sounds good, I'll see what I can come up with.

sshane:
This looks like a good minor cleanup regardless. Verified the output is the same with a few thousand random bytearrays

PR #3029, opened 2026-01-09
Honda - clear startup Bosch faults

mvl-boston:
Bosch cars retain faults from prior drives during the next time they start up.  They clear within a minute of the new drive.

This PR disables cruise enable (which is reversible) for any faults that appear in the first 10 seconds, and allows them time to clear.  10 seconds was chosen because Bosch C startup faults seem to occur upon gateway open, which occurs around 7 seconds after startup.

Any faults after 10 seconds have normal accFaulted (permanent, until restarting engine) carstate flagged.

Recent drive from a canfd loaner: ad9840558640c31d/00000195--03d04b0d1c


sshane:
Do we know why they happen in the first place and how to avoid them? I vaguely remember stock LKAS trying to actuate and it not seeing the EPS react or something might be one cause?

mvl-boston:
We don't know how to avoid the Bosch C faults during the next drive after long control is enabled.   The car is somehow detecting the absence of a radar from a prior drive.  During the occasional times I had Bosch C loaners I attempted to match every message from stock vs hondacan and it still faulted.    My next attempt will be to find all the other messages that shut down when the radar is disabled and attempt to send a default value across all of them.

Other than above, the only reason for a fault has been loose/bad wiring.  Or snow on the radar.   This PR would allow successful usage in the first drive after these faults, rather than the current state where you need to drive until the fault clears, and then restart the engine for an opportunity for cruise to resume.

modhash:
@mvl-boston I have a Bosch C Honda Passport, is there any way I can help you test it?

PR #3036, opened 2026-01-15
Tesla: Add FSD active state

dzid26:
Someone posted on canserver discord these updated states from decompiled firmware.


sshane:
Can you see if they have information about why 2 and 1 swapped in FSD 14? https://github.com/commaai/opendbc/blob/ac94e0502326bee8e9c04f3759dc83d3eaac07c4/opendbc/dbc/tesla_model3_party.dbc#L429

sshane:
We may be able to use this to detect if a user ports their car without putting the FW in https://github.com/commaai/opendbc/blob/ac94e0502326bee8e9c04f3759dc83d3eaac07c4/opendbc/car/tesla/values.py#L76, or automatic detection ideally

sshane:
Looks good 
image

dzid26:
> Can you see if they have information about why 2 and 1 swapped in FSD 14?
> 
> https://github.com/commaai/opendbc/blob/ac94e0502326bee8e9c04f3759dc83d3eaac07c4/opendbc/dbc/tesla_model3_party.dbc#L429

Apparently in the MCU part of the software that they had there is no reference to `DAS_steeringControlType` msg id, so no clue there.

PR #3043, opened 2026-01-17
Chrysler: correct radar lateral position sign

nikitap066:
Implements TODO to fix radar lateral position (yRel) sign for Chrysler platforms to match the left-positive coordinate convention.

No functional behavior change beyond correcting the sign; improves correctness and consistency with other radar interfaces.

Validation
Dongle ID: N/A
Route: N/A


jyoung8607 (commented):
We don't implement longitudinal control for Chrysler yet, but this is still a minor behavior change. Do you have references or data showing this is a correct change? Known good DBC references, or debug UI observations from a fork that does implement longitudinal control? It's also customary to share an uploaded route demonstrating the issue and fix.

nikitap066:
From the Chrysler DBC, LAT_DIST is encoded as a signed lateral distance with scale 0.005 and offset -1000, producing both negative and positive values. Based on the DBC signal definition, positive LAT_DIST corresponds to objects on the right side of the vehicle. Since openpilot defines yRel as left-positive, the sign inversion is required to align coordinate frames.

I recognize that, without longitudinal control, validation is more limited, and I’m happy to close the PR if you’d prefer to wait for additional supporting data.

jyoung8607:
Ordinarily we'd want data to validate. However, since the radar DBC contains an [explicit CM_ annotation on the LAT_DIST signal](https://github.com/commaai/opendbc/blob/33c1840d0466c2983d4c87b55a73d81861289c15/opendbc/dbc/chrysler_pacifica_2017_hybrid_private_fusion.dbc#L230) indicating it's right-positive, and the impact here is low if we're wrong (not yet actually used for long control) I'll go ahead and merge. Thanks for the bug fix!

PR #3050, opened 2026-01-21
Tesla: Add support for Model 3

smurf88sys:
Validation
* Dongle ID: 81357f73c77eb4f0
* Route: 81357f73c77eb4f0/00000000--84e071bbb1/0


sshane:
Do you have FSD? Do you know what version if so?

smurf88sys:
@sshane No. I had the FSD free trial but it expired 2 weeks ago.
However, the fingerprint still changed, as I used Openpilot without problems before.

sshane:
Do you know the latest version of FSD you were using before? Also please upload a full route and its logs with Autosteer in operation

sshane:
If you ever updated to FSD14 and still have its EPS FW changes, I assume you will get an invalid LKAS setting alert sometime in the future when stock LKAS activates. Lmk if that ever happens

PR #3053, opened 2026-01-22
Tesla: add FSD 14 failsafe detection

sshane:
with help of https://github.com/commaai/opendbc/pull/3036, though need to check what its value is with just Autosteer, but we should already cover that with existing `invalidLkasSetting`

I was mildly concerned in https://github.com/commaai/opendbc/pull/2934 about merging new FW, but with https://github.com/commaai/opendbc/pull/3050 I realized we have no way of knowing if FSD 14 if user doesn't tell us that stock LKAS assist is broken, or that they aren't using openpilot at all if FSD is turned on still since we pass through its message thinking it's LKAS

Cases we cover:

### DIFF
--- a/opendbc/car/tesla/carstate.py
+++ b/opendbc/car/tesla/carstate.py
@@ -124,7 +124,7 @@ class CarState(CarStateBase):
       # 1. If in Autosteer or FSD, already caught by invalidLkasSetting
       # 2. If in TACC and DAS ever sends ANGLE_CONTROL (1), we can infer it's trying to do LKAS on FSD 14+
       angle_control = cp_ap_party.vl["DAS_steeringControl"]["DAS_steeringControlType"] == 1  # ANGLE_CONTROL
-      if angle_control and not self.CP.flags & TeslaFlags.FSD_14:
+      if not ret.invalidLkasSetting and angle_control and not self.CP.flags & TeslaFlags.FSD_14:
         self.suspected_fsd14 = True
 
       if self.suspected_fsd14:


sshane:
Found this signal that changes when the fault happens on my drive here from not passing through LKAS before the support:

https://connect.comma.ai/0f79c454f812791a/000000b2--45d83a671a/1551/1798

```
hex(msg)='0x3e9', bit_mismatches=205 of 733 (27.97%), byt_idx=6, bit_idx=7
hex(msg)='0x34f', bit_mismatches=515 of 1992 (25.85%), byt_idx=1, bit_idx=6
hex(msg)='0x2b9', bit_mismatches=2093 of 9192 (22.77%), byt_idx=0, bit_idx=1
hex(msg)='0x32b', bit_mismatches=151 of 905 (16.69%), byt_idx=4, bit_idx=5
hex(msg)='0x39b', bit_mismatches=1 of 800 (0.12%), byt_idx=3, bit_idx=0
```

it's `DAS_sideCollisionAvoid` and it goes to 0

sshane:
When it's doing LKA (FSD 13 route) this signal switches from L to R but never 0, so I assume 0 is a faulted/unavailable state? If it's unavailable we can't use it as a fault

image

sshane:
@elkoled why does the bot say those two segments' `invalidLkasSetting` drops? The entire segment it should be true both before and after from running carstate and printing `invalidLkasSetting`.

og segment:

image

sshane:
Oh you don't pass FW into the CarInterface I suspect

elkoled:
yes currently not, will try to add it in the next update.

sshane:
I run car_diff.py offline with `2c912ca5de3b1ee9/000002ca--b6db903149/16` and it always says invalidLkasSetting is true, so not sure why the comment says it falls on master

PR #3056, opened 2026-01-22
Fixup car differ

sshane:


opendbc/car/tests/car_diff.py
```diff
@@ -49,25 +49,31 @@ def load_can_messages(seg):
 def replay_segment(platform, can_msgs):
   from opendbc.car import gen_empty_fingerprint, structs
   from opendbc.car.can_definitions import CanData
-  from opendbc.car.car_helpers import FRAME_FINGERPRINT, interfaces
+  from opendbc.car.car_helpers import FRAME_FINGERPRINT, interfaces, can_fingerprint
```
sshane: @elkoled why are all the imports not at the top?

sshane:
:/ we have rlog.capnp, and carParams is ordinal `@69`, so we must add a bunch of junk lines to read it. we need openpilot logreading package!

elkoled:
Added FW to your changes:
[car-diff-add-fw](https://github.com/commaai/opendbc/tree/car-diff-add-fw)

sshane:
but it doesn't find it since it's not in rlog.capnp

opendbc/car/tests/car_diff.py
elkoled: relict from openpilot lazy imports

sshane: Those are not the predominant style we have in openpilot

opendbc/car/tests/car_diff.py
```diff
@@ -42,20 +43,21 @@ def dict_diff(d1, d2, path="", ignore=None, tolerance=0):
...
-  for msg in can_msgs[:FRAME_FINGERPRINT]:
-    for m in msg.can:
-      if m.src < 64:
-        fingerprint[m.src][m.address] = len(m.dat)
+  _can_msgs = ([CanData(can.address, can.dat, can.src) for can in m.can] for m in can_msgs)
+
+  def can_recv(wait_for_one: bool = False) -> list[list[CanData]]:
+    return [next(_can_msgs, [])]
+
+  _, fingerprint = can_fingerprint(can_recv)
```
sshane, line 61: to reduce duplication, uses real fp logic

opendbc/car/tests/car_diff.py
```diff
@@ -42,20 +43,21 @@ def dict_diff(d1, d2, path="", ignore=None, tolerance=0):
 def load_can_messages(seg):
   parts = seg.split("/")
   url = get_url(f"{parts[0]}/{parts[1]}", parts[2])
-  msgs = LogReader(url, only_union_types=True)
+  msgs = LogReader(url, only_union_types=True, sort_by_time=True)
```
sshane, line 47: sometimes logger logs out of order rarely from what I've seen

elkoled:
FW extraction works when extending rlog.capnp to @69

PR #3057, opened 2026-01-22
Add misc bug template

sshane:


sshane:
> No car changes detected

@elkoled I don't think we need this comment? Or only if car/dbc/safety is touched

adeebshihadeh:
I think the comment is good, otherwise you're unsure if it ran.

PR #3067, opened 2026-01-24
Car diff: skip comment on first timer PR

elkoled:
On first timer PR, tests.yml never runs, so car diff / comment which relies on tests.yml artifact must be skipped to prevent showing a failed action.


elkoled:
ready for review

PR #3072, opened 2026-01-27
VW: swap specific distance value checks for truthiness checks

RJWadley:
### goal: working distance buttons regardless of car-specific inconsistencies

VW's distance buttons generally have three possible values. 
- a value of 1 is in theory distance -1
- a value of 2 is in theory distance +1
- a value of 3 is in theory distance toggle

the actual behavior varies. I know that some PQ cars send a value of 1 when distance is pressed depending on equipment. Many cars will send a value of 3, but never use 1 or 2 even when manually adjusting distance. Presumably these inconsistencies could surface across the entire VW lineup.

Since openpilot doesn't really care about distance +1 vs -1 vs toggle, I find the cleanest fix here is to simply check the truthiness of the signal, rather than a specific value.

> a side note: why include 2? I've included 2 here for consistency. If a vehicle actually did use values 1 and 2, I don't want DIST+ and DIST- to be treated differently


RJWadley:
I don't have access to the routes above, but the diffs look as expected to me.

In the `VOLKSWAGEN_ATLAS_MK1` routes, the driver likely pressed the distance button four times in a row to adjust the following distance to the minimum.

In the `VOLKSWAGEN_GOLF_MK7` route, the driver likely either pressed the distance button but did not adjust it OR pressed the distance button and adjusted distance using the +/- buttons (these are sent as ACCEL+ and ACCEL- instead of reusing the distance button signal)

jyoung8607 (commented):
You're right that there's different behavior from different styles of button/stalk controls. This is probably a step in the correct direction, but I'll have to review logs from my cars.

sshane:
@jyoung8607 can we merge or should close? We want to start making faster decisions to stop PRs from sitting for months

jyoung8607:
> @jyoung8607 can we merge or should close? We want to start making faster decisions to stop PRs from sitting for months

I agree with you as a matter of general policy.

In this specific case, I need this open for a bit. It's an actual problem; `gapAdjustCruise` presses have never worked to change the longitudinal personality on my cars. I've never looked into it, but this is probably the fix. There's little sense in converting this to an issue and back to a PR again. I'll try to validate within the next week.

jyoung8607:
I'm at comma_hack 6 this weekend. I'll try to look at it next week.

RJWadley:
@jyoung8607 just bumping this one for whenever you have the time to check it out

PR #3073, opened 2026-01-28
Car diff: add typing

elkoled:
Add type hints to car_diff.py for easier review of the data structure.

Added 3 aliases for multiple used tuple types.


elkoled:
@sshane ready for review

sshane:
if you're verified these are all accurate, LGTM

PR #3075, opened 2026-01-28
Toyota: clean up radarUnavailable

sshane:
pt 2 of https://github.com/commaai/opendbc/pull/3074

`(NO_DSU_CAR - TSS2_CAR)` is just `{<CAR.TOYOTA_CAMRY>, <CAR.TOYOTA_CHR>}`, so just set these DBCs if we truly can't use them, not sure why (set here https://github.com/commaai/openpilot/pull/28707)


sshane:
@elkoled is it easy to add carParams to the refs to test for these kinds of changes?

elkoled:
should be not too hard, return CP from replay_segment and add CP to the refs.
I will make a branch.

PR #3070, opened 2026-01-24
Car diff: use refs for waveform

elkoled:
Resolves #3060

Use actual ref data instead of diff-only to get start and end value of the signals for ASCII waveform visualization.

## Reproduce
change toyota/carstate.py
**ret.brakePressed = True**

### Before 

`git checkout 776841f0`
`source setup.sh && python opendbc/car/tests/car_diff.py --platform TOYOTA_COROLLA`

```
TOYOTA_COROLLA - bc7a33f454ca54e8/2025-03-16--12-34-45/39
  brakePressed (6001 diffs)

  frames 0-6005
  master:   \___________________________________________________________________...
  PR:       ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾...
            fall: PR lags by 6001 frames (60009ms)

TOYOTA_COROLLA - bc7a33f454ca54e8/2025-03-16--11-54-08/27
  brakePressed (3064 diffs)

  frames 0-6
  master:   \______
  PR:       ‾‾\____
            fall: PR lags by 2 frames (19ms)

  frames 2933-6004
  master:   ‾‾‾‾‾\______________________________________________________________...
  PR:       ‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾...
            fall: PR lags by 3062 frames (30619ms)

 ```

### After

`git checkout 004e742`
`source setup.sh && python opendbc/car/tests/car_diff.py --platform TOYOTA_COROLLA`

```
TOYOTA_COROLLA - bc7a33f454ca54e8/2025-03-16--12-34-45/39
  brakePressed (6001 diffs)

  frames 0-6000
  master:   ____________________________________________________________________...
[...]

PR #3077, opened 2026-01-28
Car diff: add CarParams

elkoled:
Based on #3070 

Diff carParams alongside carState in car_diff.py, so changes to car flags and tuning values are caught.


sshane:
An idea: what if we just ran process replay with card as the sole process? It already has all the infrastructure set up to diff carState, carParams, sendcan, carOutput, liveTracks, which covers everything the car interfaces generate. If we want to add liveTracks here (radar interface), we need another big diff.

It also already fully re-creates the car interface with CAN fingerprints and carFw afaik, which we don't here yet.

Something similar to using openpilot's test_models here to reduce the duplication of LogReaders, capnp structs, etc. @adeebshihadeh any reason we didn't do this? The only thing we'd need is opendbc's better diff visualization which we already did

elkoled:
The reason was to have a self-contained diff implementation in opendbc without openpilot deps
In my openpilot diff tool I also reuse process_replay so we could combine both later to reduce duplication.

sshane:
If I change flags for a platform and it shows up here, LGTM

PR #3081, opened 2026-01-28
Car Diff: add ref states

elkoled:
This is a PR to reduce diff for #3070. It is sovereign and does not change any functionality.
It pipes the ref/signal though the signal chain before applying the functionality change in #3070.


elkoled:
@sshane I split this from #3070 for easier review

sshane:
This is fine to keep together since on its own it does not add any functionality or clean anything up

PR #3084, opened 2026-01-29
Fix lack of test coverage for angle_meas scaling in subaru.h

jacobwaller:
Validation
* Dongle ID: 
* Route:

### DIFF
--- a/opendbc/safety/tests/test_subaru.py
+++ b/opendbc/safety/tests/test_subaru.py
@@ -59,7 +59,7 @@ def fwd_blacklisted_addr(lkas_msg=SubaruMsg.ES_LKAS):
   return {SUBARU_CAM_BUS: [lkas_msg, SubaruMsg.ES_DashStatus, SubaruMsg.ES_LKAS_State, SubaruMsg.ES_Infotainment]}
 
 
-class TestSubaruSafetyBase(common.CarSafetyTest):
+class TestSubaruSafetyBase(common.CarSafetyTest, common.AngleScalingSafetyTest):
   FLAGS = 0
   RELAY_MALFUNCTION_ADDRS = {SUBARU_MAIN_BUS: (SubaruMsg.ES_LKAS, SubaruMsg.ES_DashStatus, SubaruMsg.ES_LKAS_State,
                                                SubaruMsg.ES_Infotainment)}
@@ -76,9 +76,6 @@ class TestSubaruSafetyBase(common.CarSafetyTest):
   DEG_TO_CAN = 100
 
   INACTIVE_GAS = 1818
-
-  # About maxed out in one direction
-  STEER_ANGLE_MAX = 545
 
   def setUp(self):
     self.packer = CANPackerSafety("subaru_global_2017_generated")
@@ -113,11 +110,6 @@ class TestSubaruSafetyBase(common.CarSafetyTest):
   def _pcm_status_msg(self, enable):
     values = {"Cruise_Activated": enable}
     return self.packer.make_can_msg_safety("CruiseControl", self.ALT_MAIN_BUS, values)
-
-  def test_steering_angle_measurements(self):
-    self._common_measurement_test(self._angle_meas_msg, -self.STEER_ANGLE_MAX, self.STEER_ANGLE_MAX, self.DEG_TO_CAN,
-                                  self.safety.get_angle_meas_min, self.safety.get_angle_meas_max)
-
 
 class TestSubaruStockLongitudinalSafetyBase(TestSubaruSafetyBase):
   def _cancel_msg(self, cancel, cruise_throttle=0):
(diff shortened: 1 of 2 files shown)


opendbc/safety/modes/subaru.h
```diff
@@ -101,7 +101,7 @@ static void subaru_rx_hook(const CANPacket_t *msg) {
 
     int angle_meas_new = (GET_BYTES(msg, 4, 2) & 0xFFFFU);
     // convert Steering_Torque -> Steering_Angle to centidegrees, to match the ES_LKAS_ANGLE angle request units
-    angle_meas_new = ROUND(to_signed(angle_meas_new, 16) * -2.17);
+    angle_meas_new = ROUND(-2.17 * to_signed(angle_meas_new, 16));
```
jacobwaller, line 104: This change is not needed, but is there to demonstrate that the tests added cover the mutant tests for this line

Feel free to accept this suggestion to undo the change:

```suggestion
    angle_meas_new = ROUND(to_signed(angle_meas_new, 16) * -2.17);
```

opendbc/safety/tests/test_subaru.py
```diff
@@ -111,6 +121,21 @@ def _pcm_status_msg(self, enable):
     values = {"Cruise_Activated": enable}
     return self.packer.make_can_msg_safety("CruiseControl", self.ALT_MAIN_BUS, values)
 
+  def test_angle_meas_scaling(self):
```
sshane: Don't we have a test to make sure `angle_meas_new` is parsed properly in common.py?

jacobwaller: I'm not really sure if that's the intention of those tests as they don't catch the mutation:
Changing only one line to trigger a diff in opendbc master:
```
❯ git diff
diff --git a/opendbc/safety/modes/subaru.h b/opendbc/safety/modes/subaru.h
index 9b01d39d..6871cf06 100644
--- a/opendbc/safety/modes/subaru.h
+++ b/opendbc/safety/modes/subaru.h
@@ -101,7 +101,7 @@ static void subaru_rx_hook(const CANPacket_t *msg) {
 
     int angle_meas_new = (GET_BYTES(msg, 4, 2) & 0xFFFFU);
     // convert Steering_Torque -> Steering_Angle to centidegrees, to match the ES_LKAS_ANGLE angle request units
-    angle_meas_new = ROUND(to_signed(angle_meas_new, 16) * -2.17);
+    angle_meas_new = ROUND(-2.17 * to_signed(angle_meas_new, 16));
     update_sample(&angle_meas, angle_meas_new);
   }
```

Results in the mutation tests failing as before:
```
/Users/jacobwaller/Documents/personal/comma/opendbc/opendbc/safety/modes/subaru.h:104:34: warning: Survived: Replaced * with / [cxx_mul_to_div]
    angle_meas_new = ROUND(-2.17 * to_signed(angle_meas_new, 16));
                                 ^
```

jacobwaller: Oh I see, that steering angle test is only being done for angle-based cars
```
class AngleSteeringSafetyTest(VehicleSpeedSafetyTest):

... other code

  def test_steering_angle_measurements(self):
    self._common_measurement_test(self._angle_meas_msg, -self.STEER_ANGLE_MAX, self.STEER_ANGLE_MAX, self.DEG_TO_CAN,
                                  self.safety.get_angle_meas_min, self.safety.get_angle_meas_max)
```

jacobwaller: Alright @sshane I updated to use the helper in common.py

One note -- I wanted to just add this to all the cars, but not everything has an easily accessible `STEER_ANGLE_MAX` or a way to grab the `angle_meas`. In fact, from what I can see, some cars don't even have `angle_meas` in safety (hyundai.h for example)? I might be missing something there though.

As a workaround, I just added it in test_subaru.py

sshane: If the safety test/mode is for angle, it should subclass from that base angle test

jacobwaller: The case we're looking at here is for torque-based, I'm hesitant to subclass from that because subarus don't support angle based steering yet.

jacobwaller: Hey @sshane , sorry to ping you again :). I refactored the scaling check into a dedicated test (`AngleScalingSafetyTest`) and made both `AngleSteeringSafetyTest` and `TestSubaruSafetyBase` inherit from it.

This should keep these tests a little more DRY, let me know if this is kind of what you were thinking

sshane:
I see we merged a PR to add angle-based signals before we actually added angle safety... https://github.com/commaai/panda/pull/1528 Then we never merged https://github.com/commaai/panda/pull/1533

sshane:
`AngleSteeringSafetyTest` *should* catch this in https://github.com/commaai/opendbc/pull/2864. I'll check out why it doesn't

PR #2878, opened 2025-11-06
Honda: Acura ADX

mvl-boston:
Identical controls to radarless Honda HRV.  Separately obtained 2025 and 2026 fingerprints.

Validation route: f2d6180c6007adc2/00000041--75a01c0fb1

Dashcam route confirmed 3840 max factory steering: ad9840558640c31d/000001c5--f41141955f

Todo:
- [x] initial learned steer parameters (after 30 min - wow this car is good)
- [x] more steer fine tuning after 60 min - see routes.py for routenum
- [x] plug comma power, get more fingerprints for 2025 - ad9840558640c31d/00000009--6c2fb60cf5
- [x] non-comma power fingerprint added for 2026 - ad9840558640c31d/000000a1--d10039197c
- [x] additional comma power fingerprints are unchanged for 2026 - ad9840558640c31d/000000a1--d10039197c
- [x] owner provided test route with official mount

Note: braking fix moved to a [separate PR](https://github.com/commaai/opendbc/pull/2050).   Will not impact release software since gas/brake gated behind dev alphalong.  And fix likely to impact other radarless Hondas.

PR #3089, opened 2026-01-30
Honda: Update to latest community car support

mvl-boston:
Adding more models to community support:

Acura ADX - community forks include code from https://github.com/commaai/opendbc/pull/2878
Acura MDX Hybrid - community forks include code from https://github.com/commaai/opendbc/pull/2671
Honda Accord Hybrid - community forks include code from https://github.com/commaai/opendbc/pull/2671


sshane:
@jyoung8607 we should add links to PRs/branches in the community car docs if you have a simple patch you had in mind

PR #3091, opened 2026-01-30
Toyota: make RAV4 TSS2 use torque control

sshane:
We tried to do this before, but that was before recent improvements (+ torqued)? It was reverted after https://github.com/commaai/openpilot/issues/21707. From my data exploration, the lat accel factors are very similar, enough I would merge them for any other case/platform if this was the only difference. Plus we now have torque control with learning torqued which has improved controls drastically since 4 years ago. So if there is a problem after this, we don't fully understand how torque works for Toyota.

Data from last 3 months.

```
frictions:
normal median 0.1264222413301468 mean 0.1277268213910591
special median 0.1150946244597435 mean 0.11550192771879052
```

```
lat accels:
normal median 2.0320370197296143 mean 2.0384563336594725
special median 1.879465937614441 mean 1.8955343907009194
```

image

image


ghostdev137:
@sshane Feels amazing on my "Special" Rav 4 Hybrid, this resolves the regression that Willem fixed with Zorro, thanks!

sshane:
Glad to hear it!

PR #3092, opened 2026-01-30
Car diff: lazy load

elkoled:
Enable import from openpilot by lazy loading comma_car_segments

### DIFF
--- a/opendbc/car/tests/car_diff.py
+++ b/opendbc/car/tests/car_diff.py
@@ -15,7 +15,6 @@ from collections import defaultdict
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


elkoled:
@sshane ready

sshane:
what's different with this change?

elkoled:
it is needed to fix comma_car_segments import error when using car_diff from openpilot.

sshane:
What is the error? Is there a PR?

elkoled:
Here it is:
https://github.com/commaai/openpilot/pull/37048

sshane (approved):

PR #3094, opened 2026-01-31
Subaru global: use universal angle signal for angle cars

sshane:
For https://github.com/commaai/opendbc/pull/2864

I think we can swap to Steering_2 for all globals if it matches on the other platforms, just to keep the grouping sane. `pre-global` vs `global` instead of `preglobal + global - angle cars` vs. `angle cars`


sshane:
Damn, Ascent and Impreza are missing this message despite being global :( Then we need to split by angle cars for now like you had @jacobwaller

jacobwaller:
> Damn, Ascent and Impreza are missing this message despite being global :( Then we need to split by angle cars for now like you had @jacobwaller

I do know the '25 Impreza works with this new message FYI. But yeah, it's annoying how many changes they've had

PR #3096, opened 2026-01-31
Subaru ANGLE_LKAS: fix cruise control signals

sshane:
Check:

- [x] SUBARU_FORESTER_2022
- [x] SUBARU_OUTBACK_2023
- [x] SUBARU_ASCENT_2023

---

Forester ES_DashStatus going low during user gas, plus current CruiseControl signals always being zero! ES_Status and ES_Brake are exactly the same, both looking good

image

---

Outback showing old CruiseControl at zero while ES_DashStatus and ES_Brake are activated and matching:

image

---

Ascent showing DashStatus falling to 0 at standstill + current CruiseControl signal always zero as well

image

### DIFF
--- a/opendbc/car/subaru/carstate.py
+++ b/opendbc/car/subaru/carstate.py
@@ -81,10 +81,12 @@ class CarState(CarStateBase):
     ret.steeringPressed = abs(ret.steeringTorque) > steer_threshold
 
     cp_cruise = cp_alt if self.CP.flags & SubaruFlags.GLOBAL_GEN2 else cp
+    cp_es_brake = cp_alt if self.CP.flags & SubaruFlags.GLOBAL_GEN2 else cp_cam
+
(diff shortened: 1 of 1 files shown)


sshane:
This is SUBARU_FORESTER_2022. It looks like `ES_DashStatus` will not work as it says ACC deactivates when user presses gas and goes back on when they release. It's likely a signal of whether it's applying gas itself.

But ES_Brake or ES_Status may work. Have to check the hybrids to see if this signal was always wrong

image

sshane:
Looks like I found the same issue when we first merged hybrid support but the route (https://github.com/commaai/openpilot/pull/25378#issuecomment-1682856769) never had a gas override while engaged, so it likely masked the bug and my found issues were ignored.

Noted here in that same PR: https://github.com/commaai/openpilot/pull/25378#issuecomment-1228971789

PR #3100, opened 2026-01-31
Subaru: remove unused safety angle measurement

sshane:
We switched to using Steering_2 which is actually available for all LKAS angle cars (specifically newer ones), so this is not needed (and never was tested). Should resolve mutation failure in https://github.com/commaai/opendbc/pull/2864


sshane:
seems like I'm merging too fast for the bot @elkoled, old refs?

PR #3101, opened 2026-02-01
Mazda: document blocking LKAS engagement 

nikitap066:
Document and resolve TODO to clarify Mazda LKAS engagement blocking when lane lines are unavailable.

No functional changes.

Validation
Dongle ID: N/A
Route: N/A

### DIFF
--- a/opendbc/car/mazda/carstate.py
+++ b/opendbc/car/mazda/carstate.py
@@ -87,7 +87,7 @@ class CarState(CarStateBase):
     ret.cruiseState.speed = cp.vl["CRZ_EVENTS"]["CRZ_SPEED"] * CV.KPH_TO_MS
 
     # stock lkas should be on
-    # TODO: is this needed?
+    # LKAS requires detected lane lines; block engagement if none are present
     ret.invalidLkasSetting = cp_cam.vl["CAM_LANEINFO"]["LANE_LINES"] == 0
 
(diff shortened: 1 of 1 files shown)


sshane:
I'm still not sure how this works. Why can't we send LKAS messages unless stock LKAS is on?

nikitap066:
I believe the EPS will only accept steering assist when the camera/EPS agree LKAS is allowed. If stock LKAS isn’t available, sending torque is either ignored, faulted, or unsafe.

sshane:
Is there a signal we can send to the EPS to tell it LKAS is on? Fake the user pressing the button

opendbc/car/mazda/carstate.py
```diff
@@ -87,7 +87,7 @@ def update(self, can_parsers) -> structs.CarState:
     ret.cruiseState.speed = cp.vl["CRZ_EVENTS"]["CRZ_SPEED"] * CV.KPH_TO_MS
 
     # stock lkas should be on
-    # TODO: is this needed?
+    # LKAS requires detected lane lines; block engagement if none are present
```
sshane, line 90: this reads as if you need actual lane lines on the road or you can't use openpilot? does it not work on roads without two clear lane lines?

sshane:
Closing since this comment doesn't fully explain what's happening here

PR #3107, opened 2026-02-03
test: lazy load car interfaces to speed up collection

rob0tics:
## Description
This PR addresses the slow pytest collection time in `test_car_interfaces.py`. By moving heavy imports (specifically car interfaces and firmware metadata) inside the test functions, we eliminate the "Import Tax" during the collection phase.

## Performance Impact
- **Before:** ~20s to collect tests.
- **After:** **0.18s** to collect 240 tests (measured on macOS M-series).
- **Total Run Time:** Significantly reduced overhead for CI and local development.

## Context
This fix addresses the bottleneck identified in **commaai/openpilot#32536**.

Fixes commaai/openpilot#32536

### DIFF
--- a/opendbc/car/tests/test_car_interfaces.py
+++ b/opendbc/car/tests/test_car_interfaces.py
@@ -1,7 +1,6 @@
 import os
 import math
 import pytest
-import importlib
 from functools import cache
 from hypothesis import Phase, given, settings
 import hypothesis.strategies as st


sshane:
Collection goes from 0.9s to 0.2s on my machine, not 100x speedup like you show. If the end result test is not any faster, there's not much point in merging this. Can we figure out how to speed up the car interface loop? That's the biggest time usage

rob0tics:
> Collection goes from 0.9s to 0.2s on my machine, not 100x speedup like you show. If the end result test is not any faster, there's not much point in merging this. Can we figure out how to speed up the car interface loop? That's the biggest time usage

I profiled the loop and discovered that the overhead isn't in the 10-20 iterations of update()/apply(), but in the CarInterface initialization and DBC parsing that happens for every Hypothesis example. Since we can't cache the Interface without losing the fuzzing coverage on the init phase itself, reducing the iterations remains the most effective way to shave time off the execution phase without a massive architectural rewrite of how DBCs are loaded.

PR #3109, opened 2026-02-04
safety: rafactor libsafety

tknsgk:
## Summary
- Remove redundant forward declaration of `microsecond_timer_get()`
- Wrap debug printf in `#ifdef DEBUG` for conditional compilation

## Test plan
- [x] Verify preprocessor output without DEBUG flag excludes printf
- [x] Verify preprocessor output with `-DDEBUG` includes printf


opendbc/safety/tests/libsafety/safety.c
```diff
@@ -26,7 +25,9 @@ bool safety_config_valid() {
     const RxCheck addr = current_safety_config.rx_checks[i];
     bool valid = addr.status.msg_seen && !addr.status.lagging && addr.status.valid_checksum && (addr.status.wrong_counters < MAX_WRONG_COUNTERS) && addr.status.valid_quality_flag;
     if (!valid) {
-      // printf("i %d seen %d lagging %d valid checksum %d wrong counters %d valid quality flag %d\n", i, addr.status.msg_seen, addr.status.lagging, addr.status.valid_checksum, addr.status.wrong_counters, addr.status.valid_quality_flag);
+#ifdef DEBUG
```
tknsgk, line 28: **Without DEBUG flag (printf excluded):**
```
$ gcc -I. -E opendbc/safety/tests/libsafety/safety.c | grep -A5 'if (!valid)'
if (!valid) {
return
```

**With `-DDEBUG` flag (printf included):**
```
$ gcc -I. -DDEBUG -E opendbc/safety/tests/libsafety/safety.c | grep -A5 'if (!valid)'
if (!valid) {
printf("i %d seen %d ...")
return
```

You don’t need to comment and uncomment this each time.

opendbc/safety/tests/libsafety/safety.c
```diff
@@ -4,7 +4,6 @@
 
 // TODO: time should just be passed into the hooks we expose
 uint32_t timer_cnt = 0;
-uint32_t microsecond_timer_get(void);
```
jyoung8607, line 7: It's possible this is now redundant, but it's one of a few declarations that needs to be maintained with great care between commaai/panda and commaai/opendbc safety. It has to build with both GCC and clang, it has to build in very different ways for opendbc CI or Panda STM32 environments, and it has to pass MISRA testing for both. Passing CI in opendbc alone isn't sufficient evidence we can remove it.

opendbc/safety/tests/libsafety/safety.c
```diff
@@ -26,7 +25,9 @@ bool safety_config_valid() {
     const RxCheck addr = current_safety_config.rx_checks[i];
     bool valid = addr.status.msg_seen && !addr.status.lagging && addr.status.valid_checksum && (addr.status.wrong_counters < MAX_WRONG_COUNTERS) && addr.status.valid_quality_flag;
     if (!valid) {
-      // printf("i %d seen %d lagging %d valid checksum %d wrong counters %d valid quality flag %d\n", i, addr.status.msg_seen, addr.status.lagging, addr.status.valid_checksum, addr.status.wrong_counters, addr.status.valid_quality_flag);
+#ifdef DEBUG
+      printf("i %d seen %d lagging %d valid checksum %d wrong counters %d valid quality flag %d\n", i, addr.status.msg_seen, addr.status.lagging, addr.status.valid_checksum, addr.status.wrong_counters, addr.status.valid_quality_flag);
+#endif
```
jyoung8607, line 30: No real value in adding a brand-new DEBUG #define for the very occasional use this would see, especially one that's easily confused with ALLOW_DEBUG.

Issue #1147, opened 2024-04-15
Rumbling while collaboratively steering at low speeds


voximity:
### Describe the bug

At low speeds (around 15 MPH), my steering wheel will "rumble" if I collaboratively steer alongside lateral input from openpilot. To be precise, this is not the kind of vibration/rumbling that Hondas give the driver to indicate a warning (like lane departure), but almost feels as if I am going over a rumble strip. More accurately, it sounds like I'm fighting the torque actuator -- or something.

### Which car does this affect?

Honda CR-V 2020

### Provide a route where the issue occurs

e4b43f41460a7547/00000001--02564f7b2d/1

### openpilot version

27c26c2 (master-ci)

### Additional info

I initially encountered this issue on Sunnypilot when using its separated LKAS / ACC feature. My CR-V's stock ACC does not allow me to set the cruise below 25 MPH, so I was able to reproduce this on stock OP using OP long and setting my max speed to 15 MPH. In the provided route, the issue happens a lot as I zig-zag through a street near the very end (immediately preceding the bookmarks at the end).

*Possibly* related to commaai/opendbc#1112, except this feels *and* sounds like a rumble, rather than just being an electric humming sound.

jyoung8607:
> To be precise, this is not the kind of vibration/rumbling that Hondas give the driver to indicate a warning (like lane departure), but almost feels as if I am going over a rumble strip. More accurately, it sounds like I'm fighting the torque actuator -- or something.

At those times, your EPS motor output includes a perfect 10Hz waveform of decent amplitude. It's extremely likely this is an EPS haptic warning. Not quite sure what the trigger is yet.

<img width="3147" height="2130" alt="Image" src="https://github.com/user-attachments/assets/fa4bd037-4425-4be3-aa9c-700c16fd8cc7" />

<img width="3147" height="2130" alt="Image" src="https://github.com/user-attachments/assets/a13bcfbe-a176-41fd-9bfc-801863e3d719" />

jyoung8607:
The [2020 Honda CR-V owner's manual](https://cdn.dealereprocess.org/cdn/servicemanuals/honda/2020-crv.pdf) shows three causes of tactile warnings:
* Lane Keeping Assist System
* Road Departure Mitigation
* Driver Attention Monitor

For LKAS and RDM, I'm not immediately able to find anything interesting coming from the camera, and we're outside the documented speed envelope for either of them to activate.

[...]

PR #3114, opened 2026-02-06
Honda - cutoff LKAS below minSteerSpeed

mvl-boston:
Fix for https://github.com/commaai/opendbc/issues/1147

Had to increase CRV_5G minsteer speed to 16.5mph to prevent haptic feedback at lower speeds

See carstate.steeringRateDeg for Haptic warning
- Before: 0b3cfa082cbbe54e/00000002--9a41a91161/55:60
- After: 0b3cfa082cbbe54e/00000006--138313f050

Also per conversation set actuatorsOutput.torque to not build up integral and clean up the torque ball.


opendbc/car/honda/carcontroller.py
```diff
@@ -152,7 +153,10 @@ def update(self, CC, CS, now_nanos):
         can_sends.append(make_tester_present_msg(0x18DAB0F1, 1, suppress_response=True))
 
     # Send steering command.
-    can_sends.append(hondacan.create_steering_control(self.packer, self.CAN, apply_torque, CC.latActive, self.tja_control))
+    send_lkas = CC.latActive
+    if CS.out.vEgo <= self.CP.minSteerSpeed and self.CP.carFingerprint in (HONDA_LKAS_MINSPEED_CUTOFF):
+      send_lkas = False
+    can_sends.append(hondacan.create_steering_control(self.packer, self.CAN, apply_torque, send_lkas, self.tja_control))
```
jyoung8607:
```suggestion
    lkas_active = CC.latActive and CS.out.vEgo >= self.CP.minSteerSpeed

    can_sends.append(hondacan.create_steering_control(self.packer, self.CAN, apply_torque, lkas_active, self.tja_control))
```

jyoung8607 (changes requested):
I agree there's problems here. This was something I hoped to clean up during Hondapalooza last year but didn't have time. Honda is one of the oldest openpilot ports, and its (lack of) proper min steer speed handling reflects that.

For this PR, none of the HondaFlags logic should be necessary. I don't think there's ever any reason to set `STEER_TORQUE_REQUEST` while below CP.minSteerSpeed. We can just make that condition global. That much is simple, and I would merge it. Make sure we're reflecting that in the vehicle HUD as well.

The challenge is that many Honda models won't have CP.minSteerSpeed set. If you see `minSteerSpeed` set on CarSpecs (like HONDA_CITY_7G in your PR) this change will apply. If you see it specified as `min_steer_speed` on HondaCarDocs (like HONDA_CRV_5G in your PR) that's going to show up in CARS.md and the shop, but not actually apply at runtime.

This weirdness dates back before FPv2, when we had trouble figuring out which cars had limits and which didn't at fingerprint time. This is an annoying Honda-only thing and [we'd all love for all the code around this Honda-only docs case to go away](https://github.com/commaai/opendbc/pull/2309).

Separately, in the future, we can look at PRs migrating individual models away from CarDocs min_steer_speed so they actually work right. Those will be easy PRs to write but data-intensive to validate; we ideally want to keep steer-to-zero for any cars that have it today.

opendbc/car/honda/carcontroller.py
mvl-boston: Pending fixing carspecs

mvl-boston: Just fixed carspecs for CRV, only other car with Haptic warning that required cutoff.

opendbc/car/honda/carcontroller.py
jyoung8607, line 143: I forgot to mention, you're going to want to zero out apply_torque whenever lkas_active is false. Two reasons: first, safety should stop you from sending non-zero torque in that case, not sure if Honda has that enabled. second, upper layer controls needs to know if output is limited by the car port, to avoid integral wind-up when rate-limited or disabled, and the final apply_torque is saved for that purpose.

mvl-boston: Zeroing out apply torque is already done in hondacan:
https://github.com/commaai/opendbc/blob/master/opendbc/car/honda/hondacan.py#L119

Doesn't look like the pid controller looks at apply_torque any more?
https://github.com/commaai/openpilot/blob/master/selfdrive/controls/lib/latcontrol_torque.py#L74

jyoung8607: That explains how messages are flowing through safety, but doesn't help the lateral accel torque controller because we [didn't let it know](https://github.com/commaai/openpilot/blob/c393973916de984ae2444383e1de5685505f67ce/selfdrive/controls/lib/latcontrol_torque.py#L92) that we're [clipping actuator output in the port](https://github.com/commaai/openpilot/blob/c393973916de984ae2444383e1de5685505f67ce/selfdrive/controls/controlsd.py#L175). The [legacy PIF controller](https://github.com/commaai/openpilot/blob/c393973916de984ae2444383e1de5685505f67ce/selfdrive/controls/lib/latcontrol_pid.py#L35) responds the same way.

Looks like in Honda's case it's currently using self.last_torque, sorry, I'm used to the way VW does it. Unfortunately this is very old code and you've inherited some problems. I think this is why I left it alone last fall, fixing it was going to be a Whole Thing.

~~Upper layer controls currently don't look at CP.minSteerSpeed, they rely on the car port to zero out the returned actuator and the Honda port doesn't, even though it does so on-the-wire. That's no bueno.~~

~~With current master, no matter which controller you're using, I think you may be winding up integral anytime you're engaged below minSteerSpeed and don't have enough driver input torque to set `steerPressed`. That probably results in a nice yank on the wheel when they accelerate past minSteerSpeed. The `torqued` learner doesn't even check the actuator clipped condition, so it's got even worse problems.~~

I'm down to work through this stuff, just saying there's even more bugs than you may know.

mvl-boston: Thanks, I wanted to make sure I was touching the right values, so I had traced the code.   Easy to miss this:
- I saw steer_limited_by_safety, but of course didn't expect that to be "limited by car_port", but I understand now that every other brand probably has too much torque and had to cut to 0.9m/s2 in the car port.
- Comments at https://github.com/commaai/opendbc/blob/master/opendbc/car/car.capnp#L372 also threw me off, but I see that they don't apply to torque.   (Should have realized as I also stole this value in one of mytestforks for my tuning logging and had a C4 driver complain that the torque ball went crazy.   I'll steal an unused Honda value like angle_accel going forward).

I think it was an easy fix here, let me know if this is missing anything else now.

Presume that torqued learner is not within scope of a single car brand.   I could write a PR for it separately, but frankly doubt I have enough replay experience to be able to test it to my satisfaction, let alone comma's satisfaction.

jyoung8607: For torqued, TBH I think we should have upper layer controls start respecting CP.minSteerSpeed. Starting a dialogue about it in Discord #dev-openpilot. Either way it's still correct to clean up the car port such that it reports its actual output correctly.

mvl-boston: None of you guys sleep.   Just saw the discussion and simplified by backing out the lkas_active field.

Explains the mystery of why the Brazil City driver never complained of haptic feedback.

jyoung8607: I stand corrected from @sshane, we do disengage latActive when below minSteerSpeed, it's just confusingly labeled as a standstill check. It's still correct to fix actuator output reporting here at the car port.

mvl-boston: Agreed - updated code, let me know if anything further needed.

PR #3125, opened 2026-02-15
safety: move CAN ignition hook from panda

### LINKED ISSUE #1834: Move CAN ignition hooks and add test

nexicturbo:
Fixes #1834.

- Move CAN-based ignition detection (GM 0x1F1, Rivian 0x152, Tesla 0x221, Mazda 0x9E) into opendbc safety code.
- Add reset + libsafety bindings.
- Add CI unit tests to validate correct triggering and guard against false positives from non-sequential counters.

### DIFF
--- a/opendbc/safety/declarations.h
+++ b/opendbc/safety/declarations.h
@@ -263,6 +263,12 @@ extern bool acc_main_on; // referred to as "ACC off" in ISO 15622:2018
 extern int cruise_button_prev;
 extern bool safety_rx_checks_invalid;
 
+// CAN-based ignition detection (used by panda)
+extern bool ignition_can;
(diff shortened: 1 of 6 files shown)


genesis-ai-labs-star:
Ran local ./test.sh on current master to confirm baseline: all good (4529 passed, 1764 skipped). If you want I can also pull your branch and verify tests there.

PR #3126, opened 2026-02-15
Subaru: fix EPS torque mismatch fault prevention

mpurnell1:
Similar to #2859, certain models of Subaru are experiencing EPS faults during sustained turns, causing steering to be temporarily disabled.

I initially suspected this was the same regression as the Toyota PR, where the original fix (commaai/openpilot#29116) predates the breakup of controlsd into three separate services, adding an extra frame of reaction time. However, analysis of the fault events revealed the EPS can't be tracking steering rate directly, since one fault occurred with zero steering rate and the wheel stationary for 0.6s.

The real fault condition is the integral of torque mismatch (|EPS output| - |driver torque|) over a ~3s sliding window, with a consistent threshold of ~3900 across all analyzed faults. This resolves the FIXME in the code: "These limits aren't exact. The real limit is more than likely over a larger time period and involves the total steering angle change rather than rate"

Changes:
- Replace rate-based common_fault_avoidance trigger with torque mismatch
integral over a 3s sliding window
- Linearly scale down commanded torque as the integral approaches the fault
threshold (soft limit at 2500, hard limit at 3500)
- Keep common_fault_avoidance as a backup safety net using the mismatch
integral condition
- Decrease min_valid_request_frames from 7 to 6 to match, same as #2859 did
for Toyota

Affected models/years:
- '20-'22 Outback SUBARU_OUTBACK (logs below)
- '19-'21 Forester SUBARU_FORESTER (logs below)

Failing '21 Forester routes
```
fca4fccad2e7b029/00000026--232120d429/10/12
9fb5e8c2cb1265f8/00000089--458b208471/2
9fb5e8c2cb1265f8/00000089--458b208471/7
9fb5e8c2cb1265f8/00000088--88ea6ce1c9/31
```

Failing '20 Outback routes
```
52f064b903b83e62/00000000--6371ad187a/1
52f064b903b83e62/00000001--585af3f4ab/2
```

This PR is WIP until I can get some confirmed successful routes using this
code.

### DIFF
--- a/opendbc/car/subaru/carcontroller.py
+++ b/opendbc/car/subaru/carcontroller.py
@@ -1,4 +1,5 @@
 import numpy as np
+from collections import deque
 from opendbc.can import CANPacker
 from opendbc.car import Bus, make_tester_present_msg
 from opendbc.car.lateral import apply_driver_steer_torque_limits, common_fault_avoidance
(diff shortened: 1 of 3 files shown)


wtogami:
Nice you figured out that long standing FIXME!

mpurnell1:
After doing further research on this, I have found that this is not a correct model for the fault. It appears to be closer to requesting steering when the wheel is turned beyond ~90 degrees with no driver torque detected. I will consider opening a new PR for that in the future, once I have completed further testing.

wtogami:
FWIW something later changed where the bug fixed in https://github.com/commaai/opendbc/pull/2859 came back for Toyotas. That same change is likely affecting your testing as well.

PR #3127, opened 2026-02-15
VW: Rename aux bus to alt

tealtwo:
The VW carport by default maps internal panda bus 1 to "aux", when this is added to the CANParser in carstate it causes a crash unless it is defined here, another solution would be just renaming the bus inside values.py if we don't want to update init.

image

### DIFF
--- a/opendbc/car/__init__.py
+++ b/opendbc/car/__init__.py
@@ -80,7 +80,6 @@ DbcDict = dict[StrEnum, str]
 
 class Bus(StrEnum):
   pt = auto()
-  aux = auto()
   cam = auto()
   radar = auto()
   adas = auto()


jyoung8607 (commented):
I agree this is a bug. I added it in #2582 but mostly as self-documenting boilerplate. As you've realized, we're not currently using it. What are you planning to use it for?

If you're working on something you plan to upstream, tell us more about your intentions. To maintain compatibility between gateway and camera harnesses, we have to be very thoughtful about adding dependencies of any sort on that bus.

If you're working on something you don't plan to upstream, let's just change the VW port to use `alt` instead of `aux` since it's on that list already. I would merge that, and then you can move on with your work.

tealtwo:
> I agree this is a bug. I added it in #2582 but mostly as boilerplate. As you've realized, we're not currently using it. What are you planning to use it for?
> 
> If you're working on something you plan to upstream, tell us more about your intentions. To maintain compatibility between gateway and camera harnesses, we have to be very thoughtful about adding dependencies of any sort on that bus.
> 
> If you're working on something you don't plan to upstream, let's just change the VW port to use `alt` instead of `aux` since it's on that list already. I would merge that, and then you can move on with your work.

Some MLB vehicles (Audi Q5) only have door signals over bus 1, also in general I think even though it's not being used in stock at the moment, if someone wants to call parse it in a fork, this has to be fixed manually every time, refactoring aux to alt works just as well.

PR #3128, opened 2026-02-15
VW: map OP FCW alert to VW cluster

tealtwo:
This PR maps openpilots forward collision warning output from the model to the car's "Apply Brake!" HUD message.


jyoung8607:
Is this tested to actually work? Are you sure this triggers more than an audible beep? I don't have a route with a FCW handy. I feel like there were more signals set.

tealtwo:
> Is this tested to actually work? Are you sure this triggers more than an audible beep? I don't have a route with a FCW handy. I feel like there were more signals set.

Yes, I have tested on both PQ and MQB. It is used by the stock radar to command take over incase of cruiseFault or take over at a stop (PQ/MQB), it sends "APPLY BRAKE!" along with an audible alert to the cluster, I can send video / logs later today.

PR #3129, opened 2026-02-15
MLB: Steering timeout bomb workaround

Dennis-NL:
Fixes the HCA timeout mitigation for MLB platform
Has been tested and driven over 10.000km with it, no issues found.

Tested on Porsche Macan and Audi Q5

### DIFF
--- a/opendbc/car/volkswagen/carcontroller.py
+++ b/opendbc/car/volkswagen/carcontroller.py
@@ -32,10 +32,19 @@ class CarController(CarControllerBase):
     self.hca_frame_timer_running = 0
     self.hca_frame_same_torque = 0
 
+    # EPS timer reset workaround for MLB platforms (Porsche Macan, Audi, etc.)
+    # MQB racks reset the timer after a single frame of HCA disabled.
+    # MLB racks need > 1 second to reset; we try to reset when
+    # engaged for a long time and torque output is currently low.
+    self.eps_timer_workaround = bool(CP.flags & VolkswagenFlags.MLB)
+    self.hca_frame_timer_resetting = 0
+    self.hca_frame_low_torque = 0
+
   def update(self, CC, CS, now_nanos):
     actuators = CC.actuators
     hud_control = CC.hudControl
     can_sends = []
+    output_torque = 0
 
     # **** Steering Controls ************************************************ #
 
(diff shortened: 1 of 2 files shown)


jyoung8607:
As we've discussed, I agree the basic concept is sound, and I'm glad to hear it's working for you. I'm willing to merge solid PRs that move us toward those goals. This PR makes VW carcontroller update into something that's hard to understand at a glance, especially since there's several timers at work here, and I'd like to go a different way.

Here's a concept sequence of PRs I planned to work on, and would merge if others get to it first.

**PR 1:** Without changing the logic, refactor the VW HCA mitigation logic off into a standalone function, just as it exists today. Here's an [unfinished example](https://github.com/commaai/opendbc/commit/c3cdfeada81c9151d2afc28250fdad769bf8460a). I think I'd actually push this into its own class, to make CI testing easier without instantiating all of CarInterface and to move all the statekeeping variables out of carcontroller scope. This stage should pass without any process replay or car interface replay changes.

**PR 2:** Update the new function to implement your new timer-reset logic. Do this for ALL cars unconditionally, but set the timer for PQ/MLB to 1.1 seconds as you have it, and set the timer for MQB to 0.05 seconds. This will change our behavior, and thus replay outcomes. It will need to come with CI tests that probe all the conditions we might encounter, but this should be easy with the mitigation in a standalone function.

**PR 3:** Implement the [`steerTimeLimit`](https://github.com/commaai/openpilot/blob/4166c9fccbf1fbfbaba2da1964f15585d24145a3/selfdrive/selfdrived/events.py#L754-L757) alert, see commaai/opendbc#1315. The driver needs to know if we don't get a reset opportunity. Ideally this never happens, but the notification itself will serve as telemetry for the MTBF report. We'll know how often it fails in the field, we'll have qlogs to help retune, and CI tests to let us iterate with confidence.

**PR 4 / PR 5:** Once we have clean telemetry for awhile, take MLB/PQ out of dashcam. They'll have to pass comma safety review, but we'll have credible data showing a release-quality driving experience.

We once joked the MQB mitigation would always work because nobody would ever turn left for six minutes. Telemetry showed us this actually happens in the real world, on very straight highways with a strong bank or crosswind. We never dealt with it, and that's why we're going to cover MQB with MLB/PQ-style mitigation. I expect we're going to learn things about the MLB/PQ mitigation too.

Closing this one, but you're welcome to open a new one with a different approach.

PR #3131, opened 2026-02-16
VW: add dynamic comfort signals

tealtwo:
This PR aims to add dynamically calculated comfort signals for PQ & MQB when using openpilot longitudinal control for a smoother experience and to more closely match the behavior of Stock ACC. It also adds some missing signals to the accel controller for PQ to properly control longitudinal. The result should be much smoother acceleration, braking, and allow proper coasting so minor brake commands don't use the ESP pump to generate brake pressure as often, and prefer engine braking in those scenarios.


jyoung8607:
What is this intended to accomplish? 

How did you validate it? 

Before and after routes?

tealtwo:
> What is this intended to accomplish? 
> 
> 
> 
> How did you validate it? 
> 
> 
> 
> Before and after routes?

It is the dynamic jerkLimit and comfort band for the longitudinal controls actuator to smooth it like the stock acc radar. I can provide logs. It is intended to do it similar to stock acc. Anything specific you are looking for in validation, I can provide logs for Stock ACC vs current and PR'd edits.

jyoung8607:
> It is the dynamic jerkLimit and comfort band for the longitudinal controls actuator

I can see that much, and something like this may be appropriate, but we've not yet been shown it's helpful.

> I can provide logs.

Logs are definitely required, yes.

> Anything specific you are looking for in validation, I can provide logs for Stock ACC vs current and PR'd edits.

The values in master today are placeholders I made up. I'm sure they can be improved on. I'm looking for evidence that shows this code is an improvement. It's not obvious from code alone how the car will respond.

In this case, "improvement" would be measurably better conformance with openpilot's longitudinal plan. Running the [longitudinal maneuver test and report](https://github.com/commaai/openpilot/tree/master/tools/longitudinal_maneuvers), before and after the change, would be an excellent start.

tealtwo:
updates are coming. these arbitrary values will be replaced with proper tuning.

PR #3137, opened 2026-02-17
HKG: Add FW for 2022 Hyundai Tucson

ricksticks1:
Added-->  b'\xf1\x00NX4 FR_CMR AT USA LHD 1.00 1.00 99211-CW000 14M',
 to CAR.HYUNDAI_TUCSON_4TH_GENERATION firmware.

Validation
* Dongle ID: 693273e1705f8fed
* Route: 693273e1705f8fed/00000001--064690ee00


jyoung8607:
FYI this isn't being ignored, but much like #2374 there's a systemic conflict between Tucson and Santa Cruz that prevents us from merging it right away. We'll keep this around until it's fixed.

Setting to Draft for now.

PR #3145, opened 2026-02-21
mazda: filter false steerFaultTemporary at standstill (2022+ CX-5)

zephleggett:
## Motivation
Users with 2022+ Mazda CX-5s experience unneeded 'Place Hands on Wheel' alerts during stop and go traffic when cruise engaged. The alert only happens at standstill speeds so it's more an annoyance than helping the user stay safe.

## Details
The 2022+ Mazda CX-5 has `minSteerSpeed=0`, so openpilot engages steering from 0 kph.
However, the EPS asserts `LKAS_BLOCK` at standstill as a low-speed lockout (not actually a fault).
This causes every stop-and-go cycle to trigger a false `steerFaultTemporary` alert.

Older Mazdas never hit this because they don't engage LKAS below 45 kph — `LKAS_BLOCK` is already clear by the time steering activates.

### How the filtering works
- `lkas_init_complete` starts as `False` at every standstill
- When the car moves and `LKAS_BLOCK` clears for the first time, `lkas_init_complete` flips to `True`
- From that point on, if `LKAS_BLOCK` comes back while driving, it's treated as a real fault
- When the car stops again, the flag resets and the cycle repeats

This means the EPS low-speed lockout at standstill is ignored, but any real steering fault while driving is still caught.

VW uses the same pattern (`eps_init_complete`) to filter EPS init faults before the HCA module is ready.

### Testing Notes
I've been testing this on real drives for a couple day. I have also replayed old rlogs through Mazda CarState to verify that any real  `steerFaultTemporary` are not filtered.


zephleggett:
I've tested this for a couple days now and everything is working as expected.

zephleggett:
I added a safety timer (500 frame) incase we enter fault during standstill and `LKAS_BLOCK` never goes to true.
This ensures this fix will never cause a real alert to be suppressed even during faults.

yummydirtx:
Tested this on a few real drives in my Mazda CX-5, seems to successfully eliminate the annoying and frequent pings whenever the car reached a standstill in traffic on ACC.

chinna244:
I tried this out on my Mazda CX-5, and I’m super impressed! It works really well!

zephleggett:
@adeebshihadeh Sorry to bug you, but I think this change would make the experience much better for our Mazda drivers. Please let me know if I can assist or change anything. thanks for your time! :3

yummydirtx:
Seconding on what @zephleggett said, this is a needed QOL change, hope for review or merge soon!

zephleggett:
I'll get some before and after segments to post :)

sdeban:
This sounds great! I've had the same issue. Sorry if this is a naive question, but how do I get this fix on my comma four?

PR #3161, opened 2026-02-27
HKG: add KIA_SPORTAGE_5TH_GEN (NQ5) to CANFD_FUZZY_WHITELIST

syrull:
## Description

Adds `CAR.KIA_SPORTAGE_5TH_GEN` to `CANFD_FUZZY_WHITELIST` in `opendbc/car/hyundai/values.py`.

## Problem

The KIA Sportage 5th Gen (NQ5) is a CAN-FD car that is currently blacklisted from fuzzy fingerprinting. Cars with firmware versions not present in the exact fingerprint database cannot be matched at all — both exact and fuzzy matching fail, resulting in "unsupported car" on the device.

## Fix

Adding the NQ5 to `CANFD_FUZZY_WHITELIST` enables platform code-based fuzzy matching. This is safe because:

- Both ICE and Hybrid variants share the same platform (`KIA_SPORTAGE_5TH_GEN`) and platform code (`NQ5`)
- The same approach is already used for other Kia CAN-FD platforms (Sorento 4th Gen, K8, Carnival)

## Testing

Tested on a KIA NQ5 Sportage that was previously showing as unsupported.


adeebshihadeh:
This needs more validation beyond a single car fingerprinting, as the issue is more false positives rather than false negatives.

PR #3167, opened 2026-03-03
volkswagen: fix brake pressure normalization and add EPB parking brak…

Dragoadri:
Description
Two FIXMEs in the VW MQB carstate:

ret.brake (MQB): ESP_Bremsdruck has offset=-30 in the DBC, which can produce negative values before the first CAN message. Added max(0, ...) clamp. The /250.0 normalization was already correct (maps Bar pressure to 0.0–1.0 as expected by CarState).
ret.parkingBrake (MQB): Added EPB check via EPB_01.EPB_Status == 1 (brake closed). EPB_01 is defined in both MQB and MQBEvo DBCs. On cars without EPB the signal defaults to 0 ("offen"), so no regression.
ret.brake (PQ): The /250.0 was already correct (BR5_Bremsdruck has offset=0, range [0, 250] Bar). Removed the FIXME comment.

Verification
Verified against DBC signal definitions in vw_mqb_2010.dbc and vw_pq35.dbc.
Validation

Dongle ID: N/A — no hardware available
Route: N/A — changes verified against DBC signal definitions (ESP_Bremsdruck offset=-30, EPB_01.EPB_Status values 0=offen, 1=closed, 2=actuating, 3=error)

### DIFF
--- a/opendbc/car/volkswagen/carstate.py
+++ b/opendbc/car/volkswagen/carstate.py
@@ -79,11 +79,11 @@ class CarState(CarStateBase):
       if self.CP.flags & VolkswagenFlags.STOCK_HCA_PRESENT:
         ret.carFaultedNonCritical = bool(cam_cp.vl["HCA_01"]["EA_Ruckfreigabe"]) or cam_cp.vl["HCA_01"]["EA_ACC_Sollstatus"] > 0  # EA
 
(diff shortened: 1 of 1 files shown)


tealtwo:
I don't see how this functionally fixes anything beside the EPB check, that is useful, and keep in mind that PRs for two separate unrelated fixes should be made separately.

PR #3173, opened 2026-03-04
Nissan: cluster speed alignment fix

Sathans:
Fix speed mismatch with cluster and can message.
Offset is moved to dbc file.

### DIFF
--- a/opendbc/car/nissan/carstate.py
+++ b/opendbc/car/nissan/carstate.py
@@ -84,7 +84,7 @@ class CarState(CarStateBase):
       else:
         conversion = CV.MPH_TO_MS if cp.vl["HUD"]["SPEED_MPH"] else CV.KPH_TO_MS
       ret.cruiseState.speed = speed * conversion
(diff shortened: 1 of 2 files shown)


Sathans:
Do we still need `ret.cruiseState.speedCluster = speed * conversion` if there is no offset needed?

Sathans:
Removed redundant speedCluster

opendbc/car/nissan/carstate.py
```diff
@@ -110,8 +109,7 @@ def update(self, can_parsers) -> structs.CarState:
     can_gear = int(cp.vl["GEARBOX"]["GEAR_SHIFTER"])
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
-    # stock lkas should be off
-    # TODO: is this needed?
```
Sathans, line 114: Turning on stock lkas shows warning about keeping hand on steering wheel

Sathans:
bump

PR #3224, opened 2026-03-24
Increase KIA EV6 torque

elkoled:
Max lat accel measured (roll compensated): -2.3m/s² to +2.2m/s²

EV6 can pair match identification, validated with CommaCarSegments


adeebshihadeh:
Let's revisit this later

PR #3226, opened 2026-03-24
add stock ACC description for cars without openpilot long

haraschax:


opendbc/car/docs_definitions.py
```diff
@@ -358,6 +358,9 @@ def get_detail_sentence(self, CP):
       exp_link = "<a href='https://blog.comma.ai/090release/#experimental-mode' target='_blank' class='highlight'>Experimental mode</a>"
       if CP.openpilotLongitudinalControl and not CP.alphaLongitudinalAvailable:
         sentence_builder += f" Traffic light and stop sign handling is also available in {exp_link}."
+      else:
+        sentence_builder += " Gas and brakes are controlled by the car's ACC system and behaves as it did without openpilot."
```
sshane:
```suggestion
        sentence_builder += " Gas and brakes are controlled by the car's ACC system and will behave as they do without openpilot."
```

TheSecurityDev: Nitpick, but I would also recommend "acceleration and braking" instead of "gas and brakes" to account for EVs

haraschax: both addressed

PR #3249, opened 2026-03-28
Tesla: smooth transition after gas pedal override                                                                                                            

lukasloetkolben:
ramps jerkMin/jerkMax from 0 back to their normal values after gas release. This limits how fast the DI rate limiter can move away from 0 after the override ends, giving a smooth transition instead of an instant jump.

Validation
* Dongle ID: 0f79c454f812791a
* Route: 0f79c454f812791a/00000011--f1fa3e189e


lukasloetkolben:
Without the fix, the acceleration **undershoots to -2.2 m/s²** on gas release:
Bildschirmfoto 2026-03-28 um 14 10 14

With the fix, the undershoot is **reduced to -1.02 m/s²**  feels responsive and smooth:

Bildschirmfoto 2026-03-28 um 14 12 57

dzid26:
Consider this one https://github.com/dzid26/opendbc/commit/3c3b826bdf0088837e1396935529fdcd86851980
I think using the cruise OVERRIDE state is a bit more robust. -  I noticed that without limiting the jerk, whenever there is a high deceleration, the the OVERRIDE stayed on until the speed decreased even when after the gas was released. 

It also was needed to only rate limit maximum jerk funnily enough.

lukasloetkolben:
I tested this again today, the bug with the rapid pedal pushes @sshane and i noticed during the test drive, is already present in the current implementation and isn't caused by this fix. The override state also doesn't seem to be necessary.

In my view, this is a clear improvement. I've already submitted a similar solution a year ago.
I'd like to just merge this fix for now. If a better solution comes along later, we can always merge that one instead, but as it stands, this PR is a clear improvement.

sshane:
> I tested this again today, the bug with the rapid pedal pushes @sshane and i noticed during the test drive, is already present in the current implementation and isn't caused by this fix. The override state also doesn't seem to be necessary.

The overshoot we saw on our test drive was under these conditions:
- set speed 20 mph, traveling 20 mph
- tap gas light enough to cause almost no deviation from set speed
- after override, openpilot requests ~-0.1 m/s^2
- car brakes to 15 mph then accelerates back up to 20 mph

I tested this exact case on our M3 on master and couldn't reproduce it.

However, with https://github.com/commaai/opendbc/pull/3314 this is totally eliminated and overriding is smooth.

sshane:
In our HW 4 M3, I tried the following to no avail, still overshooted after override:

- match stock setSpeed (v_ego + accel)
- match inactive accel signals
- set jerkMin to 0 (still braked!)
- set both accel signals to 0 (still braked! sometimes...)
- set setSpeed to 400, both accels 0/positive 2, jerkMin 0

You can even request max acceleration all the time, and it won't ever accelerate above the set speed on the dash. I think the reason FSD doesn't jerk is because there's no DI set speed anymore. Even stock TACC had a very similar amount of jerk as openpilot does, it seems expected and desired. Noting that this jerk ramp is a hack.

The only thing that worked was setting both jerkMin AND jerkMax to 0. Not sure we fully understand these signals, but this PR is a good experience now

---

@lukasloetkolben I hope it's clear now why I didn't want to merge the PR in this state before, no statements were made that described how in depth you went to find that ramping up jerk worked, and why I thought maybe [matching stock signal behavior](https://github.com/commaai/opendbc/pull/3290) more closely might have worked.

opendbc/car/tesla/values.py
```diff
@@ -131,6 +131,7 @@ class CarControllerParams:
   ACCEL_MIN = -3.48  # m/s^2
   JERK_LIMIT_MAX = 4.9  # m/s^3, ACC faults at 5.0
   JERK_LIMIT_MIN = -4.9  # m/s^3, ACC faults at 5.0
+  JERK_RAMP_RATE = JERK_LIMIT_MAX * 0.002  # m/s^3 per control step, for smooth gas override
```
sshane: Oh this is 20 seconds until we can achieve full jerk again! This is way too long, let me lower and test

lukasloetkolben:
> In our HW 4 M3, I tried the following to no avail, still overshooted after override:
> 
> * match stock setSpeed (v_ego + accel)
> * match inactive accel signals
> * set jerkMin to 0 (still braked!)
> * set both accel signals to 0 (still braked! sometimes...)
> * set setSpeed to 400, both accels 0/positive 2, jerkMin 0
> 
> You can even request max acceleration all the time, and it won't ever accelerate above the set speed on the dash. I think the reason FSD doesn't jerk is because there's no DI set speed anymore. Even stock TACC had a very similar amount of jerk as openpilot does, it seems expected and desired. Noting that this jerk ramp is a hack.
> 
> The only thing that worked was setting both jerkMin AND jerkMax to 0. Not sure we fully understand these signals, but this PR is a good experience now
> 
> @lukasloetkolben I hope it's clear now why I didn't want to merge the PR in this state before, no statements were made that described how in depth you went to find that ramping up jerk worked, and why I thought maybe [matching stock signal behavior](https://github.com/commaai/opendbc/pull/3290) more closely might have worked.

The logic is that the DI tries to reach the setSpeed. In doing so, we set the limits for the maximum and minimum allowed acceleration, as well as the limits for the maximum and minimum jerk.
Since we don’t calculate setSpeed but rather acceleration in openpilot, the idea was to set setSpeed to either 0 or V_CRUISE_MAX so that DI stays as close as possible to the set acceleration.

opendbc/car/tesla/teslacan.py
```diff
@@ -28,15 +29,23 @@ def create_steering_control(self, angle, enabled):
...
     return self.packer.make_can_msg("DAS_steeringControl", CANBUS.party, values)
 
-  def create_longitudinal_command(self, acc_state, accel, counter, v_ego, active):
+  def create_longitudinal_command(self, acc_state, accel, counter, v_ego, gas_pressed):
     set_speed = min(max(v_ego + accel, 0) * CV.MS_TO_KPH, 400)
 
+    # There exists a jerk after overriding with gas above the set speed.
+    # This jerk exists on stock TACC as well, so we ramp up jerk to avoid this.
+    # Setting both jerks to 0 seems to be the only thing that prevents the unintentional braking.
+    if gas_pressed:
```
sshane, line 39: look into making this active

haraschax:
After looking into this I agree this is a stock tesla issue. It only happens when going over the set speed, which can be very frequent in tesla since it sets the set speed automatically. 

Doesn't seem possible to correctly mitigate since it respects neither the jerk or accel limits when you are going faster than the set speed and brakes hard regardless. 

This is also trivial to mitigate to just set the tesla set speed to be speed_limit+30. Closing this as I see no good fix that can be merged.

PR #3263, opened 2026-03-29
GWM Haval H6: Initial Platform Support

AlexandreSato:
Video: https://youtu.be/5AI3S0qTVkQ

Routes:
```075b133b6181e058/00000163--443c7a62d5/25```
```075b133b6181e058/00000142--8e663b9956/0```
```075b133b6181e058/00000150--63f9240463```

Another Users:
Conrado's comma three + red panda:
 ```fa46c09404cdf098/00000036--a5485afd0f```
```fa46c09404cdf098/00000002--e9b78964ca/0```

discord otaviobps comma threex:
github https://github.com/otaviobonder
Dongle ID ```756ee253ae4159bc``` (not shared connect to me)
Test of fingerprint1: https://connect.comma.ai/756ee253ae4159bc/00000001--89e6d44254
Test of fingerprint2: https://connect.comma.ai/756ee253ae4159bc/00000002--59690af643

discord: facundomon comma four Haval Jolion (Will need a separate port because the RX that EPS sends back to the camera happens differently.)
github: https://github.com/fmonpelat
```43c6c755ff0b0948/00000003--9385d3942d```

Harness: 
https://discord.com/channels/469524606043160576/1257100388763566243/1361026751425151046
connectors:  1376352-1 and 1565804-1

Harness Schematic:
https://discord.com/channels/469524606043160576/1257100388763566243/1483033966075445269
V1: https://discord.com/channels/469524606043160576/1257100388763566243/1470540491765321758
V3:
image

Tools to disassemble camera cover:
https://discord.com/channels/469524606043160576/1257100388763566243/1470160666232099019

### CODE REVIEW

**Comment by adeebshihadeh** [comment]
Is this ready for an initial review? If so, can you get CI passing first?

**Comment by AlexandreSato** [comment]
> Is this ready for an initial review? If so, can you get CI passing first?

Yes!!

File: opendbc/safety/safety.h [hunk 1]
```diff
@@ -408,6 +409,7 @@ int set_safety_hooks(uint16_t mode, uint16_t param) {
     {SAFETY_RIVIAN, &rivian_hooks},
     {SAFETY_TESLA, &tesla_hooks},
     {SAFETY_HYUNDAI_CANFD, &hyundai_canfd_hooks},
+    {SAFETY_GWM, &gwm_hooks},
```
adeebshihadeh: this should go under `ALLOW_DEBUG`. once we're able to do final validation ourselves, we'll move it out

opendbc/car/gwm/interface.py
```diff
@@ -0,0 +1,89 @@
...
+    ret.brand = 'gwm'
+
+    cfgs = [get_safety_config(structs.CarParams.SafetyModel.gwm)]
+
+    # If multipanda mapping is detected (offset >= 4), keep the first safety slot
+    # as `noOutput` so an internal panda remains silent and the vehicle safety config
+    # stays as the last entry (`-1`). This enables external panda to control the vehicle.
+    CAN = CanBusBase(None, fingerprint)
+    if CAN.offset >= 4:
+      cfgs.insert(0, get_safety_config(structs.CarParams.SafetyModel.noOutput))
```
adeebshihadeh: let's remove this. it's no longer supported

opendbc/car/gwm/interface.py
```diff
@@ -0,0 +1,89 @@
...
+    self.pcm_follow_distance = cp_cam.vl["ACC"]["CAR_DISTANCE_SELECTION"]
+
+    ret = super().update(can_packets)
+    ret.steerFaultTemporary |= self.steer_fault_temporary_counter > 100
+    if (self.pcm_follow_distance == 4 and self.current_personality != 3) or \
+       (self.pcm_follow_distance == 3 and self.current_personality != 3) or \
+       (self.pcm_follow_distance == 2 and self.current_personality != 2) or \
+       (self.pcm_follow_distance == 1 and self.current_personality != 1):
+      self.press_gac_button = not self.press_gac_button
+    ret.buttonEvents = create_button_events(self.press_gac_button, True, {1: ButtonType.gapAdjustCruise})
```
adeebshihadeh, line 46: not sure what's going on here, but these should generally go in the car state or controller files

opendbc/car/gwm/fingerprints.py
```diff
@@ -0,0 +1,13 @@
...
+from opendbc.car.structs import CarParams
+from opendbc.car.gwm.values import CAR
+
+Ecu = CarParams.Ecu
+
+FW_VERSIONS = {
+  CAR.GWM_HAVAL_H6: {
+    (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x873612100XEC56000\xf1\x89S013A01XKN17002',
+    ],
```
adeebshihadeh, line 11: do we get any other firmware?

opendbc/safety/tests/test_gwm.py
```diff
@@ -0,0 +1,137 @@
...
+from opendbc.car.gwm.values import GwmSafetyFlags
+from opendbc.car.structs import CarParams
+import opendbc.safety.tests.common as common
+from opendbc.safety.tests.libsafety import libsafety_py
+from opendbc.safety.tests.common import CANPackerSafety
+from opendbc.car.gwm.gwmcan import checksum as _checksum
+
+
+opendbc = "gwm_haval_h6_mk3_generated"
+dbc = DBC(opendbc)
```
adeebshihadeh, line 14: what's this global for?

adeebshihadeh (commented):
Nice work, this is a pretty complete port! 

I did a quick pass, but there's still more to cleanup here. I recommend working with an LLM, like codex, to refine this a bit more.

opendbc/safety/safety.h
AlexandreSato: Done!

opendbc/car/gwm/interface.py
AlexandreSato: Done!

opendbc/car/gwm/interface.py
AlexandreSato: The reason this ended up in `interface.py` is that I need to compare state originating from both `CarState` and `CarController`.

For the EPS fault detection, I first tried moving the loopback CAN parser logic into `CarState` (that attempt is visible in the commit history), but it did not work reliably there. The issue is that `CC.latActive` only exists by design on the `CarController` side, while `A_RX_STEER_REQUESTED` is coming from CAN state. So the temporary EPS fault detection needs to evaluate:

* what OP *requested* (`CC.latActive`)
  vs
* what EPS is actually *obeying* (`A_RX_STEER_REQUESTED`)

That comparison inherently crosses the CarController/CarState boundary, which is why `interface.py` became the practical place to glue both together.

The follow distance synchronization has the same architectural issue. `hud_control.leadDistanceBars` also originates from `CarController`, while the actual vehicle-selected distance (`CAR_DISTANCE_SELECTION`) comes from CAN/CarState.

The motivation here is that the car itself already has a physical button to increase/decrease the ACC distance bars, while openpilot personality only “rotates the carousel” internally. That was creating a confusing mismatch where the personality selected in openpilot did not correspond to the number of bars currently shown in the vehicle cluster.

I also tried the opposite approach (spamming the vehicle buttons to continuously force the car into the desired state), but the GAC distance button cannot really be spammed reliably. So the more robust solution was:

* read the actual distance currently selected in the car
* compare it against OP personality
* inject a single button press only when they diverge

This implementation needed access to `hud_control.leadDistanceBars`, which again is only available from the controller side, so `interface.py` was the only shared location that had access to both worlds cleanly.
video: https://youtu.be/JEcRE7gwpzU

opendbc/car/gwm/fingerprints.py
AlexandreSato: Unfortunately, I haven't been able to. I've tried everything I could, and it only responds on the OBD port.

opendbc/safety/tests/test_gwm.py
AlexandreSato: I needed this because these hardcoded 
image
https://discord.com/channels/469524606043160576/1282121586002104402/1489357983858163904 values ​​cause an overflow in the test.
So I needed to create a get_signal_range() function to avoid overflowing the bits.
image

elmarculino:
Hi @AlexandreSato — I've been reviewing this port for my own use and ended up cleaning up several of the open items. Everything is on a branch off your PR head (`dde1d1ee`) so it's easy to cherry-pick what you want: **`elmarculino/opendbc:gwm-haval-h6-improvements`**. All GWM safety tests, the car-interface fuzz test, MISRA, and ruff pass.

Biggest wins were driven by pulling real CAN from your **public** route `075b133b6181e058/00000163` (the `files` endpoint returns the rlogs without auth), which let me verify things against the actual car instead of guessing:

**1. BRAKE2 and gas checksums are now validated.** I derived the missing per-block CRC xor constants from the log — `BRAKE2` (0x120) block A = `0xEE`, gas `CAR_OVERALL_SIGNALS2` (0x60) block B = `0x95` — 100% consistent across ~9k/12k frames and a second segment. The method reproduces the constants we already knew (`0x7F`, `0x61`) exactly. So the two remaining safety-critical RX messages no longer run with `ignore_checksum`. I kept their **counters** ignored: in the logs both show a systematic ~1-in-15 counter irregularity that would trip false faults.

**2. `RX_STEER_RELATED` (0x147) checksum + counter validation.** Block B (bytes 8–15, xor `0x61`) carries the EPS/driver torque the torque safety checks rely on — it was being trusted unchecked.

**3. EPS-fault detection moved back into `carstate.py`.** Re @adeebshihadeh's note about this living in `interface.py`: I got your original loopback approach working. The failure was `canValid` dropping before engagement — registering the loopback `STEER_CMD` with `float('nan')` frequency (the GM pattern) sets `ignore_alive` and fixes it. The `interface.py` overrides are gone.

**4. Cruise state from the PCM signal.** Replaced the synthetic stalk-lever state machine (in both carstate and `gwm.h`) with `CRUISE_STATE_2` in `ACC` (0x2AB): `>2` = engaged. Semantics are from @otaviobonder's `haval-new-pcm-signal` branch and I confirmed them on your H6 log (values cluster {1,2,3}, `3` present exactly when OP was steering).

**5. Gap-adjust button fix.** The old `create_button_events(press_gac_button, True, …)` emitted a release event **every frame** in steady state (openpilot cycles personality on release, so this spun personality continuously). Rewrote it as rate-limited press/release pulses. I also saw in your `haval_match-distance-lines` branch that you'd tried the stalk-button-edge approach first — kept the display-sync design since the stock ACC consumes those button presses.

Smaller: dropped the stale multipanda comment, simplified the accel normalization, and added a `opendbc/car/gwm/README.md` documenting the CRC scheme, bus layout, and these decisions.

Two things I **couldn't** verify without the car, worth a road test: that the panda loopback echoes actually arrive for the fault detection, and the `STEER_CMD` block-3 CRC (`0x34`) from your `haval_fwd-hook` branch — this PR's `create_steer_command` passes that block through stale, which may produce occasional invalid CRCs.

Happy to open a PR against your branch if any of this is useful. Thanks for the port!

PR #3267, opened 2026-03-31
VW MQB: fix temporary "front assist not available" on startup

RJWadley:
goal: cluster should not show "Front assist not available" when the car first starts.

this is the first of two PRs to replace #3228

From what I can tell, the actual faulting bit here is `KBI_Konsistenz_ACC` on `Kombi_01` which presumably indicates a counter or checksum issue on `ACC_02`. The fix is to seed the counter of `ACC_02` so that it doesn't jump around when openpilot starts. While we're at it though, we might as well seed the counters for `ACC_06`, `ACC_07` and `ACC_10`.

route before where the counter happened to line up: `9feef6a0323edf69/000001a8--0d03868708 `
route before where the counter jumped: `9feef6a0323edf69/000001a9--62b71a9e3d `
route after: `9feef6a0323edf69/000001b3--d33ea2e343`


JaCzekanski:
@RJWadley Looks like a correct approach to me.
Starting OP with long mode enabled causes warnings on the dash to appear and DTCs to be saved in module 13 (sometimes also in cluster, camera, and engine).
These DTCs might prevent Front Assist / ACC from working entirely until codes are cleared (happened to me before).

@jyoung8607 Can we land these changes?

PR #3276, opened 2026-04-02
HKG: Car Port for G80 2022 (3.3T, non-HDA2)

sunnyhaibin:
Validation
* Dongle ID: `a10338a12c34322a`
* Route: `00000016--ee7e39da11`


sunnyhaibin:
@sshane can we reopen this?

PR #3288, opened 2026-04-04
docs: add a bunch of yt vidoes (codex's picks)

adeebshihadeh:


gerrylum (commented):
The Rivian YouTube link in this PR looks like it may not be the best example. A more representative video would be:

https://www.youtube.com/watch?v=dflSSGQwYNc

Happy to open a follow-up docs PR if that’s easier.

PR #3289, opened 2026-04-04
body v1 updater

stefpi:
body can flash script. runs on initialization of comma body car interface and flashes the body if the firmware signature does not match current version signature

- STM32F4 config, can handle, flash routine
- version and address values in values.py
- run update in init() of body interface.py
- add second UDS request to body FW query config for software fingerprint
- increase fw query timing from 0.1 to 0.2 to match 2 requests instead of 1


opendbc/car/body/fingerprints.py
```diff
@@ -9,7 +9,7 @@
 
 FINGERPRINTS = {
   CAR.COMMA_BODY: [{
-    513: 8, 516: 8, 514: 3, 515: 4
+    513: 8, 516: 8, 514: 3, 515: 4, 517: 8, 518: 6, 519: 6
```
adeebshihadeh, line 12: let's split this out into its own PR

opendbc/car/body/values.py
```diff
@@ -34,6 +44,11 @@ class CAR(Platforms):
       [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.UDS_VERSION_RESPONSE],
       bus=0,
     ),
+    Request(
+      [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.APPLICATION_SOFTWARE_FINGERPRINT_REQUEST],
+      [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.APPLICATION_SOFTWARE_FINGERPRINT_RESPONSE],
+      bus=0,
+    ),
```
adeebshihadeh: remove this and just pass the hash with the existing request

opendbc/car/body/values.py
```diff
@@ -7,6 +8,15 @@
 
 SPEED_FROM_RPM = 0.008587
 
+FIRMWARE_VERSION = "v0.3.1"
+BIN_URL = f"https://github.com/commaai/body/releases/download/{FIRMWARE_VERSION}/body.bin.signed"
+BIN_NAME = f"body-v1-{FIRMWARE_VERSION}.bin.signed"
+BODY_DIR = os.path.dirname(os.path.realpath(__file__))
+_PANDA_BIN_DIR = os.path.join(BODY_DIR, "../../../../panda/board/body/v1")
```
adeebshihadeh, line 15: let's just move the v1 stuff into panda

opendbc/car/body/flash.py
```diff
@@ -0,0 +1,148 @@
...
+
+REQUEST_IN = 0xC0
+REQUEST_OUT = 0x40
+DEFAULT_ISOTP_TIMEOUT = 2
+
+F4Config = {
+  "sector_sizes": [0x4000 for _ in range(4)] + [0x10000] + [0x20000 for _ in range(11)],
+}
+
+class CanHandle:
```
adeebshihadeh, line 17: this should live in panda

opendbc/car/body/flash.py
```diff
@@ -0,0 +1,148 @@
...
+  os.makedirs(os.path.dirname(bin_path), exist_ok=True)
+  result = subprocess.run(
+    ["curl", "-L", "-o", bin_path, "-w", "%{http_code}", "--silent", update_url],
+    capture_output=True, text=True
+  )
+  status = result.stdout.strip()
+  if status == "200":
+    print("downloaded latest body firmware binary")
+  else:
+    raise RuntimeError(f"download failed with HTTP {status}")
```
adeebshihadeh, line 71: openpilot will pass this in from the panda build obj dir

PR #3292, opened 2026-04-04
I fingerprinted my 2018 Leaf G (IC) Its a built in Japan model for the Japan Domestic market, later exported to New Zealand as a used car.

NZJDMLeafG2018:
Validation
* Dongle ID:
10ac256861c8fee0
* Route:
0ac256861c8fee0%7C00000002--3b9765496f


sshane:
Can you engage with openpilot?

PR #3294, opened 2026-04-05
safety: dynamically relax lateral jerk limits during accel conflicts

devtekve:
When a vehicle accelerates in a sharp curve, the steering correction required to stay within safe lateral acceleration limits can sometimes violate the maximum lateral jerk limit, triggering an unnecessary TX block.

The approach is to allow jerk limits to be exceeded as long as they comply with lateral acceleration, and only on the unwind direction. 

In practice, if you are on a tight turn and you don't decelerate, instead of faulting, we allow the steering wheel to keep unwinding to comply with the lat acceleration limit, even if it means that the jerk would exceed our calculated limits. Lateral acceleration being prioritized over jerk considering that excessive lateral acceleration could cause loss of grip. 

The alternative to this (which is what happens today) is that due to the safety violation controls would be dropped and an even higher jerk would be experienced as the physics of the car straighten the wheels on it's own. 

This PR resolves the conflict by prioritizing physical acceleration limits over _~~comfort-based~~_ ISO 11270 jerk limits:
* Reorders calculations to determine the absolute `max_angle_can` boundaries before applying jerk limits.
* Implements `SAFETY_MAX` and `SAFETY_MIN` to dynamically relax the `highest/lowest_desired_angle` boundaries.
* This allows the steering wheel to **unwind** safely to the acceleration edge in an emergency, while remaining mathematically branchless and strictly preventing commanded oversteer.

***
* More details on Discord: https://discord.com/channels/469524606043160576/1409799762190008331/1490274285246480424


sshane:
> comfort-based jerk limits:

it's the ISO 11270 jerk limits, not comfort

how does the solution look if we instead wind down considering jerk limits in opendbc, not safety? more or less complex?

dzid26:
@devtekve 
Relaxing command jerk limits could be bad if this whole thing was to be happening due to corrupted speed signal. 

Safety wise, ideally, I think jerk limit should take a priority, the steering should _not_ disengage and the driver should be required to take over immediately, and disengaged only after enough  driver torque is detected. But there is no framework for such reaction currently.  

This is a broader issue with panda safety that it optimistically assumes disengagement is always a safe state...

_____
What may mitigate the issue in practise is using better jerk calculation from:
https://github.com/commaai/openpilot/pull/24755#issuecomment-1146861863


$$
jerk = v^2\dot{\kappa} + 2va\kappa
$$

Positive (forward) acceleration will partially cancel out negative curvature (steering) rate and therefore make it harder to exceed the negative jerk limit.

devtekve:
@dzid26 that makes sense about the corrupted speed signal. I agree that is a risk, but because this only happens when the wheel is unwinding, I think the danger is different.

Even if a bad signal makes the wheel center too fast, it still feels safer than what happens today. Right now, we just drop all torque and let the mechanical trail of the car snap the wheels back to the center with zero control. In my view, a controlled unwind (even if it is fast) is always better than just letting go and letting the car physics take over.

Regarding the formula... I don't feel qualified enough to properly assess whether that better jerk calculation would prevent the issue though, maybe @Discountchubbs can chime in here if he has time :D 

*** 
@sshane 
> how does the solution look if we instead wind down considering jerk limits in opendbc, not safety? more or less complex?

Do you mean putting stricter limits in opendbc (_lower max lat accel and lower max jerk_) for a given car than ISO and leaving safety compliant with ISO?

devtekve:
Got a report from a user that they experienced safety tx blocks with this fix

https://community.sunnypilot.ai/t/support-for-newer-vehicles-equipped-with-angle-steering-lfa2/22/407?u=devtekve

And I can confirm the issue is still present, albeit at a smaller frequency, but it is still a problem, so I am converting this back to draft while I investigate.

image

### Route info
routeId: `23d8f79aed3fed05/00000220--000d201443`

The user was actively with his hands on the steering wheel and `steeringPressed` was true,  so the user was applying some torque on the steering wheel right before it happened

image

image

sshane:
You need to write a fuzzer to be able to validate this well

devtekve:
I am coming to the painful realization that there may not be a **_solution_** to this problem. 

#### TL; DR
Instead of relaxing the control limits, we need to implement a safe, gradual transition to the `stand-by` state before violating any safety constraints. By tracking our proximity to these limits and the inevitable conflict point where no compliant solution exists, we can warn the driver earlier to take over. This allows us to gracefully drop controls before Panda forcefully intervenes.

***
#### Background context on my existential crisis 

Fundamentally I am trying to fix a problem that has its root cause in a variable we can't assume we control from the controls perspective which is `vEgo`. 

an increase of `vEgo` is ultimately what causes the condition to be possible 
image

Reading the ISO 11270[^1], the limits are clear, 
* lateral acceleration shall not exceed 3m/s² and 
* the jerk shall not exceed 5m/s³... 
 
Basically we shall never exceed either. This PR is asking to allow exceeding one of those due to the lack of control of the `vEgo`. 

The ISO even acknowledges implicitly that some situations may require a deceleration
> The lane keeping action shall not cause a longitudinal deceleration larger than 3 m/s2. If the lane
keeping action causes a longitudinal deceleration larger than 1,0 m/s2, this shall not cause a speed
reduction more than 5 m/s.

Considering we can't assume we have control of the `vEgo` for the purposes of lateral safety, the only next logical "best thing" is literally what's hinted on the next point of the ISO:
> In case of a transition from LKAS active state to LKAS stand-by state, the lane keeping action shall
not end suddenly but shall be faded out smoothly.

Ultimately, since we cannot guarantee compliance under these specific conditions, pivoting to a graceful degradation strategy seems to be the only viable path forward. I would love to hear others' thoughts on this approach.

[^1]: [ISO 11270:2014 PDF Sample](https://cdn.standards.iteh.ai/samples/50347/558ca926c9a44182aa8af3ff1c9fb0ac/ISO-11270-2014.pdf)

devtekve:
I've recorded a stock route where I drove a little _questionable_ when LFA2 was engaged, trying to force the conflict situation and see how it resolves it / prioritizes. I am not sure if I reached the limits yet. This route is private but comma can see it: `4dcb6256482436c6/00000203--c53275626b` it's preserved and logs are uploaded.

On the bookmark it's where I was surprised it hadn't disengaged yet :D

PR #3315, opened 2026-04-09
Ford: Use Path Angle and Offset for CAN FD Lateral Control

hiimisaac:
**YOU ACCEPT ALL RISKS IF YOU TRY TO USE THE UNTESTED BRANCH** 

Routes are preserved but not public. Let me know if they need to be flipped public.

Validation
* Dongle ID: 112e4d6e0cad05e1
* Route (master): 112e4d6e0cad05e1/00000011--37860199e2
* Route (proposed): 112e4d6e0cad05e1/00000015--1e2f55ba5f

Background: How Ford's PSCM works
Ford's Power Steering Control Module (PSCM) accepts a path polynomial over CAN FD:
`y(x) = c0 + c1*x + c2*x²/2 + c3*x³/6`

Where:
c0 (path_offset): lateral offset in meters, range [-5.12, 5.11]
c1 (path_angle): heading angle in radians, range [-0.5, 0.5235]
c2 (curvature): inverse meters, range [-0.02, 0.02]
c3 (curvature_rate): set to 0

The PSCM is a black box. We don't know its internal controller or how it decides to actuate or anything. We can't read its internal state. All tuning is empirical.

**What master does today**
Master sends only c2 and zeros the other three coefficients:

```
can_sends.append(fordcan.create_lat_ctl2_msg(..., mode, 0., 0., -self.apply_curvature_last, 0., counter))
#                                                  ^    ^                                    ^
#                                              ramp c0=0 c1=0   
```

The curvature goes through apply_ford_curvature_limits (rate limits, lateral accel cap at ~2.4 m/s^2, error bounds against measured yaw rate). For the F-150 and Bronco Sport, there's an anti_overshoot function that applies a 5-second exponential filter with hysteresis to smooth curvature requests and prevent oscillation.

Mode is always 1 (active) or 0 (inactive). Ramp type is always 0. The PSCM only gets one signal out of the four available.

There's a TODO in master that acknowledges this:

> "in order to get further control on steer actuation, the other three signals are necessary"

This works for highway lane keeping. It does not work for:
90-degree turns (c2 maxes out at 0.02, roughly 2.4 m/s^2 of lateral accel at speed)
Moderate curves at city speeds where the PSCM needs more authority than curvature alone provides
Responsive low-speed steering (the anti_overshoot filter adds 5 seconds of smoothing lag)

**Why not just send bigger curvature?**
The DBC range caps c2 at 0.02 1/m. But more importantly, the PSCM accumulates internal state when curvature is sustained. We can't see this state and we can't flush it. On a long curve, the PSCM builds up an internal offset that persists after the curve ends, causing the car to drift or pull.

c1 (path angle) and c0 (path offset) are also stateful in the PSCM, but to a lesser degree. Through testing we found that keeping c1 below 95% and c0 below 90% of their DBC ranges prevents noticeable state accumulation on everything except the most aggressive 90-degree turns. And even when it accumulates, it flushes EXTREMELY quickly so holding the wheel for seconds after a turn is not a problem. (There have been attempted fixes in the past to combat what's seen in upstream: (https://github.com/commaai/openpilot/pull/31515))

**Proposal**
For CAN FD vehicles: stop sending curvature. Read the model's predicted trajectory directly and send it as c1 (heading) and c0 (lateral offset). The model outputs heading (orientation.z) and lateral position (position.y) at 33 points along the predicted path. We interpolate at a distance-based lookahead.

For non-CAN FD vehicles: no changes. The existing curvature pipeline (including anti_overshoot and apply_ford_curvature_limits) is preserved as-is. 

```
if self.CP.flags & FordFlags.CANFD and self.model is not None and len(self.model.orientation.z) > 0:
  d_look = max(v_ego * 1.0, 7.0)
  x_pts = np.array(self.model.position.x)
  path_angle = float(np.interp(d_look, x_pts, np.array(self.model.orientation.z)))

  d_c0 = float(np.interp(v_ego, [11., 14.], [d_look, 6.0]))
  path_offset = float(np.interp(d_c0, x_pts, np.array(self.model.position.y)))
else:
  # Non-CAN FD: curvature-only control (unchanged from upstream)
  ...
```
c2 = 0 always on CAN FD. All steering authority comes through c1 and c0.

**The Why**
7m distance floor on d_look. d_look = max(v_ego * 1.0, 7.0). At 10 mph the time-based lookahead only sees 4.5m ahead. The 7m floor looks further into the turn, giving the PSCM 2-3x more heading signal at low speed where it's naturally sluggish. This is what makes 90-degree turns possible.

Dual c0 lookahead. c1 and c0 serve different roles. c1 (heading) is feedforward: it tells the PSCM where to point the car. Looking far ahead is correct because you want to be at that heading when you arrive. c0 (lateral offset) is positional: the PSCM interprets it as "the path is offset by this much right now." If you feed it position.y at 22m ahead on a gentle curve, that might be 0.75m of offset. The PSCM reacts immediately, pulling the car into the curve before it arrives. This causes early entry and corner cutting at 35+ mph.

But you can't zero c0 at speed either. Without it, the PSCM has no lateral centering signal. c1 says "point this direction" but doesn't say "you're drifting off-center." On gentle highway curves, c1 alone is too slow to keep the car centered.

The fix: at speed, shorten c0's lookahead to 6m. At that distance, a gentle curve only produces about 0.05m of offset. Enough for micro-centering corrections, too small to corner cut. At low speed (below 25 mph), c0 uses the same d_look as c1 for full turn authority.

`d_c0 = float(np.interp(v_ego, [11., 14.], [d_look, 6.0]))`

**mode=2 (PathFollowingExtended) and ramp_type=3 (Immediate)**. More aggressive path following and faster internal state flush compared to master's mode=1/ramp=0. ramp_type=3 made a noticeable difference in testing. mode=2 may or may not have an effect; it doesn't hurt. In empirical testing, it seems to actuate a bit harder but I can't verifiably prove this.

**c1 clipped to 95%, c0 to 90% of DBC range**. Both signals cause PSCM state accumulation when sustained near max (like on 90-degree turns). The clips reduce this without losing meaningful authority. 95% of c1 is still 0.475 rad (27 degrees), which covers nearly every turn the car could be making autonomously. There are still some hardware/pscm limits, but this enables so much more control for open pilot in nearly all cases across the board.

**Rate limits fully uncapped from the opendoc side**. The PSCM has its own internal rate limiting that prevents more than ISO lat accel limits when being controlled autonomously. You can observe this in the plots below. I didn't see a way to measure latAccel int jotpluggler so I was using a custom series:

`v1 ** 2 * value` (v1 = /carState/vEgoRaw) and value being /controlState/curvature.

**Biggest Takeaway**
Curvature causes state buildup; heading and offset are better. This was the core discovery. Every previous approach tried to make curvature work: I tried gain schedules, EMA smoothing, overtrack correction, error blending with yaw rate. All of them fought the same fundamental problem. Moving to c1/c0 eliminated this class of issues, gave us FAR more actuation, far more accurate control and allowed previously thought impossible maneuvers in Ford vehicles.

Plots:
This is mostly just a straight 
(master)
image
(Proposed)
image

This is a very minor right turn. You can see at multiple points I had to hold the wheel to correct it. This illustrates the state that the c2 (curvature signal) in the polynomial has when driving.
(master)
image
(proposed. this completed the turn so didn't require me to hold the wheel to straighten any state).
image

This is a right turn the current controls are incapable of making. In efforts of trying to keep the plots as close to 1:1 as possible, I did not take this turn with open pilot on proposed either (despite me believing it's possible)
image

This is a very gentle S-curve the current controls are incapable of making:
(master)
image
(proposed)
image

**This may or may not work for Q3 Fords. I was unable to find someone to validate this.**


hiimisaac:
I'm pulling this back while I work through some tuning on steady-state curves

hiimisaac:
After a 45 mile trip with this as-is, I think I was testing old logic when I posted the previous comment because this centered well at almost all speeds. (I didn't test 75mph+)

PR #3316, opened 2026-04-09
Honda - Acura RDX 22-26 release support

mvl-boston:
Updating Acura RDX to release support

New validation route: 6402b5bcb5d0f678/00000001--ab688904db

- Increasing steering to factory 3840 - see dashcam route 23412c73dbeccd3a/00000018--84d1807b5d
- Bumping covered years to 2026 - no fingerprint changes - 3792d010590cb83a/00000002--81bb430204
- Excluding 2025 per adeeb, due to inability to validate completion of NHTSA EPS recall


mrajaypalsingh1-coder:
Hello,
​I am still very interested in this Pull Request and believe it remains relevant to the project. Please keep this PR open as I am tracking its progress. Thank you for your work on this."

mvl-boston:
The earlier routes expired.  Can you record a new test route and preserve it?

opendbc/car/honda/values.py
```diff
@@ -258,7 +258,11 @@ class CAR(Platforms):
     flags=HondaFlags.BOSCH_ALT_BRAKE,
   )
   ACURA_RDX_3G_MMR = HondaBoschPlatformConfig(
-    [HondaCarDocs("Acura RDX 2022-24", "All", min_steer_speed=70. * CV.KPH_TO_MS)],
+    [
+      HondaCarDocs("Acura RDX 2022-24", "All", min_steer_speed=70. * CV.KPH_TO_MS),
+      # excluding 2025 due to inability to validate NHTSA 25V582 recall completion
```
adeebshihadeh, line 263: only the 2025 is subject to this?

adeebshihadeh: also is there any way for us to tell if the car has the updated FW? can we read a bit over CAN? something in the FW version?

opendbc/car/honda/values.py
```diff
@@ -258,7 +258,11 @@ class CAR(Platforms):
     flags=HondaFlags.BOSCH_ALT_BRAKE,
   )
   ACURA_RDX_3G_MMR = HondaBoschPlatformConfig(
-    [HondaCarDocs("Acura RDX 2022-24", "All", min_steer_speed=70. * CV.KPH_TO_MS)],
+    [
+      HondaCarDocs("Acura RDX 2022-24", "All", min_steer_speed=70. * CV.KPH_TO_MS),
```
adeebshihadeh, line 262: the min steer speed is that high on a modern MY car?

opendbc/car/honda/values.py
mvl-boston: Only 2025 was subject to the recall.   A new 25 OBD reg was miscoded by a vendor and fixed via recall in Sept 2025, before MY 26 went on sale.

The FW ID is different post recall, but only visible with comma power.  The Honda gateway filters out the firmware communications from the LKAS camera wire.

opendbc/car/honda/values.py
mvl-boston: Yes, factory LKAS has this limit in the owners manual and is enforced by the car-side ECUs, not the camera.

Other than shutting down and replacing the ECU messages (which is done in alphalong) we have not found a way around this.  It is the identical limitation in the upstreamed 2021-2026 Honda Odyssey.

mvl-boston:
Updated with new validation route.

PR #3318, opened 2026-04-09
Hyundai: Car Port for Elantra 2024-25

sunnyhaibin:
**Requires**
- https://github.com/commaai/opendbc/pull/3320
- https://github.com/commaai/opendbc/pull/3319

Closes https://github.com/commaai/opendbc/issues/1119

PR #3319, opened 2026-04-09
Hyundai CAN: split DBC for different messages definitions

sunnyhaibin:
**Description**
- Generates new DBC `hyundai_can_refresh_generated.dbc` for some newer Model Year Hyundai CAN with different DBC definitions
- Some newer Hyundai CAN platforms have different signal positions for the same CAN IDs (e.g. `1157 LFAHDA_MFC` is now 8-byte in https://github.com/commaai/opendbc/pull/3318)
- `hyundai_can_refresh_generated.dbc` will not be used in this PR yet
- No functional change in this PR

**Requires**
- https://github.com/commaai/opendbc/pull/3320

**Prerequisites of**
- https://github.com/commaai/opendbc/pull/3318


sunnyhaibin:
@sshane can we reopen this?

PR #3330, opened 2026-04-14
Honda - 2026 Pilot - mid-model refresh

mvl-boston:
Support for 2026 Honda Pilot.

Same controls as 2023-2025, but the mid-model refresh added substantially stronger EPS, so car needed separate platform and torque values.

Max 3628 factory steer: 023b97f3d8c08c33/00000012--89ee1ba221


adeebshihadeh:
Thanks!
image

PR #3333, opened 2026-04-15
Fix HONDA_CITY_7G mass in honda/values.py

baninfelipe:
Changed the mass for HONDA_CITY_7G from a calculated value (3125 * CV.LB_TO_KG) to a fixed value of 1184 kg for improved accuracy and consistency.

Validation
* Dongle ID:
* Route:


adeebshihadeh:
This doesn't do anything

PR #3335, opened 2026-04-18
Honda - additional comma power fingerprints for Acura RDX 3G MMR

mvl-boston:
2024: 4d8ba79da0eabd91/00000031--ecbad56760
2026: 3792d010590cb83a/00000009--8b3387841b


sshane:
It should be able to FP without comma power though, right?

mvl-boston:
Yes all these fingerprint wo comma power already.  The camera side of the gateway only exposes fwdcamera and fwdradar ecus and those IDs are the same for every year of the RDX_3G_MMR.

Would it be preferred to remove the other Ecu types instead so fingerprinting works without this inventory?

PR #3336, opened 2026-04-18
car: common function for delayed factory cruise canceling

### LINKED ISSUE #1173: [$300 Bounty] Toyota: cancel cruise without chime
The way openpilot cancels cruise control on most Toyotas has the side effect of a chime from the car and a momentary fault on the dash. When openpilot sends the cancellation message, it also sends a message to play a good chime to mask the fault chime that would play, so start by [removing that](https://github.com/commaai/openpilot/blob/fa62b9d3f9882ec33559ceb7377fa991be1d4fb5/selfdrive/car/toyota/toyotacan.py#L82). Once you remove the good chime, you should hear the fault chime, which we're trying to get rid of.

The Lexus IS uses a [different cancellation message](https://github.com/commaai/openpilot/blob/fa62b9d3f9882ec33559ceb7377fa991be1d4fb5/selfdrive/car/toyota/carcontroller.py#L94) than the rest of the Toyotas, so that's probably a good place to start.

$200 for a merged PR that meets this criteria:
1. no chime sound or even momentary fault display
2. reliable, i.e. always immediately cancels cruise
3. generic across all Toyotas

Feel free to open a PR once steps 1 and 2 are done to lock the bounty pending confirmation of 3.

sunnyhaibin:
May resolve https://github.com/commaai/opendbc/issues/1173, testing pending.

- [x] consolidate down to 1-liner within carcontroller


sunnyhaibin:
@sshane got a chance to test this on the Toyota yet?

opendbc/car/toyota/carcontroller.py
sshane, line 1: A refactor shouldn't change behavior, this is why it hasn't been merged

sshane:
If you split out Toyota we can merge

PR #2429, opened 2025-07-02
Non-Chinese BYDs Brand port

dotslashofficial:
This pull request is initiated for BYD models that has been exported to countries like Thailand,  Singapore, Malaysia, Australia etc. Primarily only for Atto 3 only. Will eventually be expandable to Seal, Sealion, Dolphin etc

Validation
* Dongle ID: 6c120d820b5e46ab
* Route: 6c120d820b5e46ab/2025-02-21--11-00-10/0 (not the actual BYD recorded route YET)

This PR is to start the bounty lock, continuation of the cleanup will be further discussed below

PR #3337, opened 2026-04-19
BYD: Atto 3 / Initial Platform Support

lukasloetkolben:
I've been digging into the steering message and the unknown bits flagged in PR #2429 are most likely the rate limits for the angle control.

If I can get the bounty locked, my plan would be:
1. Work remotely on an Atto 3 I have access to testing everything we have so far and iterating until it's in a solid state
2. Once things look good, rent a car locally to do final validation and get BYD fully supported


adeebshihadeh:
Locked!

shemps:
With the help of the information here I got both lateral and longitudinal control working on a BYD Atto 3 — though on carrotpilot on a mini-PC with a webcam rather than a comma device, since comma doesn't ship to where I'm based.

Not claiming the bounty and not opening a competing PR — I am publishing what I found in case it saves someone else the time:

https://github.com/shemps/byd-atto3-openpilot-port

The bits most likely to be useful here:

- FW query has to use 0xF195 — BYD ECUs NRC on 0xF188, and the MPC camera (0x704) doesn't answer the bus-0 sweep at all, so it needs to be non-essential.
- 0x318 is constant 0xFF on this car; the real EPS state is on 0x1FC.
- gasPressed has to come from PEDAL.GAS_PEDAL, not DRIVE_STATE.RAW_THROTTLE — the latter isn't the pedal and silently blocks longitudinal engagement.
- Independent confirmation of this PR's read of the 0x1E2 rate-limit bits, and the 0x32E ACCEL_CMD scale I measured (raw × 0.05 − 5 m/s²).
- The Veoneer private bus is CAN-FD and the rich object tracks are private-bus only — bears on radarUnavailable = True here.

adeebshihadeh:
Merged, thanks @lukasloetkolben!

@shemps want to make a follow-up PR with the stuff you're actually driving on?

PR #3339, opened 2026-04-19
CI: Increase car_diff timeout to 20 minutes

mvl-boston:
Car_diff needs longer to run, adding time.

See: https://github.com/mvl-boston/opendbc/actions/runs/24632710264/job/72022753334 for a job that needed more than 10 minutes.

### DIFF
--- a/.github/workflows/car_diff.yml
+++ b/.github/workflows/car_diff.yml
@@ -8,7 +8,7 @@ jobs:
   comment:
     name: comment
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sshane:
@elkoled is it expected to be this slow?

elkoled:
It was a race condition in the comment posting workflow, car diff execution time is usually 1-7 min depending on car brand on forks and 2-4x faster on fast action runners.
Closing as fixed in #3344

PR #3341, opened 2026-04-21
tesla: match stock ACC envelope convention for DAS_control

haraschax:


opendbc/safety/tests/test_tesla.py
```diff
@@ -429,6 +429,32 @@ def test_stock_aeb_passthrough(self):
     self.assertEqual(0, self.safety.safety_fwd_hook(2, aeb_msg_cam.addr))
     self.assertFalse(self._tx(no_aeb_msg))
 
+  def test_encoder_panda_compatible(self):
```
sshane: We should write a fuzzing test instead of these tests for every brand

PR #3353, opened 2026-04-23
Tesla: fingerprint 2022 M3 long range

lukasloetkolben:
Validation
* Dongle ID: 50aebc7022420304
* Route: 50aebc7022420304/00000002--09e43b3451


sshane:
Is there a stock Autosteer route?

PR #3355, opened 2026-04-24
mazda: add alpha longitudinal support

yummydirtx:
Adds alpha openpilot longitudinal control for the Mazda CX-5 2022.

Ports the current Zoompilot Mazda-long behavior onto upstream APIs, including the 0.36 s actuator delay, -1.024 m/s² stop acceleration, speed-based acceleration ceiling/build rates, and the current hold, release, unlatch, repulse, and breakaway tuning.

The controller takes radar ownership only after the forward camera has settled and the car is stopped. It sends CANPacker-built cruise messages and captured empty radar tracks on both required buses, and restores the stock diagnostic session on an interrupted takeover.

Safety requires a physical SET/RES press, limits acceleration to -3.5–2.0 m/s², and byte-allowlists diagnostic and replacement-radar traffic.

Updated from [@zephleggett](https://github.com/zephleggett)’s current Zoompilot implementation.


opendbc/car/mazda/interface.py
```diff
@@ -14,7 +17,14 @@ class CarInterface(CarInterfaceBase):
   @staticmethod
   def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_long, is_release, docs) -> structs.CarParams:
     ret.brand = "mazda"
-    ret.safetyConfigs = [get_safety_config(structs.CarParams.SafetyModel.mazda)]
+    ret.alphaLongitudinalAvailable = candidate == CAR.MAZDA_CX5_2022
```
adeebshihadeh, line 20: have you asked around on discord for other mazdas to test?

opendbc/car/mazda/interface.py
```diff
@@ -14,7 +17,14 @@ class CarInterface(CarInterfaceBase):
...
+    ret.openpilotLongitudinalControl = alpha_long and ret.alphaLongitudinalAvailable
+    # Mazda-long still engages on the stock ACC-active transition even though
+    # we suppress the radar-owned CRZ_CTRL path and synthesize replacement
+    # longitudinal messages.
```
adeebshihadeh, line 24: i think this is self-explanatory without the comment

opendbc/car/mazda/interface.py
```diff
@@ -27,6 +37,30 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
     if candidate not in (CAR.MAZDA_CX5_2022,):
       ret.minSteerSpeed = LKAS_LIMITS.DISABLE_SPEED * CV.KPH_TO_MS
 
+    if ret.openpilotLongitudinalControl:
```
adeebshihadeh, line 40: you can write these unconditionally

opendbc/car/mazda/longitudinal.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def _update_crz_info_checksum(raw: bytes, bias: int = 0) -> bytes:
+  dat = bytearray(raw)
+  dat[7] = (_compute_inverted_sum_checksum(dat) + bias) & 0xFF
+  return bytes(dat)
+
+
+def clip(value: float, lower: float, upper: float) -> float:
+  return min(max(value, lower), upper)
```
adeebshihadeh, line 86: this exists elsewhere in the codebase

opendbc/car/mazda/longitudinal.py
```diff
@@ -0,0 +1,242 @@
...
+
+MAZDA_LONG_DBC = DBC("mazda_2017")
+
+RADAR_ADDR = 0x764
+RADAR_BUS = 0
+
+CRZ_INFO_ADDR = 0x21B
+CRZ_CTRL_ADDR = 0x21C
+
+CRZ_INFO_TEMPLATE = bytes.fromhex("01ffe20006800000")
```
adeebshihadeh, line 21: what is this?

adeebshihadeh (changes requested):
Thanks for the PR!

The overall concepts look good, but the implementation needs some cleanup, particularly `longitudinal.py` and `carcontroller.py`. This is a lot more code than all the other longitudinal implementations.

You should be able to use this tool to tighten up this implementation. https://github.com/commaai/openpilot/tree/master/tools/longitudinal_maneuvers

Also please cleanup the PR description... it's almost as long as the diff. I assume an LLM wrote that?

yummydirtx:
> Also please cleanup the PR description... it's almost as long as the diff. I assume an LLM wrote that?

I did actually write the PR description, I tried to make it match the comprehensive nature of the sample one in CONTRIBUTING for a car port to explain the totality of my approach and changes. AI was only used to clean up wording or find specific sources. I'll shorten it for you to make it significantly more concise and to the point.

Thanks for reviewing all this, I'll work to tighten things up and address the issues you found.

opendbc/car/mazda/interface.py
yummydirtx: Yes, it appears the shape of the MRCC signals are largely the same on other Mazda models, but the range of inputs seems different. For example, a user with a CX-9 ran my branch with the CX-5 manually selected reported symptoms similar to those shown in my CX-5 when out-of-bounds inputs are sent to MRCC. More testing is needed (and ideally a developer with a CX-9)

yummydirtx: Kept this limited to the CX-5 2022. The CX-9 report above is why this revision does not enable other Mazdas.

opendbc/car/mazda/interface.py
yummydirtx: Removed.

opendbc/car/mazda/interface.py
yummydirtx: Removed the old custom PID tune; the remaining stop acceleration and actuator delay are now assigned unconditionally.

opendbc/car/mazda/longitudinal.py
yummydirtx: Removed the local helper. The controller now uses numpy.clip and the shared rate_limit helper.

opendbc/car/mazda/longitudinal.py
yummydirtx: Removed the raw template and manual bit patching. CRZ_INFO and CRZ_CTRL are now built with CANPacker from named DBC signals, with their wire bytes covered by unit tests.

yummydirtx:
Reworked this on current master around Zoompilot’s current Mazda-long implementation. The old 242-line raw-template module is now a compact standstill-hold helper with the current Zoompilot release/unlatch, repulse, acceleration-ceiling/build, lift, and breakaway behavior; CRZ_INFO/CRZ_CTRL use CANPacker; radar takeover/restoration is isolated; and alpha long remains CX-5 2022-only.

The Zoompilot calibrations are preserved: 0.36 s actuator delay, -1.024 m/s² stop acceleration, -3.5–2.0 m/s² command range, 1.25/0.8 m/s³ build rates, -2.0 m/s³ lift limit, and the current hold/release/breakaway constants. This revision is 690 added lines, down from the prior 964-line pass and below the original PR’s 694 additions. The full local suite passes (Ruff, ty, codespell, cpplint, MISRA, and 4,007 runnable tests). @adeebshihadeh this is ready for another look.

opendbc/safety/modes/mazda.h
```diff
@@ -57,34 +115,108 @@ static bool mazda_tx_hook(const CANPacket_t *msg) {
...
+    {MAZDA_LKAS, 0, 8, .check_relay = true},
+    {MAZDA_CRZ_BTNS, 0, 8, .check_relay = false},
+    {MAZDA_LKAS_HUD, 0, 8, .check_relay = true},
+    {MAZDA_CRZ_INFO, 0, 8, .check_relay = false},
+    {MAZDA_CRZ_CTRL, 0, 8, .check_relay = false},
+    {MAZDA_RADAR_STATIC, 0, 8, .check_relay = false},
+    {MAZDA_RADAR_TRACK_1, 0, 8, .check_relay = false},
+    {MAZDA_RADAR_TRACK_1 + 1U, 0, 8, .check_relay = false},
+    {MAZDA_RADAR_TRACK_1 + 2U, 0, 8, .check_relay = false},
+    {MAZDA_RADAR_TRACK_1 + 3U, 0, 8, .check_relay = false},
```
adeebshihadeh, line 206: why are we sending radar track messages?

opendbc/car/mazda/interface.py
```diff
@@ -15,6 +15,13 @@ class CarInterface(CarInterfaceBase):
   def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_long, is_release, docs) -> structs.CarParams:
     ret.brand = "mazda"
     ret.safetyConfigs = [get_safety_config(structs.CarParams.SafetyModel.mazda)]
+    ret.alphaLongitudinalAvailable = candidate == CAR.MAZDA_CX5_2022
```
adeebshihadeh, line 18: doesn't work on others or hasn't been tested?

opendbc/car/mazda/mazdacan.py
```diff
@@ -1,6 +1,69 @@
+from opendbc.car.can_definitions import CanData
 from opendbc.car.mazda.values import Buttons, MazdaFlags
 
 
+RADAR_STATIC_MSG = (0x499, bytes.fromhex("0008c00000000000"))
+RADAR_TRACK_MSGS = {
```
adeebshihadeh, line 6: why do we have to send these?

adeebshihadeh (commented):
This is quite a bit different and more complex than the rest of the longitudinal control implementations in openpilot. Is there a compelling reason for this?

PR #3360, opened 2026-04-26
[codex] move CAN ignition hook

adeebshihadeh:


elkoled:
done:
https://github.com/commaai/opendbc/pull/3370

PR #3368, opened 2026-05-01
Add new fingerprints for Toyota ECUs RAV4 Prime (UK)

dadamsuk:
This PR adds fingerprints for a UK RAV4 PHEV (called the RAV4 Prime in North America).  Previously I used the fixed vehicle build, which worked, so I then created this fork and extracted the fingerprints as directed. It then installed first time and worked correctly.

Validation
* Dongle ID: 755e385db29aabb2
* Route: 755e385db29aabb2/00000005--bb606a5a97


dadamsuk:
One of the tests failed. I'm not familiar with this codebase at all, but it looks like one of the test's segments is not available to download. From all I can tell, it's nothing to do with me or the changes I've made.

PR #3369, opened 2026-05-01
Tesla - add HW3 detection to avoid false detection of FSD14

erikfleischer:
In FW 2026.8.6 Tesla has changed the behavior of stock LKAS. During an intervention AP sends DAS_steeringControlType == 1 which triggers the FSD14 detection heuristic. This causes an invalid LKAS settings fault in Openpilot despite Autosteer being deactivated in the vehicle settings.

Segment which demonstrates the fault in nightly-dev: c0fe59cc1ebbd1b3/000000a0--cdc57134fe/16

Validation
* Dongle ID: c0fe59cc1ebbd1b3
* Route: c0fe59cc1ebbd1b3/000000a4--7aa1bd45e0
* Stock LKAS intervention no longer triggers an openpilot fault: 08:23 into the route
* Vehicle follows openpilot steering commands: 3:45 into the route.

### DIFF
--- a/opendbc/car/tesla/values.py
+++ b/opendbc/car/tesla/values.py
@@ -75,6 +75,7 @@ FW_QUERY_CONFIG = FwQueryConfig(
 )
 
 # Cars with this EPS FW have FSD 14 and use TeslaFlags.FSD_14
+# For these cars we need to send a different value on DAS_steeringControlType.
 FSD_14_FW = {
   CAR.TESLA_MODEL_3: [
     b'TeMYG4_Main_0.0.0 (77),E4HP015.04.5',
@@ -89,6 +90,15 @@ FSD_14_FW = {
   ]
 }
 
+# HW3 cars with FW 2026.8.6+ can trigger a false FSD14 detection.
+HW3_FW = {
+  CAR.TESLA_MODEL_3: [
+    b'TeM3_E014p10_0.0.0 (24),E014.20.2',
+  ],
+  CAR.TESLA_MODEL_Y: [
+    b'TeM3_E014p10_0.0.0 (24),YP002.21.2',
+  ]
+}
 
 class CANBUS:
   party = 0
@@ -143,6 +153,7 @@ class TeslaFlags(IntFlag):
   LONG_CONTROL = 1
   FSD_14 = 2
   MISSING_DAS_SETTINGS = 4
+  HW3 = 8
 
 
 DBC = CAR.create_dbc_map()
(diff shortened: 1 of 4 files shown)


opendbc/car/tesla/values.py
```diff
@@ -89,6 +90,15 @@ class CAR(Platforms):
...
 
+# HW3 cars with FW 2026.8.6+ can trigger a false FSD14 detection.
+HW3_FW = {
+  CAR.TESLA_MODEL_3: [
+    b'TeM3_E014p10_0.0.0 (24),E014.20.2',
+  ],
+  CAR.TESLA_MODEL_Y: [
+    b'TeM3_E014p10_0.0.0 (24),YP002.21.2',
+  ]
+}
```
sshane: We actively want to remove FW lists, we shouldn't add any more right now. I've started on detecting FSD14/HW4 in the tesla test file, maybe can do that for HW3

Also don't we have to change the DBC definition to properly parse the 3 bit message now?

erikfleischer: Agreed. I do not like these FW lists as well but I am looking for a solution to avoid the described faults with HW3 vehicles.
I will take a look at test_tesla.py. Maybe that leads to better solution.

Regarding the 3rd bit in steering control message: Tesla actually moved the autosteer activation to the the 3rd bit. Thus openpilot actually uses the LKAS intervention bit to activate steering control in HW3 cars with FW 2026.8.6. I am a little torn on this. It is nice as the new behavior allows for some kind of torque blending. I have patched the DBC and the panda safety (https://github.com/erikfleischer/opendbc/tree/hw3_fw_2026_8_6_support) to make openpilot to use the same bit as autosteer. This makes the steering feel much nicer. But that branch breaks openpilot on HW3 with FW 2026.8.3 or older so is a no go for merging right now.

Once the false LKAS settings problems is fixed, I would be happy to take another look at the changed steering control message.

erikfleischer:
Changes:
 * Reworked FW detection to use regular expressions based on existing patterns in test_tesla.py
 * added tests to test_tesla.py to verify patterns against known FW fingerprints

New valiadion route: c0fe59cc1ebbd1b3/000000b1--427689fd57

At 0:46 in the route there is a LKAS intervention which does not trigger an openpilot fault.

At 0:56 in the route openpilot is engaged and steers the vehicle.

erikfleischer:
integrated changes for 3bit DAS_steeringControlType into a branch based on this: https://github.com/erikfleischer/opendbc/tree/tesla_update_can

erikfleischer:
this PR is made obselete by https://github.com/commaai/opendbc/pull/3389 → closing this

PR #3386, opened 2026-05-11
hyundai: reclassify availability-toggle button as main

arbazkhan971:
Reclassify the ambiguous Hyundai cruise button as mainCruise when the cruise availability state flips, so higher-level logic doesn't treat the availability toggle as a resume request.\n\nIncludes a focused regression test for the reclassification helper.


adeebshihadeh:

PR #3323, opened 2026-04-11
Tesla: Fingerprint 2019 Model 3 AWD Longrange (2026.8.6)

keenin:
Added Tesla: 2019 Model 3 AWD Longrange  2026.8.6 fingerprint

Validation
* Dongle ID: 424b2a6c1b69a3bc
* Route: https://connect.comma.ai/424b2a6c1b69a3bc/00000005--200c5f1bec/140/1190

PR #3389, opened 2026-05-14
Tesla: Fix DAS_steeringControlType on newer firmware

AmyJeanes:
On newer Tesla vehicle firmware across the fleet, we are seeing the `DAS_steeringControlType` field expanded from 2 bits to 3 bits in the dbc. It is in the same position, the field simply expands by 1 bit into previously unknown/unmapped bits which we can assume were not in use before.

This change first appeared in FSD v14 builds, as the expansion of the field was to allow a new value `4 = FSD` to be set as the previous field could only go up to 3, but after some time this change found it's way to normal vehicle builds from update 2026.8.6 onwards and thus is no longer a FSD v14 specific change as was previously noted in the code here.

The resulting behaviour of openpilot when assuming a 2-bit field when it was actually 3-bit resulted in the following behaviour:
- Engaging openpilot would result in `LANE_KEEP_ASSIST` steering instead of `ANGLE_CONTROL` steering as `0b01` would be sent by openpilot but read by the car as `0b010`
- Vehicle LKAS ("Lane Departure Avoidance") events would result in the FSD v14 suspect being triggered, resulting in openpilot forcefully disengaging and locking out the user with `invalidLkasSetting` until the device went off-road.
- In the case of sunnypilot with it's Coop Steering setting, enabling that would also result in complete loss of steering as `0b10` would be sent, which was read by the car as `0b100` aka FSD steering which apparently does not work

The changes in this PR essentially allow us to swap between the two different behaviours, the new one with the 3-bit field and the "legacy" one with the 2-bit field, determined using a fixed `LEGACY_DAS_STEERING_FW` list, similar to the previous `FSD_14_FW` list but with the behaviour inverted, as it seems Tesla will now be using 3-bit going forwards in all updates to all cars. All firmwares that were previously classified as FSD v14 have been removed from the legacy list to maintain behaviour with the existing code, and additionally some others that have been since confirmed to be using 3-bit.

The test around comparing the firmware list against the suspected FSD v14 rule has been simplified to just ensuring that all entries within the list are also present in the `FW_VERSIONS` list so they don't get orphaned, but the logic is too complex to accurately add heuristics too and given we can expect all new firmwares from Tesla to include the 3-bit field now that it is in the standard release tracks and no longer exclusive to FSD v14 this shouldn't be an issue going forwards.

The previous FSD v14 handling has been stripped out as part of this PR as it runs on incorrect assumptions that the values themselves had changed, but it was actually the bit size of the field that had changed. Without that knowledge, it appeared as if `ANGLE_CONTROL` (`0b01`) had become `0b10` and `LANE_KEEP_ASSIST` (`0b10`) had become `0b01` when in fact they had become `0b100` (new FSD value) and `0b010`, the same value as before just shifted. Autosteer on these vehicles also went seemingly missing with `0b01` becoming `0b00` but it had actually just moved to the new bit `0b001`.

A [sunnypilot variant of these changes](https://github.com/sunnypilot/opendbc/compare/29d9bf3b85101c4c1b7ffcb7d56c263b43563892...1dae92c4c139e65430cec342363dfa2d03f36814) was installed and tested on several vehicles using both 2-bit and 3-bit firmwares and in one case testing directly before and after upgrading and the new code correctly both read and set the correct values according to several validation routes shared privately with myself.

Some new firmware strings were also added that have been confirmed to all be using 3-bit firmware and working against this fix.

An openpilot fork is available at https://github.com/AmyJeanes/openpilot-upstream/tree/tesla-das-steering-3bit containing only these changes from openpilot/master.

### Routes:

Note: These routes were tested on the original version of this PR with the 2 different dbc files. I am trying to gather routes from people on the new version now, but the behaviour should be identical as proved by the unit tests.

- Old firmware stock AP: 4b3cbd6038d07ca6/00000209--31be0354df
- Old firmware openpilot engaged: 4b3cbd6038d07ca6/0000020a--0029635bdb
- New firmware openpilot engaged / LKAS pass through (courtesy of Tinker2A) : c0fe59cc1ebbd1b3/00000014--3db99b3a52
- Newly added firmware string stock AP: 8ce570e8e64f5bd9/00000016--220a42a198
- Newly added firmware string openpilot engaged: 8ce570e8e64f5bd9/0000000a--0edd90c9cb


AmyJeanes:
@sshane this is now ready for review, I've added routes from my own vehicle on the older firmware and a route showing everything working from a car on the newer 3-bit firmware. Please let me know if you need any other routes or anything.

opendbc/dbc/tesla_model3_party_legacy.dbc
sshane, line 1: Just one DBC, and set values in reference to the new 3 bit signal? Or use a generator so we don't duplicate message definitions

AmyJeanes: Will look into!

AmyJeanes: Generator seemed more effort than it was worth given the couple usages just needing a simple bit shift so decided to do that instead but let me know what you think.

Please note I haven't yet tested this in vehicle as my car is in the shop, but I've asked others if they could help test this, and I will test it as soon as I can and let you know.

AmyJeanes: Just as an update - car is still in the shop - hoping to get it back within the next week or so then I will test this. Unfortunately, the repair shop has upgraded my vehicle firmware to 2026.20.3 - so I will no longer be able to test the legacy path directly.

AmyJeanes: Alright. I have been able to test this properly now I've got my car back - at least the happy new firmware path. I do not know of anyone still running old enough Tesla firmware to still have the 2 bit field, but the previous version of these changes worked perfectly with that and the tests ensure the logic is intact, just the method is different now.

@sshane are there any blockers preventing this from being merged in now? I've also added a couple of new firmware fingerprints from my own vehicle I've seen as well as others reported to me that use the new 3 bit field.

opendbc/car/tesla/fingerprints.py
```diff
@@ -11,6 +11,7 @@
       b'TeM3_E014p10_0.0.0 (16),EL014.17.00',
       b'TeM3_ES014p11_0.0.0 (25),ES014.19.0',
       b'TeM3_E014p10_0.0.0 (24),E014.20.2',
+      b'TeM3_E014p10_0.0.0 (24),EL014.20.2',
```
sshane, line 14: How sure of this list are you? May be missing this 2019 Model 3 https://github.com/commaai/opendbc/pull/3323

AmyJeanes: I'm very confident that any Model 3 and Model Y since 2026.8.6 which is what all of these new additions are, use the 3-bit steering control type

sshane: I was wrong, it was HW2.5

opendbc/dbc/tesla_model3_party_legacy.dbc
AmyJeanes: Have spoken to Shane, the main blocker was about better detection of old vs new, the list based approach in this PR was insufficient and Shane will be replacing this with a new can based heuristic.

A new PR has been created to split out the core changes from this one but without the table changes so it uses the current FSD v14 heuristic just renamed: https://github.com/commaai/opendbc/pull/3797

AmyJeanes:
This PR is superseded by https://github.com/commaai/opendbc/pull/3797 as @sshane has asked me to split out the core changes from it into a new PR as the list of old firmwares is not good enough from a safety pov and needs proper heurestics which he will be handling in a follow up PR after my new one.

The new fingerprints that were added in my PR have also been raised as a separate PR, although the CI is currently failing on that due to these incorrect heuristics not flagging them as 3 bit / FSD v14 so it will need to wait until that's sorted: https://github.com/commaai/opendbc/pull/3798

I've decided to create a new PR for that instead of replacing this one as a historical record and also as it is a complete fix that can be and is used within forks until @sshane has finished working on the new heuristics.

PR #3399, opened 2026-05-20
curvature steering safety

elkoled:
Validation
* Dongle ID:
* Route:


elkoled:
integrating this into id4 port

PR #3406, opened 2026-05-27
Nissan: log EPS faults

sshane:
override fault right around ~4.25:

image


sshane:
Some of those segments look legit 
image

PR #3412, opened 2026-05-28
Nissan: rate limit max torque

sshane:


sshane:
Caused more faults (because the winddown was too slow?)

PR #3414, opened 2026-05-28
unify MAX_LAT_ACCEL and JERK

elkoled:
Unify MAX_LAT_ACCEL and MAX_LAT_JERK to be used in all vehicle model and curvature based cars.


opendbc/safety/tests/test_ford.py
```diff
@@ -360,6 +360,30 @@ def test_curvature_rate_limits(self):
             self._set_prev_desired_angle(sign * (curvature_offset + initial_curvature))
             self.assertEqual(should_tx, self._tx(self._lat_ctl_msg(True, 0, 0, sign * (curvature_offset + desired_curvature), 0)))
 
+  def test_curvature_accel_boundary(self):
```
sshane: why is this different still for CAN FD?

elkoled: We can only remove the CAN FD gate with the curvature safety PR removing the angle rate limits, so it is needed for this change.

sshane: I meant only to unify the comments and accel variables

elkoled: ok, done

elkoled:
@sshane good to go?

opendbc/car/lateral.py
```diff
@@ -1,26 +1,40 @@
...
+# - ISO 11270
 ISO_LATERAL_ACCEL = 3.0  # m/s^2
 ISO_LATERAL_JERK = 5.0  # m/s^3
 
+# - Common angle safety limits
+# Add extra tolerance for average banked road since safety doesn't have the roll
+ANGLE_AVERAGE_ROAD_ROLL = 0.06  # ~3.4 degrees, 6% superelevation. higher actual roll lowers lateral acceleration
+ANGLE_MAX_LATERAL_ACCEL = ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * ANGLE_AVERAGE_ROAD_ROLL)  # ~3.6 m/s^2
+# Lower than ISO 11270 lateral jerk limit (5.0 m/s^3) with bank tolerance, matches safety MAX_LATERAL_JERK
+ANGLE_MAX_LATERAL_JERK = 3.0 + (ACCELERATION_DUE_TO_GRAVITY * ANGLE_AVERAGE_ROAD_ROLL)  # ~3.6 m/s^3
```
sshane: pre-fixed with ANGLE_ to not confuse any future importers/readers, but maybe just don't export this at all

opendbc/car/lateral.py
```diff
@@ -1,26 +1,40 @@
...
-  MAX_LATERAL_ACCEL: float = 0
-  MAX_LATERAL_JERK: float = 0
+
+@dataclass
+class AngleSteeringLimitsVM:
+  # using apply_steer_angle_limits_vm
+  STEER_ANGLE_MAX: float
+  MAX_LATERAL_ACCEL: float = ANGLE_MAX_LATERAL_ACCEL
+  MAX_LATERAL_JERK: float = ANGLE_MAX_LATERAL_JERK
   MAX_ANGLE_RATE: float = math.inf
```
sshane, line 37: new dataclass with all the defaults you'll want for future angle VM ports

opendbc/car/tesla/values.py
```diff
@@ -106,23 +106,10 @@ class CANBUS:
...
-    ([], []),
-    ([], []),
-
-    # Vehicle model angle limits
-    # Add extra tolerance for average banked road since safety doesn't have the roll
-    MAX_LATERAL_ACCEL=ISO_LATERAL_ACCEL + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL),  # ~3.6 m/s^2
-    MAX_LATERAL_JERK=3.0 + (ACCELERATION_DUE_TO_GRAVITY * AVERAGE_ROAD_ROLL),  # ~3.6 m/s^3
-
     # limit angle rate to both prevent a fault and for low speed comfort (~12 mph rate down to 0 mph)
     MAX_ANGLE_RATE=5,  # deg/20ms frame, EPS faults at 12 at a standstill
```
sshane, line 114: now you don't have to parse junk that's not used or duplicated

opendbc/car/lateral.py
sshane: wonder if apply_steer_angle_limits_vm should be a method on the dataclass to move logic close to variable definitions

sshane: Also interface.py should just read CarControllerParams (make it a required common base class) and set CP.steerControlType based on that, this reduces the number of places you have to add the steering config for new ports. similar to PlatformConfig

PR #3415, opened 2026-05-29
Nissan: filter angle to reduce jitter

sshane:
Has same problem as new angle HKGs, Nissan steering is too reactive. Causes grinding sound when coming to a stop, and general oscillation during turns at medium speeds


adeebshihadeh:
Can we add this to the lateral maneuver reports?

sshane:
From what I've seen, a rapidly stepping request can induce some noise for Nissan, maybe the same for HKG

PR #3417, opened 2026-05-29
Nissan: use high precision steering angle

sshane:
You can see when stock LKAS is off, it's following the torque sensor message, not the angle sensor message we've been using. 

Leaf:

image

X-Trail:

image

Especially bad on the Altima we rented!

image


sshane:
How did test models not fail?! Didn't update safety

PR #3435, opened 2026-06-04
move conditions for belowEngageSpeed into opendbc

belm0:
Previously `belowEngageSpeed` conditions where calculated by `car_specific.py`.

Added `CruiseState.belowEngageSpeed` and moved conditions into carstate logic.

### DIFF
--- a/opendbc/car/car.capnp
+++ b/opendbc/car/car.capnp
@@ -202,7 +202,6 @@ struct CarState {
   espActive @51 :Bool;
   vehicleSensorsInvalid @52 :Bool;  # invalid steering angle readings, etc.
   lowSpeedAlert @56 :Bool;  # lost steering control due to a dynamic min steering speed
-  belowEngageSpeed @61 :Bool;
   blockPcmEnable @60 :Bool;  # whether to allow PCM to enable this frame
 
   # cruise state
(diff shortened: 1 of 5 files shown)


opendbc/car/car.capnp
```diff
@@ -202,6 +202,7 @@ struct CarState {
   espActive @51 :Bool;
   vehicleSensorsInvalid @52 :Bool;  # invalid steering angle readings, etc.
   lowSpeedAlert @56 :Bool;  # lost steering control due to a dynamic min steering speed
+  belowEngageSpeed @61 :Bool;
   blockPcmEnable @60 :Bool;  # whether to allow PCM to enable this frame
 
   # cruise state
```
sshane, line 207: I think it's worth exploring grouping the fields into structs like CruiseState. unclear what each signal should be used for by name alone

belm0: ~I’ll add a TODO, but can we avoid expanding the scope of this PR?  Fixing a known regression and reducing car_specific is enough of a win here.~

belm0: moved field into CruiseState struct

belm0:
not stale, I beg to differ

PR #3437, opened 2026-06-06
VW ID.4 radar parser

adeebshihadeh:
pulled out of the ID4 port as it's unclear how raw these tracks are. we need to investigate the source and quality of the tracks.

Route:
aebd8f1d4ea16066/00000009--b31e222338


opendbc/car/volkswagen/radar_interface.py
sshane, line 1: make sure x is in frame of front of car, see radard. point is showing inside car

image

opendbc/car/volkswagen/interface.py
```diff
@@ -62,6 +64,9 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
       else:
         ret.networkLocation = NetworkLocation.fwdCamera
 
+      if ret.networkLocation == NetworkLocation.gateway:
+        ret.radarUnavailable = 0x24F not in fingerprint[2]  # MEB_Distance_01
+
```
sshane: @elkoled Can we remove this? We want radar parser to output CAN error if it didn't find message rather than silently falling back to vision only

sshane: Do we support camera harness with MEB at all?

adeebshihadeh: no

elkoled: we have to disable AEB for long support, so currently only gateway harness is allowed.

sshane:
@elkoled I covered the radar with some foil and drove around. Any of these signals worth registering a car fault for in carstate.py? Dash said something about ACC or safety systems unavailable, sensor covered. AEB could be impacted. We don't display a permanent error for this right now:

https://connect.comma.ai/f73c01590368ee5b/00000024--aeb38e2060

image

elkoled:
we could use MEB_Distance_01.Distance_Status == 3 for carFaultedNonCritical.

sshane:
It's probably not nonCritical, it can't see leads. It doesn't engage with stock ACC. Does EA_OFF mean no emergency assist?

elkoled:
correct

elkoled:
ACC_STATUS_ACC==6 is "reversible fault in ACC system"

image

sshane:
great, wanna log that as accFaulted?

elkoled:
yeah sounds good

elkoled:
Add ACC fault:
https://github.com/commaai/opendbc/pull/3551

PR #3441, opened 2026-06-07
Toyota: radar unavailable bits

sshane:
Showing the startup on a TSS-P Prius, this bit rises before any track shows information:

image

Interestingly, shows a low speed lockout during startup:

image

On a Brazilian TSS2, showing it goes low when coming to a stop and dropping all points:

image


opendbc/car/toyota/radar_interface.py
```diff
@@ -59,6 +62,9 @@ def _update(self, updated_messages):
     if not self.rcp.can_valid:
       ret.errors.canError = True
 
+    if self.rcp.vl['STATUS_MSG']['RADAR_AVAILABLE'] == 0 or self.rcp.vl['STATUS_MSG']['RADAR_AVAILABLE_2'] == 0:
+      ret.errors.radarUnavailableTemporary = True
```
sshane: need to remove noEntry and softDisable

PR #3445, opened 2026-06-08
safety: add option to ignore frequency check for RX checks

sunnyhaibin:
Add optional `ignore_frequency_check` flag to `CanMsgCheck` for RX messages that are physically broadcasted below the 10Hz minimum enforced in #3359.


adeebshihadeh:
Discussed on Discord.

PR #3451, opened 2026-06-08
Add new fingerprint for CUPRA Born 2023 (Volkswagen MEB Gen 1)

infiniteCable2:
This PR adds a CUPRA Born fingerprint for Volkswagen MEB plattform. 

Validation
Dongle ID: d4dd69160a48f11f
Route: d4dd69160a48f11f/00000003--9cfe00cb74


opendbc/car/volkswagen/fingerprints.py
```diff
@@ -1090,6 +1090,17 @@
       b'\xf1\x875Q0907572R \xf1\x890771',
     ],
   },
+  CAR.CUPRA_BORN_MK1: {
+    (Ecu.srs, 0x715, None): [
+      b'\xf1\x871EA959655EH\xf1\x890381',
+    ],
+    (Ecu.fwdCamera, 0x74f, None): [
+      b'\xf1\x871EA980654L \xf1\x895332',
+    ],
```
elkoled:
```suggestion
```

elkoled: fwdCamera can't be used for fingerprinting on VW as it is used for data collection in extra_ecus

elkoled (changes requested):

opendbc/car/volkswagen/fingerprints.py
infiniteCable: > fwdCamera can't be used for fingerprinting on VW as it is used for data collection in extra_ecus

Thank you. I have removed the fwdCamera firmware.

elkoled:
Thanks for the PR!

PR #3455, opened 2026-06-09
add laneChange to AudibleAlert

elkoled:
Adds laneChange to the AudibleAlert enum so openpilot can play a sound when the user confirms a lane change.

### DIFF
--- a/opendbc/car/car.capnp
+++ b/opendbc/car/car.capnp
@@ -433,6 +433,8 @@ struct CarControl {
       prompt @6;
       promptRepeat @7;
       promptDistracted @8;
+
+      laneChange @9;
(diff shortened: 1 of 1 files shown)


opendbc/car/car.capnp
```diff
@@ -433,6 +433,8 @@ struct CarControl {
       prompt @6;
       promptRepeat @7;
       promptDistracted @8;
+
+      laneChange @9;
```
adeebshihadeh, line 437: oh we shouldn't be making changes to opendbc for this. we should just move (redefine to not break capnp) this to log.capnp

i think this was originally here since we used the car for sounds

PR #3457, opened 2026-06-10
Add Volkswagen ID.3 2025 SAIC (MEB Gen 2) fingerprint

remoonlight:
Adds firmware fingerprint for Volkswagen ID.3 2025 SAIC (MEB Gen 2).


elkoled:
we need to add VW MEB Gen2 support before adding fingerprints

PR #3459, opened 2026-06-11
Toyota radar: remove valid track count

sshane:
This valid cnt was added here as a "ghost fix" but it's unclear what that exactly means https://github.com/commaai/one/pull/172

The last few mismatches from removing VisionRadar stack in xx for Toyota are stopped ego and lead cases where no "valid" points properly capture lead, either from lead being really close or at a weird angle.

Here is one that looks like an easy case, but the only stable track that has its distance is not "VALID": RAV4 TSS-P 5e9be1efcd6ca25c|2023-04-01--12-03-23/19

image

However interestingly, the track A msg NEW_TRACK signal goes high when its track B score rises! But not every VALID and NEW_TRACK coincides with high score which is strange. Regardless it looks like the Toyota radar treats it as an or?

image


sshane:
Another case, track 0 has the lead the entire segment but valid is spotty and score is 100 (maxed out). When valid rises, new track doesn't!

610c0e2a1dda7c3e|2020-12-20--15-34-06/8

image

This PR would fix the "no vradar" lines to include the lead at a standstill

image

sshane:
I plotted a few segments with many new added tracks, and they mostly look like this:

Green points are valid, blue points are not valid but score > 50 (new with this PR), and red are not valid. Before this PR, blue would be not valid. The blue points are all stationary points from the tunnel the user drives through which looks bad as their y is dead center, however both online and offline filter them out. There are also valid points that are detected as stationary, so we just increase the number of stationary points.

Online filters these points out by checking if they are close to model predicted lead's distance, y position and speed. Offline LeadCar drops points where they were stationary while ego was moving.

https://github.com/user-attachments/assets/cdd3654e-a7d2-4003-ba5b-eb892237f348

image

Re-ran MeshRadar and LeadCar, they are filtered out properly:

image

sshane:
Would include points where vRel~=0 while driving and x is between <1m and 10m :(

PR #2928, opened 2025-12-03
Chrysler: reliable fuzzy fingerprinting

ReyNeill:
- add Chrysler brand-specific fuzzy FW matching using platform part numbers (revision-agnostic) across cluster/ABS/radar/EPS
- wire matcher into Chrysler FW query config
- add regression tests for platform ECU coverage and revision-tolerant matching

Testing
```
python -m pytest opendbc_repo/opendbc/car/chrysler/tests
python -m pytest opendbc_repo/opendbc/car/tests/test_fw_fingerprint.py
```

PR #3496, opened 2026-06-23
Chrysler: reliable fuzzy fingerprinting (#1092)

Acley143:
Adds fuzzy FW matching for Chrysler/Jeep/Ram/Dodge so new firmware revisions get identified without needing exact DB entries.

Addresses #1092. Builds on the analysis in #3018.

Three prior attempts (#2928, #3018, #3329) all assumed the last 2 chars of the 10-char firmware string are the revision suffix. Strip those, match on the remaining 8. That assumption is wrong. Validated against ~22 months of real firmware from opendbc's fingerprint bot (Aug 2024 snapshot vs current master), the 2-char strip correctly identifies only 1/5 (20%) of genuinely unseen real-world revisions. Chrysler part numbers themselves change across updates for the same ECU/platform, not just the tail.

Swept strip lengths 2-6 against the same dataset, requiring agreement across all PLATFORM_CODE_ECUS (combinationMeter, abs, fwdRadar, eps) simultaneously to keep collision risk low:

strip=2: 1/5 correct (20%), 0 wrong, 4 no-match
strip=3: 2/5 correct (40%), 0 wrong, 3 no-match
strip=4: 4/5 correct (80%), 0 wrong, 1 no-match <- chosen
strip=5: 3/5 correct (60%), 0 wrong, 1 ambiguous
strip=6: 3/5 correct (60%), 0 wrong, 1 ambiguous

strip=4 gets the best real-world recall with zero dangerous misidentifications at any setting. Failures degrade safely to no match and fall back to exact-match behavior rather than matching the wrong platform.

Sample size is n=5, capped by opendbc's available git history since the repo split from openpilot in Aug 2024. Worth expanding with more historical data before merge if possible, flagging this upfront.

Tests include 4 held-out regression cases built from real bot-collected firmware, not synthetic single-revision mutations. Each holds out only the genuinely novel ECU readings from the offline DB, mirroring how real cars update firmware piecemeal rather than all ECUs at once.

Validation
Dongle ID:
Route:

### DIFF
--- a/opendbc/car/chrysler/tests/test_chrysler.py
+++ b/opendbc/car/chrysler/tests/test_chrysler.py
@@ -1,4 +1,4 @@
-import pytest
+import unittest
 
 from opendbc.car.chrysler.fingerprints import FW_VERSIONS
 from opendbc.car.chrysler.values import CAR, PLATFORM_CODE_ECUS, get_platform_codes, match_fw_to_car_fuzzy
@@ -9,7 +9,7 @@ Ecu = CarParams.Ecu
 CarFw = CarParams.CarFw
 
 
-class TestGetPlatformCodes:
+class TestGetPlatformCodes(unittest.TestCase):
   def test_extracts_part_number(self):
     # Standard Chrysler FW format: 8 digit part number + 2 char revision
     codes = get_platform_codes([b'68227902AF'])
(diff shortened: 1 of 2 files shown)


Acley143:
Opened #3496 with a fuzzy fingerprinting implementation for Chrysler/Jeep/Ram/Dodge.

@sshane picking up from where #3018 left off since you asked for validation against real Chrysler FP PRs and that never landed. I ran that validation. Used opendbc's own fingerprint bot history (Aug 2024 snapshot vs current master, ~22 months of real field-collected firmware) and checked whether the matcher would have correctly identified genuinely unseen firmware revisions using only the older snapshot as the offline DB.

The original approach of stripping the last 2 chars as a revision suffix and matching on the 8-digit part number only gets 1/5 (20%) correct on real data. Chrysler part numbers themselves change across updates for the same ECU/platform, not just the trailing 2 chars. I swept strip lengths 2-6 and found stripping 4 chars gives 4/5 (80%) correct with zero dangerous misidentifications at any setting tested. Failures degrade safely to no match, never to the wrong platform.

Full methodology and the strip-length sweep table are in the PR description. Tests include 4 regression cases built from real bot-collected firmware, not synthetic mutations. All CI checks passing.

Being upfront: validation sample is small (n=5), capped by how far back opendbc's git history goes since the repo split from openpilot in Aug 2024. Happy to dig into openpilot's pre-split history for more data points if that would help get this over the line.

adeebshihadeh:

Issue #2234, opened 2025-05-14
safety: speed mismatches should make rx go invalid


sshane:
Instead it just sets controls_allowed to false

sshane:
the check is now common in: https://github.com/commaai/opendbc/pull/2294 but would still be nice to make rx go invalid

PR #3499, opened 2026-06-24
safety: speed mismatch invalidates rx checks

parsa-faraji:
## What

A speed mismatch between two speed sources (`speed_mismatch_check`) previously only set `controls_allowed = false`. This implements the behavior requested in #2234: a mismatch now also marks the safety rx checks invalid (`safety_rx_checks_invalid`), so it's treated like other rx faults and the mismatched data itself is flagged as suspect rather than producing only a silent disengage.

## How

- Track the mismatch in a persisted `speed_mismatch` flag, and the existing lagging/invalid-message result in a persisted `rx_msgs_invalid` flag.
- Compute `safety_rx_checks_invalid = rx_msgs_invalid || speed_mismatch`, recomputed in both `safety_tick()` (1 Hz) and `speed_mismatch_check()` (per message). Combining the two sources explicitly means a valid speed message can't clear (mask) a real lagging/invalid-message fault between ticks.
- Reset the new flags on safety mode init.

## Testing

- Added `get_safety_rx_checks_invalid()` and extended the existing `test_rx_hook_speed_mismatch` (common + Tesla override) to assert the flag tracks the mismatch across the full speed sweep, for every mode with a second speed source (Tesla, Rivian, Ford).
- `python -m unittest discover -s opendbc/safety/tests` passes (3196 tests); 100% safety line coverage maintained.

The issue also notes Tesla compares a signed speed with an unsigned speed; I left that out to keep this PR focused on the rx-invalid behavior and can address it as a follow-up.

### DIFF
--- a/opendbc/safety/declarations.h
+++ b/opendbc/safety/declarations.h
@@ -271,6 +271,8 @@ extern bool vehicle_moving;
 extern bool acc_main_on; // referred to as "ACC off" in ISO 15622:2018
 extern int cruise_button_prev;
 extern bool safety_rx_checks_invalid;
+extern bool speed_mismatch;            // two speed sources disagree beyond threshold
+extern bool rx_msgs_invalid;           // safety rx messages lagging/invalid (computed at 1Hz)
 
 // for safety modes with torque steering control
 extern int desired_torque_last;       // last desired steer torque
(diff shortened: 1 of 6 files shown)


adeebshihadeh:

PR #3519, opened 2026-07-03
Honda - 2025 Civic Thailand fingerprint

mvl-boston:
7849a2cc2aabe045/00000000--72e9048b58


mrajaypalsingh1-coder (commented):
Great work on adding support for the 2025 Civic Thailand model. Thanks for the contribution!"

PR #3533, opened 2026-07-11
rivian: use full range on 4-bit message counters

Jiming-C:
`ACM_lkaHbaCmd_Counter` and `ACM_longitudinalRequest_Counter` are 4-bit fields (DBC range `[0|15]`), but they're packed with `frame % 15`, which skips value 15 and makes the counter step +2 (mod 16) at every wrap instead of +1. Use `frame % 16` so the counter uses its full range, matching every other brand's 4-bit rolling counters. The checksum is computed after the counter is set, so the messages stay self-consistent.

Validation
* Dongle ID: N/A (no Rivian on hand; fix is a 2-value counter-range correction, DBC-confirmed 4-bit)
* Route: N/A

### DIFF
--- a/opendbc/car/rivian/riviancan.py
+++ b/opendbc/car/rivian/riviancan.py
@@ -21,7 +21,7 @@ def create_lka_steering(packer, frame, acm_lka_hba_cmd, apply_torque, enabled, a
   )}
 
   values |= {
-    "ACM_lkaHbaCmd_Counter": frame % 15,
+    "ACM_lkaHbaCmd_Counter": frame % 16,
     "ACM_lkaStrToqReq": apply_torque,
     "ACM_lkaActToi": active,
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
@sshane @lukasloetkolben was this a reason for this?

Jiming-C:
**Context on why I changed it:** both counters are 4-bit in the DBC (range [0|15])
with nothing marking 15 as reserved, and every other 4-bit counter in the repo
uses % 16 (chrysler, nissan, hyundai, subaru...). With % 15 the counter never
uses 15 and wraps a frame early, so anything checking for a +1 increment sees a
jump every 15 frames. The checksum is packed after the counter, so % 16 stays
consistent.

If there's a Rivian-specific reason for % 15 I'm happy to close it. Just looked
like a typo to me.

lukasloetkolben:
Bildschirmfoto 2026-07-11 um 23 32 05

The counter in this message is a 0–14 counter (15 states)
So `frame % 15` is the correct modulus.
Happy to be proven wrong if you have a log showing the stock ACM actually emitting a counter with 15

lukasloetkolben:
All messages have a counter ranging from 0 to 14. 
You can close this PR

Jiming-C:
Ah, makes sense. The 4-bit DBC field threw me off.

Thanks for pulling the screenshot and the
context. 

Closing this out.

adeebshihadeh:
It'd still be a good idea to merge some comments so someone else doesn't try to "fix" this in the future.

PR #3135, opened 2026-02-16
VW: Proper Long Override

tealtwo:
This PR aims to fix https://github.com/commaai/opendbc/issues/2874 and is pending tests by me, therefore it is marked as draft. It corrects openpilot longitudinal control to use the proper ECU cruiseOverride state instead of disengaging cruise state and reengaging for accelerator override, it will also be pending the PRs mentioned by jyoung in #2874 to allow sending acceleration value that isn't the inactive value when overriden.

PR #3536, opened 2026-07-12
VW MQB: release standstill hold on gas override

### LINKED ISSUE #1159: VW experimental long: Sudden Deceleration From Stop
### Describe the bug

Sudden braking when the gas pedal is pressed from a stop and OP engaged. 

The issue occurred in Segment 26.

Previous conversation for reference

```
you were engaged down to a standstill, and successfully went into standstill (a specific state of the ACC subsystem)
[4:23 AM]
you then pressed the gas pedal, and you appear to have disengage on gas enabled
[4:24 AM]
the drivetrain/stopping coordinator will drop the EPB if ACC is disengaged while at standstill
[4:24 AM]
we actually don't have control over that
it's a feature with the car, that I'll have to think about how to deal with
[4:27 AM]
if you were to use stock ACC to follow someone to a stop, and while stopped, press the cancel button, the EPB would drop to prevent rollaway. this is by design.
[4:27 AM]
there are some different ACC control states, which I don't use right now, that are meant for use when the driver is using accel to override. it may be time to do that.
[4:28 AM]
right now this looks to the car like a disengage, and I probably need to show it as driver override to avoid the EPB drop
[4:28 AM]
there are some other nuances to why I'm not doing so now, and I will have to think about them
[4:30 AM]
TSK-side handling of the EPB is a safety thing, like, if the ACC radar suddenly failed/broke/crashed while you were behind a stopped car, the process of dropping the EPB is automatic and taken care of by the TSK independent of the ACC radar
[4:31 AM]
that's why EPB is required for stop-and-go cars, because the driver may not be fully attentive if the car has been at standstill for a while, so the car needs to safely handle the case of radar failure
```

VW Tiguan R-Line 4Motion 2020 (US)

4e144f64a2535890|2023-03-27--18-14-02--26

### openpilot version

master-ci V0.9.1

### Additional info

@jyoung8607

andiradulescu:
Continuation of @jyoung8607’s [mqb-accel-override](https://github.com/jyoung8607/openpilot/compare/fe6aff01311ca3c5a4a549b0eb5d081fa1740f86..mqb-accel-override)

Fixes https://github.com/commaai/opendbc/issues/1159

Keep ACC in the driver-override state and request standstill hold release when accelerating from a stop while engaged. Send VW’s inactive acceleration value while openpilot longitudinal control is overridden.

Although happens way less often than in 2023/2024, it still happens to brake for ~1s when overriding from a standstill.

Validation
* Dongle ID: ca83aca15dc27f35
* Route: ca83aca15dc27f35/000000a2--caa169b80e/22

EDIT: this might not be needed anymore if https://github.com/commaai/opendbc/pull/3135 and https://github.com/commaai/opendbc/pull/3165 get merged.

### DIFF
--- a/opendbc/car/volkswagen/carcontroller.py
+++ b/opendbc/car/volkswagen/carcontroller.py
@@ -159,7 +159,7 @@ class CarController(CarControllerBase):
 
         else:
           acc_control = self.CCS.acc_control_value(CS.out.cruiseState.available, CS.out.accFaulted, CC.enabled, CC.longActive)
-          accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if CC.longActive else 0)
+          accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if CC.longActive else self.CCP.ACCEL_INACTIVE)
           driver_gas_override = CC.enabled and CS.out.gasPressed
           start_control_required = CS.esp_hold_confirmation or CS.out.vEgo < self.CP.vEgoStopping
           starting = (actuators.longControlState == LongCtrlState.pid or driver_gas_override) and start_control_required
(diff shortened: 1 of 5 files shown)


sshane:
Can you see if MQB could just use the MEB long state machine class?

PR #3541, opened 2026-07-14
VW MQB: use torque controller instead of PID controller

tealtwo:
not sure if this will improve steering or keep it roughly the same

### DIFF
--- a/opendbc/car/volkswagen/interface.py
+++ b/opendbc/car/volkswagen/interface.py
@@ -104,11 +104,7 @@ class CarInterface(CarInterfaceBase):
       ret.steerActuatorDelay = 0.3
     else:
       ret.steerActuatorDelay = 0.1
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Feel free to reopen once some validation is done that shows this is better.

tealtwo:
> Feel free to reopen once some validation is done that shows this is better.

I don’t see why torque controller* shouldn’t be used here. The current PIF configuration is a waste. Other platforms here use torque controller as well.

adeebshihadeh:
Are you familiar with Chesterton's Fence? https://fs.blog/chestertons-fence/

tealtwo:
> Are you familiar with Chesterton's Fence? https://fs.blog/chestertons-fence/

yes… I’ll go find someone to get a route….

PR #3546, opened 2026-07-15
cleanup unused radar fields

adeebshihadeh:


opendbc/car/toyota/radar_interface.py
```diff
@@ -87,9 +87,6 @@ def _update(self, updated_messages):
           self.pts[ii].dRel = cpt['LONG_DIST']  # from front of car
           self.pts[ii].yRel = -cpt['LAT_DIST']  # in car frame's y axis, left is positive
           self.pts[ii].vRel = cpt['REL_SPEED']
-          self.pts[ii].aRel = float('nan')
-          self.pts[ii].yvRel = float('nan')
-          self.pts[ii].measured = bool(cpt['VALID'])
```
adeebshihadeh, line 92: fyi @sshane should this just generate an error?

PR #3553, opened 2026-07-17
VW MEB: enable long as default

sshane:
VW MEB camera harness will no longer work at all, but we don't care about this configuration. They've always been dashcam mode'd. It's only to detect an incorrect harness plugged in

Removed pcmCruise port code in safety and carstate.py, longitudinal is the only path


elkoled:
LGTM

opendbc/car/volkswagen/interface.py
```diff
@@ -115,12 +115,17 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
...
-    if ret.flags & VolkswagenFlags.MEB:
-      ret.longitudinalActuatorDelay = 0.5
-      ret.longitudinalTuning.kiBP = [0., 30.]
-      ret.longitudinalTuning.kiV = [0.4, 0.]
+    if ret.flags & VolkswagenFlags.MEB and ret.networkLocation == NetworkLocation.gateway:
+      ret.openpilotLongitudinalControl = True
+      ret.longitudinalActuatorDelay = 0.3
+      # ret.longitudinalTuning.kiBP = [0., 30.]
+      # ret.longitudinalTuning.kiV = [0.4, 0.]
+      safety_configs[0].safetyParam |= VolkswagenSafetyFlags.LONG_CONTROL.value
```
adeebshihadeh: shouldn't need to set this anymore

opendbc/car/volkswagen/interface.py
```diff
@@ -115,12 +115,17 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
...
+      if ret.transmissionType == TransmissionType.manual:
+        ret.minEnableSpeed = 4.5
```
adeebshihadeh: is this a thing?

sshane: it's still a draft, pending long bug fixes

opendbc/car/volkswagen/interface.py
sshane: Will do, but note if camera harness users uncomment the dashcam only line, the port will be fully broken for them.

PR #3556, opened 2026-07-18
VW MEB: remove starting state

sshane:
This exposes bugs, but stopping and starting is smooth now: https://github.com/commaai/opendbc/issues/3557

Stock route: https://connect.comma.ai/f73c01590368ee5b/00000039--99d1e7be6c

### DIFF
--- a/opendbc/car/volkswagen/interface.py
+++ b/opendbc/car/volkswagen/interface.py
@@ -133,10 +133,9 @@ class CarInterface(CarInterfaceBase):
       ret.steerActuatorDelay = 0.07
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
-    if not (ret.flags & VolkswagenFlags.MEB):
-      ret.stopAccel = -0.55
-      ret.vEgoStarting = 0.1
-      ret.vEgoStopping = 0.5
+    ret.stopAccel = -0.55
+    ret.vEgoStarting = 0.1
+    ret.vEgoStopping = 0.1 if ret.flags & VolkswagenFlags.MEB else 0.5
     ret.autoResumeSng = ret.minEnableSpeed == -1
 
     CAN = CanBus(fingerprint=fingerprint)
--- a/opendbc/car/volkswagen/interface.py
+++ b/opendbc/car/volkswagen/interface.py
@@ -133,10 +133,9 @@ class CarInterface(CarInterfaceBase):
       ret.steerActuatorDelay = 0.07
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
-    if ret.flags & VolkswagenFlags.MEB:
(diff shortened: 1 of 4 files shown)


opendbc/car/volkswagen/mebcan.py
```diff
@@ -119,8 +119,8 @@ def get_acc_hold_type(CS, CC, starting, stopping, esp_hold, long_override, long_
       acc_hold_type = ACC_HMS_NO_REQUEST  # overriding / no request
   elif starting:
     acc_hold_type = ACC_HMS_RELEASE  # release request and startup
-  elif stopping or esp_hold:
-    acc_hold_type = ACC_HMS_HOLD  # hold or hold request
```
sshane, line 123: this `or esp_hold` seems like it does nothing, removing. lmk if you know of any state it's required @elkoled. Disabled and overriding block is above

opendbc/car/volkswagen/interface.py
```diff
@@ -133,13 +133,9 @@ def _get_params(ret: structs.CarParams, candidate: CAR, fingerprint, car_fw, alp
       ret.steerActuatorDelay = 0.07
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
-    if ret.flags & VolkswagenFlags.MEB:
-      ret.startingState = True
-      ret.startAccel = 0.8
```
sshane, line 138: We only needed this because carcontroller thought that you get a starting state without this... So we were never actually telling car we want to move.

 @adeebshihadeh let's remove it!

opendbc/car/volkswagen/mebcan.py
sshane: I don't think you need to adhere to the car's state. You just can't make certain transitions

opendbc/car/volkswagen/mebcan.py
```diff
@@ -164,7 +164,7 @@ def create_acc_accel_control(packer, bus, CCP, acc_type, acc_enabled, accel, acc
     "ACC_zul_Regelabw_oben":      0,
     "ACC_neg_Sollbeschl_Grad_02": CCP.JERK_LIMIT if acc_control in (ACC_CTRL_ACTIVE, ACC_CTRL_OVERRIDE) and not full_stop_no_start else 0,
     "ACC_pos_Sollbeschl_Grad_02": CCP.JERK_LIMIT if acc_control in (ACC_CTRL_ACTIVE, ACC_CTRL_OVERRIDE) and not full_stop_no_start else 0,
-    "ACC_Anfahren":               starting,
+    "ACC_Anfahren":               0,  # always zero, stock uses ACC_Anforderung_HMS
```
sshane, line 167: @elkoled what does this do?

PR #3563, opened 2026-07-18
rm hypothesis

adeebshihadeh:


sshane:
How fast is it? Can we do more examples?

PR #3568, opened 2026-07-20
safety: remove inactive is zero in curvature

elkoled:
This flag existed because the MEB carcontroller snapped the curvature command to raw measured curvature on disengage and on engage while winding down power to prevent EPS fault.
Not required anymore after fixing carcontroller behavior.

Route that blocked HCA_03:
ba3e5241ca8129d2/00000005--0ff3605397


sshane:
Make sure this behavior matches stock.

elkoled:
stock snaps to 0:
image

PR #3572, opened 2026-07-21
remove longitudinal stopping state tuning

adeebshihadeh:


opendbc/car/volkswagen/carcontroller.py
```diff
@@ -160,7 +160,7 @@ def update(self, CC, CS, now_nanos):
         else:
           acc_control = self.CCS.acc_control_value(CS.out.cruiseState.available, CS.out.accFaulted, CC.longActive)
           accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if CC.longActive else 0)
-          starting = actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or CS.out.vEgo < self.CP.vEgoStopping)
+          starting = actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or CS.out.vEgo < 0.25)
```
adeebshihadeh, line 163: @sshane any idea if we can remove this constant?

sshane: Yes, I'm rewriting this. It shouldn't be using vEgoStopping

PR #3579, opened 2026-07-21
VW MEB: fix long control states

sshane:
Logic shouldn't be split across 5 places (many more for pq, mqb, etc)

This fixes the going into park fault

### DIFF
--- a/opendbc/car/volkswagen/carcontroller.py
+++ b/opendbc/car/volkswagen/carcontroller.py
@@ -41,7 +41,8 @@ class CarController(CarControllerBase):
     self.packer_pt = CANPacker(dbc_names[Bus.pt])
     self.aeb_available = not CP.flags & VolkswagenFlags.PQ
 
-    self.meb_long_state_machine = mebcan.MebLongStateMachine(self.CP, self.CCP)
+    if CP.flags & VolkswagenFlags.MEB:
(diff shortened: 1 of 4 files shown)


opendbc/car/volkswagen/carcontroller.py
```diff
@@ -135,29 +136,16 @@ def update(self, CC, CS, now_nanos):
...
-                                                           CS.out.vEgoRaw * CV.MS_TO_KPH, long_override, CS.travel_assist_available))
+                                                           accel, acc_status, acc_hold_type, braking_to_stop,
+                                                           CS.out.vEgoRaw * CV.MS_TO_KPH, CS.travel_assist_available))
           self.accel_last = accel
 
         else:
+          stopping = actuators.longControlState == LongCtrlState.stopping
           acc_control = self.CCS.acc_control_value(CS.out.cruiseState.available, CS.out.accFaulted, CC.longActive)
           accel = float(np.clip(actuators.accel, self.CCP.ACCEL_MIN, self.CCP.ACCEL_MAX) if CC.longActive else 0)
           starting = actuators.longControlState == LongCtrlState.pid and (CS.esp_hold_confirmation or CS.out.vEgo < 0.25)
```
sshane, line 151: @elkoled is non-MEB the same behavior as MEB?

PR #3589, opened 2026-07-26
VW MEB: log AEB

sshane:
@elkoled do we need to rent MK1 ID4 again? The DBC is different for this message, plus it doesn't even label or log AEB so we can't easily mine


elkoled:
I checked the latest dbc, they use the same AWV signal, vw_meb.dbc is outdated.
You can move it to common

sshane:
Looks fine to merge, old FCW in MK1 is actually Active (50hz) in MK2 message 
image

PR #3602, opened 2026-07-28
VW MEB: share the ESC_50 EPB_Status enum with pre-2024

sshane:
The EPB_Status VAL_ table was only in vw_meb_2024.dbc, but pre-2024 defines the same signal at the same bits, so move it to the common file and let both generations decode it.

Confirmed on the ID.4 MK1 route: every observed value is in the enum, sitting at 0 ("offen") for the whole drive and transitioning once to 4 ("im_Lauf_schliessen") as the car is parked.

Validation
* Dongle ID:
* Route:


sshane:
Superseded by #3601, which now unifies the whole of ESC_50 into `_vw_meb_common.dbc` — including this `EPB_Status` VAL_ table — rather than moving the enum on its own.

sshane:
commenting for me, craze

PR #3605, opened 2026-07-28
VW MEB: unify HCA_03, VMM_02 and ACC_18 into the common DBC

sshane:
HCA_03 and VMM_02 are defined identically in both reference DBCs, so take those definitions verbatim, unknown signals included:

  HCA_03  vw_meb gains the Unit_Percent unit string on Power,
          vw_meb_2024 gains Unknown_01 (53|1)
  VMM_02  vw_meb gains FCW_Reaction_1 and FCW_Reaction_3, and its
          FCW_Active is renamed FCW_Reaction_2 (identical bits) to
          match 2024. Nothing referenced it - the FCW_Active read in
          carstate is AWV_03's, a different message.

ACC_18 differs between the two reference DBCs by SET_ME_0x2FE alone, exactly as opendbc did, so unify on the superset and document it. The signal reads 0 across 99572 frames of the ID.4 MK1 route and mebcan never packs it, so carrying it for both generations is inert.

Nine messages remain variant specific.

Validation
* Dongle ID:
* Route:

### DIFF
--- a/opendbc/dbc/generator/volkswagen/_vw_meb_common.dbc
+++ b/opendbc/dbc/generator/volkswagen/_vw_meb_common.dbc
@@ -285,6 +285,19 @@ BO_ 299 GRA_ACC_01: 8 Gateway
  SG_ GRA_reserveByte8 : 56|8@1+ (1,0) [0|255] "" Vector__XXX
 
 BO_ 312 IPA_01: 32 XXX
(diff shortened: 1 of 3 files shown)


sshane:
@elkoled can we unify ACC_19? The messages we have differ a lot in missing signals. They are the same in the reference DBCs

PR #3617, opened 2026-07-29
VW MEB: only mask TSK faults we are aware of

sshane:
replaces the brake-pressed mask with the two conditions that actually cause the temp fault.

- `Motor_54.Engine_On == 0` — chassis asleep, no power steering. this is why it faults in park, and why P→N stays faulted but P→D or P→R clears.
- `VMM_02.Long_Control_Inhibit == 2` — set by the VMM shortly after a hard brake. new signal, 52|2, only ever 1 or 2 on the bus.

both are held 100 ms after they clear. no rising edge tolerance.

### timing

every ID.4 segment in mkv — 1941 segments, 304 routes, ~31 h of Motor_51:

| edge | result |
|---|---|
| mask releases → TSK clears | n=100, max 59.8 ms, median 29.9 ms, zero over 100 ms |
| mask on before TSK faults | n=126, median 40.3 ms |
| TSK faults before mask | 0 |

so 10 frames covers the tail with 40 ms of margin. the earlier claim that TSK can lead by a few frames was a scanning artifact — all of those were a segment opening with TSK already faulted, before 10 Hz Motor_54 had published its first frame.

### behavior

| route | before | after |
|---|---|---|
| `00000084--3ff94803c8` (the reported false fault) | fault | none |
| `000000b5--41a8cf1b23` (P/N/R/D shuffle) | 6.3 s + 5.0 s false fault in N | none |
| `00000088--a71a108d64` (real TSK=7) | silenced entirely | 112 s reported |

the old mask silenced any fault while the brake was held, plus 3 s after, which is what hid that 112 s permanent fault. there are 7 genuine faults in the corpus and all are reported now, including a single-frame one.

also adds the gen2 `VMM_02` crc constant. without it the message never validates on MK2 and none of this does anything there.

replacement of https://github.com/commaai/opendbc/pull/3597


sshane:
going from P to N, R, D: https://connect.comma.ai/f73c01590368ee5b/000000b5--41a8cf1b23

hard braking causing VMM control inhibit: aebd8f1d4ea16066/00000084--3ff94803c8/2, f73c01590368ee5b/000000b3--67069de304

PR #3620, opened 2026-07-30
VW MEB: fix another park fault on hills

sshane:
anfahren looks like it mostly matches ESC_50->Standstill, and not time duration, so good we already do that.

https://connect.comma.ai/f73c01590368ee5b/00000039--99d1e7be6c

anfahren_dist

---

however we're missing going to ramp release after resuming, which may be faulting some cars on steep hills. stock on our ID4 does this on every resume:

you can see after 4 (resume) it go to 5 (ramp) until speed reaches ~5 kph every single time. it's also able to go back to halten if it needs to

image


sshane:
Superseded by https://github.com/commaai/opendbc/pull/3626

PR #3621, opened 2026-07-30
VW MEB: fix another park fault on hills

sshane:
Validation
* Dongle ID:
* Route:


sshane:
Superseded by https://github.com/commaai/opendbc/pull/3626

PR #3648, opened 2026-08-05
fuzzy tx

adeebshihadeh:
re: https://github.com/commaai/openpilot/issues/38533

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -247,7 +247,7 @@ class CarController(CarControllerBase):
         elif net_acceleration_request_min > 0.3:
           self.permit_braking = False
 
-        pcm_accel_cmd = float(np.clip(pcm_accel_cmd, self.params.ACCEL_MIN, self.params.ACCEL_MAX))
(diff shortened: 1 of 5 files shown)


opendbc/car/toyota/carcontroller.py
```diff
@@ -247,7 +247,7 @@ def update(self, CC, CS, now_nanos):
         elif net_acceleration_request_min > 0.3:
           self.permit_braking = False
 
-        pcm_accel_cmd = float(np.clip(pcm_accel_cmd, self.params.ACCEL_MIN, self.params.ACCEL_MAX))
+        pcm_accel_cmd = float(np.clip(pcm_accel_cmd, self.params.ACCEL_MIN, self.params.ACCEL_MAX)) if CC.longActive else 0.
```
sshane, line 250: I'm pretty sure all the carcontrollers assume accel and longActive are gated together, controlsd implicitly does this. Same for latActive. I'd hate to have some carcontrollers do this and others not just because they're not caught yet. And you can forget to do this when porting new cars

How about a singular place in opendbc, like in CI.apply? If using opendbc standalone I think someplace that can apply to all cars is better

PR #3652, opened 2026-08-05
safety: clarify angle rate limits at zero

adeebshihadeh:


adeebshihadeh:
Closing, will revisit later. @sshane we should think about how we can simplify this a bit more.

PR #3662, opened 2026-08-07
Added Fingerprint for 2023.5 Tesla Model Y (2026.26.1) due to the latest software update

gggplaya:
Validation
* Dongle ID:  9f07d1ae0718bb8a
* Route:  9f07d1ae0718bb8a/00000000--4bc40813d6


sshane:
Added here: https://github.com/commaai/opendbc/pull/3765

PR #3670, opened 2026-08-09
Remove dead code

adeebshihadeh:


opendbc/car/docs.py
```diff
@@ -78,13 +77,6 @@ def get_all_car_docs() -> list[CarDocs]:
   return sorted_list
 
 
-def group_by_make(all_car_docs: list[CarDocs]) -> dict[str, list[CarDocs]]:
```
TheSecurityDev, line 81: @adeebshihadeh isn't this used by the website?
https://github.com/commaai/website/blob/8e721b7b8938ebca9e1fbefc46510602c37638d8/scripts/compatibility.py#L8

sshane: fixed

PR #3673, opened 2026-08-09
Toyota: delay cancel command send

### LINKED ISSUE #1173: [$300 Bounty] Toyota: cancel cruise without chime

sunnyhaibin:
- May fix https://github.com/commaai/opendbc/issues/1173
- Split out from https://github.com/commaai/opendbc/pull/3336

### DIFF
--- a/opendbc/car/toyota/carcontroller.py
+++ b/opendbc/car/toyota/carcontroller.py
@@ -32,6 +32,8 @@ MAX_STEER_RATE_FRAMES = 17  # tx control frames needed before torque can be cut
 # EPS allows user torque above threshold for 50 frames before permanently faulting
 MAX_USER_TORQUE = 500
 
+CRUISE_CANCEL_DELAY_FRAMES = 10
+
 
 def get_long_tune(CP, params):
   if CP.flags & ToyotaFlags.TSS2:
@@ -57,6 +59,7 @@ class CarController(CarControllerBase):
     self.permit_braking = True
     self.steer_rate_counter = 0
     self.distance_button = 0
+    self.cancel_counter = 0
 
     # *** start long control state ***
     self.long_pid = get_long_tune(self.CP, self.params)
@@ -79,7 +82,8 @@ class CarController(CarControllerBase):
     actuators = CC.actuators
     stopping = actuators.longControlState == LongCtrlState.stopping
     hud_control = CC.hudControl
-    pcm_cancel_cmd = CC.cruiseControl.cancel
+    self.cancel_counter = self.cancel_counter + 1 if CC.cruiseControl.cancel else 0
+    pcm_cancel_cmd = self.cancel_counter > CRUISE_CANCEL_DELAY_FRAMES
     lat_active = CC.latActive and abs(CS.out.steeringTorque) < MAX_USER_TORQUE
 
     if len(CC.orientationNED) == 3:


bkonold:
@sshane Have you tested this on your Toyota? I applied this patch to my '20 RAV4 Hybrid and observed the following:

| Situation | Before | After |
|---|---|---|
| 1. Brake while ACC coasts or accelerates | beeps | silent (yay) |
| 2. Brake pressed & held while ACC is braking | beeps | beeps at 100ms (usually) |
| 3. Quick tap while ACC is braking | beeps | beeps at 100ms (usually) |
| 4. Brake at a stop | beeps | beeps at 100ms |

From what I gathered from logs, 10 frames (100ms) is not always long enough for the car to clear it's own cruise state. I did some laps and tested up to -2.25 m/s^2 braking, and saw my rav taking between 1-1.25s for the cruise state to clear in cases where ACC was already applying braking when I then pressed the pedal. The dash set speed would clear immediately, but ACC would _continue_ braking for some period of time after my initial pedal press. The harder the ACC was braking, the longer that period seemed to be.

A possible fix for (2) and (3) is to hold OP's cruise cancel from going through as long as the driver is pressing the brake pedal (if the brake pedal is released while the car's ACC is still braking, we still get TWO_BEEPS chime), but for normal cases where the driver is overriding the braking force and continues to hold the pedal down (I imagine this is the most common scenario), you will never hear any beeps as the car will clear the cruise state on its own:

```pcm_cancel_cmd = self.cancel_counter > CRUISE_CANCEL_DELAY_FRAMES and not CS.out.brakePressed```

I don't think (4) is fixable with current OP safety rules? I tested stock ACC and when the car's ACC is holding itself at a stop and the brake pedal gets pressed/held, the car does _not_ clear cruise but it _does_ put the car in standstill such that you need to press RES on the wheel or press the gas pedal for the vehicle to begin moving again.

To eliminate the beeps in case (4), correct me if I'm wrong but I think that would mean either:
1. not disengaging OP (since the car will never clear cruise state, OP will always send its own cancel triggering the beeps), OR
2. finding a silent cancel bit that doesn't otherwise cause a fault, which still hasn't been found

cc @sunnyhaibin 

Full disclaimer this is the first time I've taken a stab at modifying OP code and this investigation was only possible with Claude's help

Issue #2065, opened 2025-03-29
BYD Atto 3 Harness


qzwf:
- [ ]     which car is this for? 2022-2024 BYD ATTO3
- [ ]     schematic - Push to neo repo once confirmed working
- [ ]     beta sent to tester
- [ ]     in shop
- [ ]     Added to openpilot docs

The ADAS Camera in BYD ATTO3 is MVS4 from Veoneer. Mono Vision System 4 . Will assign proper harness name once more information is obtained.

qzwf:
![Image](https://github.com/user-attachments/assets/cbbacce1-e097-43c2-a644-fc42ece39575)

Harness Wiring as per Service Document of BYD Atto 3 

![Image](https://github.com/user-attachments/assets/00823dfe-b7ea-4516-82d4-a4b836cc19ae)

![Image](https://github.com/user-attachments/assets/130644c0-52f6-46e6-96e4-e0a002cbb3d2)

Connector Images

![Image](https://github.com/user-attachments/assets/956fd5bf-315c-40ba-9906-5d3c8f107610)

![Image](https://github.com/user-attachments/assets/b8120094-c30d-49af-b0e0-9ffd61f82e17)

The connector used in BYD is TE 1-1355211-2 
https://www.te.com/en/product-1-1355211-2.html
Mating Part
https://www.te.com/en/product-1-1670459-2.html

qzwf:
| **BYD Connector Pinout**  | **Comma 3x Connector Pinout (18pin ) V3** |
|---------------------------|-------------------------------------------|
| 1 - GND                                | 1,17                                      |
| 2 - IGN1                                | 2,14                                      |
| 3 - Private CAN_H               | 4,16                                      |
| 4 - Chassis Network CAN_H | 8,9                                       |
| 7 - Private CAN_L                  | 6,18                                      |
| 8 - Chassis Network CAN_L | 7,11                                      |


Based on the above, we can determine the pin layouts as above.

qzwf:
![Image](https://github.com/user-attachments/assets/cbae6310-5c04-46e0-b88f-16df4e31a2e0)

adeebshihadeh:
Looks good! A few questions:
* Does this support openpilot longitudinal control?
* How many BYD models does this support?

qzwf:
I completed the port for openpilot steering control 
https://github.com/qzwf/openpilot/tree/byd-atto3-stable

Issue #2017, opened 2025-03-20
BYD port


adeebshihadeh:
Let's get a BYD supported! Any model is eligible.

Requires a harness design/schematic, a high quality lateral port, and ideally longitudinal support.

1okko:
I am willing to sponsor an additional $1000

qzwf:
Awesome! I have solved the harness problem for BYD Atto 3. Have updated the same as Issue#2065. The harness works perfectly with the BYD Atto 3 and Comma 3x.
https://github.com/commaai/opendbc/issues/2065

wenchaoliu-93:
I already found a product listing on Taobao that claims BYD support. Records show they have sold over three hundred units, and that they have over two hundred reviews.

Link is [here](https://item.taobao.com/item.htm?from=cart&id=773413540122&pisk=gVxobzAN1U75XflAknsSSe7OGm3YVgsCTBEd9MCEus5X2Y_d9EXe9IMIFgpFtBAAG3H5J6L0KdODtwM5965H9p8-Hcn9FLsCYMctXcBrBa7BZ9zPYv5VQOkAUUdeJ1ICYfh3kk7SjMOyxzZa8rWVG9XUaMRF0iWCEMWP8_kcg9X_87ReTxbVp92zUwrr0sWNg7PPa6SVgOBNTT7F8xvVG9SFTB7rnKlDvWCM8ExqG3q-CyBdTnXlr1JmlkrH9lQP_ef4Yf-cEf1wUsqUY_HpSVJDebq6HgtBsOd-0uAMCnOc3ConiMOX8BWHgcNdVFdWfZA-7X9c09bNYZ2UYKjPZwW5bcPchEJJ83OuKkWW2hQh1ZDUAwI2XN8wZJM63gWeONtK1u1yLdt6WGmYzGJD8CYh4Ua478y4Atkp0yaCzt6c69ORIx4p6xTqnx4_naWfFAHmnyaCzt6c6xD07obPhTMO.&skuId=5758307511349&spm=a1z0d.6639537%2F202410.item.d773413540122.61677484ZzLxjU).

cuter-doc0:
you mean what bukapilot have already done?
https://github.com/kommuai/bukapilot/blob/snapshot/opendbc/byd_general_pt.dbc

greylot:
Are only the DBC file and schema required?
Is a working prototype mandatory?
Is delivery accepted without physical testing?

HotIce0:
[...]

PR #3677, opened 2026-08-11
BYD ATTO3: new brand port (lateral)

qzwf:
Adds BYD as a brand with the ATTO3 (2022-24) as its first platform, at community support level. Submission for the BYD port bounty, #2017. Harness work and connector pinout are in #2065.

Lateral is done and driven daily on a comma 3X. Longitudinal stays with the stock ACC.

## What the car does

The EPS takes an **absolute steering wheel angle** target in `STEERING_MODULE_ADAS` (0x1E2), so this is an angle-control port, not a torque one. Two properties of the car shaped the design and are worth calling out, because both were found the hard way:

**The camera's steering frame cannot be synthesised.** Everything in 0x1E2 other than the angle is held at the constant the camera itself transmits while steering (bytes 0-2 = `2b 55 eb`, at every speed). Treating those bytes as a sequence and walking them made the car drop its *entire* ADAS mid-drive — LKAS, ACC and the cluster indicators all went away until a power cycle. The controller now copies the camera's own latched frame when it has seen the camera steer, and falls back to the captured constant otherwise (a camera that never gets a lock never emits one).

**The EPS must never see two sources.** openpilot transmits 0x1E2 and `LKAS_HUD_ADAS` (0x316) only while it is actually steering. The safety mode blocks the camera's copies per-frame while that is happening and hands both back ~150 ms after openpilot goes quiet, so the stock LKAS resumes on disengage and there is never a window with neither source driving the EPS. Static blocking is disabled for exactly this reason; `check_relay` is kept so stuck-relay detection still works.

The angle command is anchored to the measured angle and clamped to a fixed ±12° window around it. Without that anchor a limiter that tracks only its own previous output can ratchet away from the wheel, saturate at the clamp, and get every frame rejected by panda — which is what happened before this was in place.

## Safety

`SAFETY_BYD` uses the standard angle checks: speed-dependent rate limits (matching `CarControllerParams.ANGLE_LIMITS`), inactive-angle tracking, and driver override on `DRIVER_EPS_TORQUE`, the column torque sensor. `MAIN_TORQUE` from `STEERING_TORQUE` is EPS motor output, not driver input, and is deliberately not used for override.

`ACC_CMD` is deliberately absent from the TX allowlist, so panda blocks longitudinal outright.

- `opendbc/safety/tests/test_byd.py`: 34 tests, including coverage for the frame-template behaviour above
- MISRA: clean
- `ruff` / `ty` / `codespell` / `cpplint`: clean
- Full car + safety suites pass

## Signal notes

A few signals on this platform are not what their DBC names suggest, all verified against a controlled pedal capture and ~3.6 h of driving:

- `DRIVE_STATE.BRAKE_PRESSED` is dead — byte 4 is a constant `0x0C`
- `PEDAL_PRESSED_ACTIVE_LOW` is the brake-*light* switch; 86% of its on-road assertions happened while the camera's ACC was commanding decel, not the driver. `brakePressed` uses `PEDAL.BRAKE_PEDAL` instead, the only signal that tracks the driver's foot.
- `ACC_HUD_ADAS.ACC_ON1`/`ON2` are main-switch/standby state, not engagement; they stay set while the driver brakes and drives manually. Engagement comes from `ACC_CMD` (`CMD_REQ_ACTIVE_LOW`).
- `WHEELSPEED_BR`'s high byte is a status byte, not speed; wheel speed sums the three trustworthy wheels.
- `STEER_REQ_ACTIVE_LOW` is not the inverse of `STEER_REQ` — the camera holds it at 0.

## Known gaps

- **No openpilot longitudinal.** `pcmCruise = True`, stock ACC only. The `ACC_CMD` scaffolding in `carcontroller.py` is gated behind `openpilotLongitudinalControl` and unreachable today; it needs `ACCEL_CMD` scaling calibrated on the car before it can be enabled. Happy to strip it if you'd rather not carry it.
- **No FW fingerprint.** The car matches on its CAN message set. Firmware versions collected so far were not reproducible across cars, so they are left out rather than shipped unverified; `FW_QUERY_CONFIG` still queries bus 0 so versions keep being collected.
- **No test route in `routes.py` yet.** `b8b2d1a1df1b3aad|000000a0--7c51639c22` is driven and public, but only qlogs have been uploaded so far and `test_models` needs rlogs. `BYD_ATTO3` is in `non_tested_cars` until those finish uploading, at which point it's a one-line change.
- **Harness is `CarHarness.custom`** pending a name. The camera is a Veoneer MVS4; the car-side connector is TE `1-1355211-2` (mating `1-1670459-2`). Full pinout in #2065.

## Bus layout

| Bus | Contents |
|---|---|
| 0 | car-side chassis CAN (read car ECU signals) |
| 1 | camera's private CAN (image/radar, CAN-FD) |
| 2 | camera-side chassis CAN (camera's ADAS messages) |

Port by QZWF.


adeebshihadeh:
Merged https://github.com/commaai/opendbc/pull/3337. Can you rebase and rebuild off master?

PR #3679, opened 2026-08-12
Honda - DBC update - add 'B' / regen gear

mvl-boston:
Update Honda DBC to add regen gear.

This gear is always-on regen braking.  ACC is disabled but LKAS allowed.  This will allow forks that steer without ACC to support regen mode.

See bookmark at 795b2c540a031392/00000234--1eb82984e1


sshane:
Are you sure stock ACC doesn't allow you to engage? You can with Nissan and Volkswagen MEB, but not Toyota

mvl-boston:
B gear doesn't allow ACC according to test drivers.  Reading the owners manual it says D is required for ACC (though remains silent on B).

PR #3684, opened 2026-08-13
test: fuzz carcontroller TX safety

danielpolimac:
## Summary

Adds reproducible fuzz coverage for CAN messages generated by every car controller.

Related to commaai/openpilot#32425.

## What it tests

- Generates bounded lateral and longitudinal `CarControl` inputs across active and inactive states.
- Runs each platform's own `CarController`, preserving its normal clipping and message cadence.
- Requires every generated CAN message to be accepted by `panda` TX safety with controls allowed.
- Exercises cancel/resume combinations alongside actuator values.
- Initializes Panda alternative-experience flags from `CarParams`, matching runtime safety configuration.

The test uses the existing deterministic `fuzzy_test` infrastructure (50 examples per platform) and can be reproduced with `FUZZ_SEED` and `FUZZ_EXAMPLE`.

## Validation

- `python3 -m compileall -q opendbc/car/tests/test_models.py`
- `git diff --check`
- Focused route test: `MAX_EXAMPLES=10 uv run --python 3.12.10 ... test_panda_safety_tx_fuzzy`
  - Passed on Acura ILX route `917b074700869333/2021-05-24--20-40-20` in 1.139 seconds.

The full cross-platform test matrix remains for CI.

### DIFF
--- a/opendbc/car/tests/test_models.py
+++ b/opendbc/car/tests/test_models.py
@@ -187,6 +187,7 @@ class TestCarModelBase(unittest.TestCase):
     cfg = self.CP.safetyConfigs[-1]
     set_status = self.safety.set_safety_hooks(cfg.safetyModel.raw, cfg.safetyParam)
     self.assertEqual(0, set_status, f"failed to set safetyModel {cfg}")
+    self.safety.set_alternative_experience(self.CP.alternativeExperience)
     self.safety.init_tests()
 
   def test_car_params(self):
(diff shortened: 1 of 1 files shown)


sshane:
I would expect to see some failures. We're only going to merge a test that catches real issues + fixing them first

danielpolimac:
Thanks, agreed. I traced the Toyota LTA torque-oscillation regression from panda#1948, added a regression test that fails on current master, and fixed it in #3686. The test-only PR should stay closed.

PR #3696, opened 2026-08-18
hyundai: add 2025 EU Sportage firmware

USMortality:
## Goal

Add firmware versions for a 2025 EU Kia Sportage MHEV GT-Line under the existing `KIA_SPORTAGE_5TH_GEN` platform.

## Vehicle Test

Tested on vehicle with sunnypilot staging:
- auto-detected as Kia Sportage 5th Gen
- CAN valid
- torque lateral engaged successfully
- longitudinal control tested successfully

No route or dongle ID is included in this PR body because the test device is currently offline.

## Local Checks

- `python3 -m py_compile opendbc/car/hyundai/fingerprints.py`

Full pytest was not run locally because `pytest` is not installed in the temporary checkout environment.


adeebshihadeh:
We need a test route to confirm before merging.

PR #3701, opened 2026-08-23
test_models: fuzz the messages openpilot sends

AJ0070:
Adds `test_panda_safety_tx_fuzzy` for commaai/openpilot#32425.

Replays route CAN into both the car interface and safety so they derive the same vehicle state, then fuzzes the CarControl controlsd would build on top of it. Every message the controller sends has to pass `safety_tx_hook`, so a rejection means openpilot's limits and safety's disagree. Fuzzing more state is one line in `fuzzy_controls()`.

Reverting commaai/panda#1948 makes it fail on TOYOTA_RAV4_TSS2_2023 STEERING_LTA across seeds. `test_panda_safety_tx_cases` doesn't catch that.

It also found a VW MEB bug, fixed here. The controller stepped its curvature command straight to the measured value when lateral went inactive while still engaged. HCA is still on there, so safety still applies the ISO jerk limit and rejects it. It ramps now.

Green on 191 of 192 platforms with routes over 3 sweeps with different seeds, about 2s each.

The one left is Ford CAN FD, roughly once per 3 full runs. openpilot clips to `current_curvature +- CURVATURE_ERROR` using the yaw rate and speed at control time, while safety builds its band from `curvature_state.meas`, a 6 sample window updated only when the yaw message arrives. On FORD_MAVERICK_MK1 openpilot's reading is outside that window in 413 of 922 samples, by up to 22 CAN units against 1 unit of slack. That looks like the sampling problem the FIXME in `ford_rx_hook` already mentions, so I left it for you rather than guess at the fix.

### DIFF
--- a/opendbc/car/tests/test_models.py
+++ b/opendbc/car/tests/test_models.py
@@ -46,10 +46,9 @@ ANGLE_DEG_TO_CAN = {
 TX_FUZZ_WARMUP = 200
 TX_FUZZ_FRAMES = 150
 
-# Bounds controlsd keeps its actuators within. Car controllers clamp further, and it's that clamping
-# panda has to agree with, so these only need to be wide enough to push every controller into its limits.
-MAX_ACTUATOR_ANGLE = 180.  # deg
-MAX_ACTUATOR_CURVATURE = 0.2  # 1/m, MAX_CURVATURE in openpilot's clip_curvature
+# MAX_CURVATURE in openpilot's clip_curvature, the bound controlsd keeps its desired curvature within.
+# The angle command is derived from it, so this also sets how far the angle controllers get pushed.
+MAX_ACTUATOR_CURVATURE = 0.2  # 1/m
 
 NUM_JOBS = int(os.environ.get("NUM_JOBS", "1"))
 JOB_ID = int(os.environ.get("JOB_ID", "0"))
@@ -116,7 +115,6 @@ def fuzzy_controls(fuzzy) -> dict:
     "lat_active": fuzzy.boolean(),
     "long_active": fuzzy.boolean(),
     "torque": fuzzy.real(-1., 1.),
-    "angle": fuzzy.real(-MAX_ACTUATOR_ANGLE, MAX_ACTUATOR_ANGLE),
     "curvature": fuzzy.real(-MAX_ACTUATOR_CURVATURE, MAX_ACTUATOR_CURVATURE),
     "accel": fuzzy.real(ACCEL_MIN, ACCEL_MAX),
     "long_control_state": fuzzy.choice(tuple(LongControlState.schema.enumerants)),
@@ -144,6 +142,9 @@ def build_car_control(controls: dict, CP, VM: VehicleModel, CS) -> structs.CarCo
   # while inactive the controllers command the car's current state, not the planner's, see
   # selfdrive/controls/lib/latcontrol_angle.py and longcontrol.py
   curvature = -VM.calc_curvature(math.radians(CS.steeringAngleDeg - CS.steeringAngleOffsetDeg), CS.vEgo, 0.)
+  # the angle command is the desired curvature through the vehicle model, so it reaches as far as the
+  # car's steering ratio allows rather than an arbitrary cap
+  angle = math.degrees(VM.get_steer_from_curvature(-controls["curvature"], CS.vEgo, 0.))
 
   return structs.CarControl(
     enabled=enabled,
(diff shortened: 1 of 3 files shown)


AJ0070:
@sshane @adeebshihadeh ready for review whenever you get a chance. This is the bounty on commaai/openpilot#32425.

MyTH-zyxeon:
One coverage gap at exact head `993df9e93ab8bead5bae0c9f7933330947472f43`:

`MAX_ACTUATOR_ANGLE` caps every fuzz target at +/-180 deg ([test_models.py#L49-L51](https://github.com/commaai/opendbc/blob/993df9e93ab8bead5bae0c9f7933330947472f43/opendbc/car/tests/test_models.py#L49-L51), [#L118-L120](https://github.com/commaai/opendbc/blob/993df9e93ab8bead5bae0c9f7933330947472f43/opendbc/car/tests/test_models.py#L118-L120)), but the same head allows substantially larger commands for angle controllers: Nissan 600 deg ([values.py#L13-L20](https://github.com/commaai/opendbc/blob/993df9e93ab8bead5bae0c9f7933330947472f43/opendbc/car/nissan/values.py#L13-L20)), PSA 390 deg ([values.py#L12-L19](https://github.com/commaai/opendbc/blob/993df9e93ab8bead5bae0c9f7933330947472f43/opendbc/car/psa/values.py#L12-L19)), and Tesla 360 deg ([values.py#L110-L116](https://github.com/commaai/opendbc/blob/993df9e93ab8bead5bae0c9f7933330947472f43/opendbc/car/tesla/values.py#L110-L116)).

Because the test intentionally holds each target so the controller rate limiter winds up, it can only wind up toward 180 deg. A controller/panda mismatch in `(180, controller_limit]` can therefore remain undetected while this test passes. Please derive the fuzz bound from the selected controller's angle limits, or explicitly add `+/-controller_limit` boundary targets for angle-control platforms. That keeps the generic test aligned with the bounty's goal instead of leaving the highest-command region untested.

Run-Id: `run-BC1B54CB-FF02-4D1D-912C-00F51279CE4E`
Trace-Id: `FDFD50D2-21BA-45CD-9507-B2EF722915E7`
Requester: Jun automation id `monitor-sxt-bounty-pr-1757`
Implementer: MISA 3 bot ID `1516725819517567077`

AJ0070:
Good catch, fixed in 1080d55.

You were right and the gap was bigger than 180 vs the per car limit. The angle command comes from the desired curvature through the vehicle model, so what controlsd can actually reach is `degrees(VM.get_steer_from_curvature(MAX_CURVATURE, v_ego, roll))`. That peaks around 4 m/s where the 0.2 curvature cap dominates, and it's 400 to 590 deg on every angle platform, not just the three you listed. Toyota LTA reaches 447, Nissan 533, Tesla 400.

Rather than special casing per platform I dropped `MAX_ACTUATOR_ANGLE` and derive the angle the same way `LatControlAngle` does. That makes the bound correct per car and per speed for free, and it's more faithful anyway since controlsd populates `steeringAngleDeg` and `curvature` from the same desired curvature instead of independently.

Measured commanded angle after the change: TOYOTA_RAV4_TSS2_2023 605 deg, NISSAN_LEAF 560, TESLA_MODEL_3 475. All well past each car's `STEER_ANGLE_MAX`, so the controller clamps are exercised now.

Still green on 3 full sweeps of every platform, and reverting commaai/panda#1948 still fails on STEERING_LTA across seeds.

MyTH-zyxeon:
Follow-up at exact head `1080d55f4a7ab51c30af5b28c5f0ff9c3e0c0cc1`: I compared the one-commit delta against the previously reviewed `993df9e93ab8bead5bae0c9f7933330947472f43`, and the original coverage finding is closed.

The arbitrary `MAX_ACTUATOR_ANGLE = 180` target is gone. The fuzzer now draws desired curvature through the existing `+/-0.2` controlsd bound ([test_models.py#L49-L51](https://github.com/commaai/opendbc/blob/1080d55f4a7ab51c30af5b28c5f0ff9c3e0c0cc1/opendbc/car/tests/test_models.py#L49-L51), [#L112-L119](https://github.com/commaai/opendbc/blob/1080d55f4a7ab51c30af5b28c5f0ff9c3e0c0cc1/opendbc/car/tests/test_models.py#L112-L119)) and derives `steeringAngleDeg` through `VehicleModel.get_steer_from_curvature` ([#L142-L158](https://github.com/commaai/opendbc/blob/1080d55f4a7ab51c30af5b28c5f0ff9c3e0c0cc1/opendbc/car/tests/test_models.py#L142-L158)). That matches the core desired-curvature-to-angle conversion used by `LatControlAngle` ([openpilot exact source](https://github.com/commaai/openpilot/blob/084747c75d2cbd23af65ab7a9e770bbd7b98bac9/openpilot/selfdrive/controls/lib/latcontrol_angle.py#L16-L25)) and lets each platform's controller clamp, rather than the test's global constant, define the exercised boundary.

The PR remains mergeable/clean and all 12 visible checks on this head are non-failing. I did not independently reproduce the author's full route sweeps, so this closes my prior code-level finding without making a separate claim about that runtime evidence. Maintainer acceptance, merge, and bounty selection remain outstanding.

Run-Id: `run-40B85951-2387-4F21-8F1B-1FF233B85F2A`
Trace-Id: `EBFA6D37-C845-455A-A8E3-A93BBCD28F34`
Requester: Jun automation id `monitor-sxt-bounty-pr-1757`
Implementer: MISA 3 bot ID `1516725819517567077`

AJ0070:
@sshane gentle ping!

PR #3702, opened 2026-08-23
safety: enable the branch coverage check

### LINKED ISSUE #2557: Enable 100% branch coverage check

AJ0070:
Turns on `--fail-under-branch=100` and fixes the 157 violations. Closes #2557.

Mostly removing and rewriting, as you expected. The RX check and TX message tables already pin each address to a bus and length, and the hooks only run for whitelisted messages, so the per-mode bus guards and `(addr == X) && (bus == Y)` pairs could never take their other branch. Same for flag conditions on addresses that are only RX checked when that flag is set. The trailing `else if` in the checksum, counter and quality flag dispatchers became `else`. Tests were added only where the branch is actually reachable.

`safety.h` keeps its misconfiguration guards, with test only hooks in `libsafety/safety.c` to build the configs they're there for.

Since this mostly deletes conditions from safety code, I checked it's behaviour preserving rather than just green. Built libsafety from master and from this branch, ran the whole safety suite with both loaded, and compared every call into each. 11M paired calls, no divergences. Injecting a fault at each kind of site that changed is caught, so that number isn't vacuous. Mutation testing leaves the same survivors as master, 3 pre-existing ones in `lateral.h`. No test is removed or newly skipped.

Green on `test.sh` (100% line and branch), MISRA, cpplint, ruff, ty and the full suite.


AJ0070:
@adeebshihadeh @sshane ready for review whenever you get a chance. This is the bounty on #2557.

adeebshihadeh:
Thanks for the PR - it's better than most slop people put up for this issue. But this could be a lot better. If you're curious, see what I've been doing on master.

AJ0070:
Redoing this in smaller pieces. First one: the ESP_status quality flag test picks one of the two invalid values at random, so tesla.h loses a branch about 1 in 8 runs. Flaky as soon as the branch check is on.

#3792

PR #3761, opened 2026-09-13
mazda: set enableBsm parameter

jingshenSN2:
Validation
* Dongle ID: b0d2bd701938150b
* Route: b0d2bd701938150b/00000009--f20db3c785
Mazda CarState already parses BSM status from message 0x477, but CarParams.enableBsm is never set. Set it based on the presence of 0x477 in the fingerprint.


adeebshihadeh:
So this is confusing, since it doesn't need to be set here.

I'm just removing it here: https://github.com/commaai/opendbc/pull/3788.

jingshenSN2:
I see. Let me add it there for Mazda as well.

adeebshihadeh:
No, there's no need. It literally wouldn't do anything.

PR #3796, opened 2026-09-18
Tesla: detect new 3-bit steering control enum

sshane:


opendbc/car/tesla/interface.py
```diff
@@ -27,6 +27,13 @@ def _get_params(ret: structs.CarParams, candidate, fingerprint, car_fw, alpha_lo
     if 0x293 not in fingerprint[CANBUS.autopilot_party]:
       ret.flags |= TeslaFlags.MISSING_DAS_SETTINGS.value
 
+    # Tesla expanded DAS_steeringControl->DAS_steeringControlType to 3 bits around 10-26-2025 with the FSD 14 update to HW4 vehicles.
+    # HW3 vehicles then got an update around XX-XX-XXXX to also use this new 3-bit signal definition.
+    # Checking 0x489 existence seems to detect the new signal definition
```
adeebshihadeh: can we find two different things to check? or look into this message some more? (does the new DBC have it?)

sshane: found it, 0x489 is `DAS_redundantBrakingControl`

found another signal for HW2.5, it's 0x054 which is `DI_autonomyHealth`

sshane:
Merged detection in https://github.com/commaai/opendbc/pull/3802

PR #3797, opened 2026-09-18
Tesla: Decode DAS_steeringControlType as a 3-bit field

AmyJeanes:
This PR supersedes #3389 at the request of @sshane who asked me to split out the core changes without the fingerprint table and use an affirmative name instead for the flag instead of the negative flag (is legacy) it originally had so it maintains it's meaning for old logs.

On newer Tesla vehicle firmware across the fleet, we are seeing the `DAS_steeringControlType` field expanded from 2 bits to 3 bits in the dbc. It is in the same position, the field simply expands by 1 bit into previously unknown/unmapped bits which we can assume were not in use before.

This change first appeared in FSD v14 builds, as the expansion of the field was to allow a new value `4 = FSD` to be set as the previous field could only go up to 3, but after some time this change found it's way to normal vehicle builds from update 2026.8.6 onwards and thus is no longer a FSD v14 specific change as was previously noted in the code here.

The resulting behaviour of openpilot when assuming a 2-bit field when it was actually 3-bit resulted in the following behaviour:
- Engaging openpilot would result in `LANE_KEEP_ASSIST` steering instead of `ANGLE_CONTROL` steering as `0b01` would be sent by openpilot but read by the car as `0b010`
- Vehicle LKAS ("Lane Departure Avoidance") events would result in the FSD v14 suspect being triggered, resulting in openpilot forcefully disengaging and locking out the user with `invalidLkasSetting` until the device went off-road.
- In the case of sunnypilot with it's Coop Steering setting, enabling that would also result in complete loss of steering as `0b10` would be sent, which was read by the car as `0b100` aka FSD steering which apparently does not work

The changes in this PR essentially allow us to swap between the two different behaviours, the new one with the 3-bit field and the "legacy" one with the 2-bit field, currently we are re-using the `FSD_14_FW_RULE` from master but renamed to `DAS_STEERING_3_BIT_FW_RULE` though @sshane plans to modify this shortly in his own PR with better CAN derived heuristics. The firmware lists are untouched in this PR, although there are many candidates to be added from the original PR #3389 in a follow up change.

The previous FSD v14 handling has been stripped out. It assumed the enum values had been remapped - which is only true for FSD, a genuinely new value (0b100 = 4) that didn't fit the old 2-bit field, and adding it is why the field widened.

The existing values kept their numbers (`ANGLE_CONTROL` = 1, `LANE_KEEP_ASSIST` = 2); each just sits one bit lower now, so code still reading the top two bits misreads them: `LANE_KEEP_ASSIST` (`0b010`) is cut to `0b01` and looks like `ANGLE_CONTROL`, and `ANGLE_CONTROL` (`0b001`) is cut to `0b00` and goes missing.

Note: Replay changes are expected due to the removal of the `suspected_fsd14` fallback. This will be fixed properly with @sshane's follow up PR with improved detection.

No routes can currently be made for this PR as openpilot master is currently broken against opendbc master. Once they are syncing again I would be happy to create a route for this, but the core behaviour change in the original PR #3389 has been extensively tested by myself and others in forks for months across different Tesla firmwares and vehicles.


sshane:
Triming to rename + documentation as we're going to drop 2-bit FW support in https://github.com/commaai/opendbc/pull/3796

PR #3802, opened 2026-09-21
Tesla: detect 3-bit steering control, drop 2-bit support

sshane:
This PR is to add 3-bit auto-detection, remove the fragile FW list, deprecate 2-bit cars (all that are still on 2-bit FW, 4 dongles, can update), and make sure sending never-sent before 0b001 still steers (should still be `ANGLE_CONTROL`). After the FW update our Tesla got we were sending 0b10 (0b100) which we thought was stiil `ANGLE_CONTROL`, just renamed to `LANE_KEEP_ASSIST` but was only sent by FSD after that update, so we added it to the DBC named FSD.

The next PR will add some simple safety to make sure the EPS is reporting what we expect given what steering control type we're sending.

for those current 2-bit users with drives in last 30 days:

| dongle | model | EPS FW | last seen | other cars off this FW |
|---|---|---|---|---|
| `01205fbf88d34074` | Model Y | `TeM3_E014p10_0.0.0 (16),Y002.18.00` | 2026-09-17 | 17 of 37 |
| `5d25c9abc71f2166` | Model Y | `TeM3_E014p10_0.0.0 (16),Y002.18.00` | 2026-09-16 | 17 of 37 |
| `a389127cd8fc2c04` | Model 3 | `TeM3_E014p10_0.0.0 (16),E014.17.00` | 2026-09-16 | 27 of 56 |
| `ece9855b75526162` | Model Y | `TeMYG4_Legacy3Y_0.0.0 (5),Y4003.03.2` | 2026-08-30 | 20 of 39 |

<details><summary>codex description</summary>

Follow-up to #3797, which decoded the widened `DAS_steeringControlType` but kept both encodings behind a FW list.

### Detection from CAN

Two messages only exist on firmware with the 3-bit signal, so we don't need a FW list:
- `DAS_redundantBrakingControl` (0x489, 50 Hz), sent by HW3/HW4 Autopilot computers
- `DI_autonomyHealth` (0x054, 10 Hz), sent by the car, so it also covers HW2.5

Across 7.5k labelled segments of our Tesla data, `0x489 OR 0x054` had no false negatives against the firmware lists. The three disagreements are all the first drive after an update, where the FW string came from the CarParams cache while the CAN was already 3-bit. Both messages are ≥10 Hz and show up a median of 9 ms after the car wakes, so they're always inside the fingerprint window.

### 2-bit firmware is now dashcam only

Sending the legacy `0b010` to a 3-bit car requests `LANE_KEEP_ASSIST`, which lets the EPS wind down against the driver. Rather than keep both encodings alive, cars on the older firmware are dashcam only: openpilot always sends `ANGLE_CONTROL` (`0b001`) and panda only allows `NONE` or `ANGLE_CONTROL`, so that value can't be sent to any car. A car that's misdetected either way ends up in dashcam or simply doesn't steer.

This lets `FSD_14_FW`, `TeslaFlags.DAS_STEERING_3_BIT`, `TeslaSafetyFlags.DAS_STEERING_3_BIT` and `get_steer_ctrl_type` all go away, and `tesla.h` goes back to the pre-FSD 14 code with a 3-bit read.

Of the Tesla cars active in the last 30 days of our data, 3% (4 of 147) are still on 2-bit firmware, and each of those is on an exact firmware version that 17–27 other cars have already updated from.

### Note

All four Tesla routes in `routes.py` are 2-bit firmware, so they're dashcam now and `test_models` skips their checks. They need replacing with 3-bit routes.

</details>


sshane:
Edge cases from the data behind this, with conclusions:

**Disagreed with the FW lists**
| route | conclusion |
|---|---|
| `049f9ed61c93ec27/00001404--fab752a169/5` | broken CAN |
| `a923ba6ead24ee2a/00000009--cee8b85af4/1` | broken CAN |
| `3b205ebbffadef1a/00000a5b--6d97c6ff44/11` | broken CAN |
| `4e5d75a1f3df3ae2/0000095d--9bcd9668fd/0` | cached FW |
| `717f468944a01e1e/00000183--386c572313/4` | cached FW |
| `defd22b534070bb2/00001867--b7d634994f/11` | cached FW, probably |

**HW2.5 (no 0x489, 0x054 covers them)**
| route | conclusion |
|---|---|
| `3aa9963ec61c5915/000000f2--1b838ea6b3/35` | HW2.5, 3-bit |
| `defd22b534070bb2/00000478--919d360e19` | HW2.5, 3-bit |
| `e316767caeaf1d7b/00000122--61af6316ed/1` | HW2.5, 2-bit |
| `424b2a6c1b69a3bc/000000fd--4b03b790ae/22` | HW3, 3-bit |

**Timing and bus oddities**
| route | conclusion |
|---|---|
| `9d8d7374db0df9bf/00000050--5117b1fc8d/0` | 0x489 late, 0x054 early |
| `64efe75db7b9d883/00000000--3add6452e5/2` | old AP software |
| `2e526afd5a4ab2bf/00000001--ddee4f3f02/0` | bus 2 dropout |
| `5fb9dc19614e327b/00000001--348d32ebca/24` | fork, ignored |

**EPS stayed in LANE_KEEP_ASSIST with the correct encoding**
| route | conclusion |
|---|---|
| `64e16c4237493597/00000033--332a5f9d82/15` | sunnypilot LKA latch |
| `db792d9f5ca5ef59/00000011--8e9db3cc8e/4` | same latch |
| `2a5a0210bcc9b83b/00000069--85a8451459/13` | same, assumed |

